PR #2, opened 2019-09-11
feat: add support for CodeReady Containers

MatousJobanek:
when the makefile tries to log in, it first checks if there is running VM via CRC. 
  * If there is, then it tries to log in as `kube:admin`
  * if there is no CRC VM running, then it logs as `system:admin`


make/test.mk
```diff
@@ -10,8 +10,11 @@ TEST_NS := toolchain-e2e-$(shell date +'%s')
 AUTHOR_LINK := $(shell jq -r '.refs[0].pulls[0].author_link' <<< $${CLONEREFS_OPTIONS} | tr -d '[:space:]')
 PULL_SHA := $(shell jq -r '.refs[0].pulls[0].sha' <<< $${CLONEREFS_OPTIONS} | tr -d '[:space:]')
 
+IS_CRC := $(shell crc status /dev/null 2>&1 | grep Running)
+IS_KUBE_ADMIN := $(shell oc whoami | grep "kube:admin")
```
xcoulon: why is it `kube:admin` and not `system:admin`?

xcoulon: ok, that's too bad they did not reuse the same username for the admin account, though :/

PR #3, opened 2019-09-19
fix: let CI build operators images before e2e test execution

MatousJobanek:
We need to let CI build both operators images before the actual e2e test execution so we have correct images to deploy
see this PR: https://github.com/openshift/release/pull/5140


MatousJobanek:
I see your point @alexeykazakov 
There is a way we could do it, it will slightly complicate the makefile but not significantly. 
Let's merge this PR to unblock the other guys and I'll create the generic deployment as part of the following task.

PR #4, opened 2019-09-20
fix: add missing delete case

MatousJobanek:
somehow, while moving the e2e tests, I missed the deletion test case. This PR merges:
https://github.com/codeready-toolchain/host-operator/blob/08e3262b122db594b4e06f93f077112d1e265cec/test/e2e/masteruserrecord_test.go#L79-L96
and 
https://github.com/codeready-toolchain/member-operator/blob/0fce69b64078055ef51c2ca33e793bc7c86ac33f/test/e2e/useraccount_test.go#L115-L131

### DIFF
--- a/e2e/e2e_test.go
+++ b/e2e/e2e_test.go
@@ -256,8 +256,8 @@ func verifyDeletion(awaitility *wait.Awaitility, mur *toolchainv1alpha1.MasterUs
 	assert.NoError(awaitility.T, err, "UserAccount is not deleted")
 
 	err = memberAwait.WaitForDeletedUser(mur.Name)
-	assert.NoError(awaitility.T, err)
+	assert.NoError(awaitility.T, err, "User is not deleted")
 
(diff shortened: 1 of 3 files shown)


MatousJobanek:
@alexeykazakov no, it is not - the test case deletes MUR and expects that all other resources will be deleted automatically by garbage collector, so it's just verified here: https://github.com/codeready-toolchain/toolchain-e2e/pull/4/files#diff-783e65d8d8d51bbdfd38558f35cca4caR255-R261

alexeykazakov:
Ah, you merged these tests. Makes sense.

MatousJobanek:
> Ah, you merged these tests. Makes sense.

yes, exactly. And this is where the original error in my head happened - while merging these two files I somehow missed the deletion case... But it's weird because I'm sure that at that time, I was doing the very same steps as I did a few minutes ago. So it's either lost somewhere in the merges or in other branch in my local env or it was just the déjà vu :-/

PR #12, opened 2019-10-07
deploy reg service for e2e test setup

nurali-techie:
Fixes: https://jira.coreos.com/browse/CRT-306


make/test.mk
```diff
@@ -199,3 +210,23 @@ else
...
 	sed -e 's|REPLACE_IMAGE|${IMAGE_NAME}|g' ${E2E_REPO_PATH}/deploy/operator.yaml | oc apply -f -
 
+.PHONY: build-and-deploy-deployment
+build-and-deploy-deployment:
+# when e2e tests are triggered from different repo - eg. as part of PR in host-operator repo - and the image of the deployment is (not) provided
+ifeq ($(SET_IMAGE_NAME),)
+	# check if it is running locally
+    ifeq ($(OPENSHIFT_BUILD_NAMESPACE),)
+		# if it is running locally, then build the image via docker
+		$(eval IMAGE_NAME := docker.io/${GO_PACKAGE_ORG_NAME}/${REPO_NAME}:${GIT_COMMIT_ID_SHORT})
```
alexeykazakov: We will need to stop using docker.io and switch to quay.io instead. We use docker.io currently only for local images (we don't really push anything to docker.io) but still would be better to switch to quay.io.
But it's OK do it in a separate PR.

MatousJobanek: it's partially already done - we check the OpenShift version (if it is 3 or 4) and based on that we use either quay.io (for OS 4) or docker.io (for OS 3 - minishift) for the local testing. https://github.com/codeready-toolchain/toolchain-e2e/blob/master/make/test.mk#L190-L204 
But I'm fine with dropping docker.io completely (we will need to do it anyway as soon as minishift is not usable/compatible enough)

alexeykazakov: even for minishift we don't need docker.io. Currently we use docker.io only for local images (for running them in minishift). The images which are never pushed to docker.io. So, it doesn't matter if we use docker.io or quay.io or whatever in the image names for local development :)

nurali-techie: using `quay.io` at least for reg_service deployment which is what we are introducing with this PR.  See commit https://github.com/codeready-toolchain/toolchain-e2e/pull/12/commits/027176b4ceb7eecfd996451ff4cf5a92c72db3bb

make/test.mk
```diff
@@ -43,7 +43,7 @@ test-e2e-host-local:
 e2e-run:
 	oc get kubefedcluster -n $(HOST_NS)
 	oc get kubefedcluster -n $(MEMBER_NS)
-	oc new-project $(TEST_NS) --display-name e2e-tests 1>/dev/null
+	-oc new-project $(TEST_NS) --display-name e2e-tests 1>/dev/null
```
dipak-pawar:
```suggestion
	oc new-project $(TEST_NS) --display-name e2e-tests 1>/dev/null
```

make/test.mk
```diff
@@ -120,6 +120,17 @@ ifeq ($(HOST_REPO_PATH),)
 	$(MAKE) prepare-e2e-repo E2E_REPO_PATH=$(HOST_REPO_PATH) REPO_NAME=host-operator
 endif
 
+.PHONY: get-registration-service-repo
+get-registration-service-repo:
+ifeq ($(REG_REPO_PATH),)
+	$(eval REG_REPO_PATH = /tmp/codeready-toolchain/registration-service)
+	rm -rf ${REG_REPO_PATH}
+	# clone
+	git clone https://github.com/codeready-toolchain/registration-service.git ${REG_REPO_PATH}
```
dipak-pawar: Can we do shallow clone here? something like `git clone -–depth 1 https://github.com/codeready-toolchain/registration-service.git` to improve performance.

make/test.mk
```diff
@@ -180,7 +191,7 @@ deploy-host:
 ifeq ($(HOST_REPO_PATH),)
 	$(eval HOST_REPO_PATH = /tmp/codeready-toolchain/host-operator)
 endif
-	oc new-project $(HOST_NS) 1>/dev/null
+	-oc new-project $(HOST_NS) 1>/dev/null
```
dipak-pawar: I am not sure why are we using `-` here and also to redirect error to `dev/null`,error code is `2` not `1`

make/test.mk
```diff
@@ -164,7 +175,7 @@ deploy-member:
 ifeq ($(MEMBER_REPO_PATH),)
 	$(eval MEMBER_REPO_PATH = /tmp/codeready-toolchain/member-operator)
 endif
-	oc new-project $(MEMBER_NS) 1>/dev/null
+	-oc new-project $(MEMBER_NS) 1>/dev/null
```
dipak-pawar: I am not sure why are we using `-` here and also to redirect error to `dev/null`,error code is `2` not `1`

make/test.mk
```diff
@@ -195,6 +206,19 @@ endif
 	# as the controller starts (which is a use-case for CRT-231)
 	oc apply -f test/e2e/nstemplatetier-basic.yaml -n $(HOST_NS)
 
+.PHONY: deploy-registration
+deploy-registration:
+ifeq ($(REG_REPO_PATH),)
+	$(eval REG_REPO_PATH = /tmp/codeready-toolchain/registration-service)
+endif
+	-oc new-project $(HOST_NS) 1>/dev/null
```
dipak-pawar: I am not sure why are we using `-` here and also to redirect error to `dev/null`,error code is `2` not `1`

nurali-techie:
@alexeykazakov finally e2e test green (there were just infra issues).  I also added org_name in repo_path with commit https://github.com/codeready-toolchain/toolchain-e2e/pull/12/commits/0688f0712d0637523bd67f729691bacb40d275bd the similar changes done by Xavier.

make/test.mk
```diff
@@ -250,3 +274,33 @@ ifeq ($(IS_OS_3),)
...
+        else
+			# using OS 4 then build and push image
+			$(eval IMAGE_NAME := quay.io/${QUAY_NAMESPACE}/${REPO_NAME}:${DATE_SUFFIX})
+			$(MAKE) -C ${REPO_PATH} build
+			docker build -f ${REPO_PATH}/build/Dockerfile -t ${IMAGE_NAME} ${REPO_PATH}
+			docker push ${IMAGE_NAME}
+        endif
+    else
+		# if it is running in CI than we expect that it's PR for toolchain-e2e repo (none of the images was provided), so use name that was used by openshift-ci
+		$(eval IMAGE_NAME := registry.svc.ci.openshift.org/${OPENSHIFT_BUILD_NAMESPACE}/stable:${REPO_NAME})
```
dipak-pawar: Can you please make sure that we are building this image in openshift-ci? I don't remember we have this image available in openshift-ci image registry

make/test.mk
nurali-techie: `-oc` is to ignore error in case namespace already created previously.  This is useful as for host_operator and reg_service we are using same namespace `$(HOST_NS)`

make/test.mk
nurali-techie: @dipak-pawar are you talking about this thing ?? https://github.com/openshift/release/blob/master/ci-operator/config/codeready-toolchain/toolchain-e2e/codeready-toolchain-toolchain-e2e-master.yaml#L16-L32

Can you plz provide me some doc link to do the thing you asking for ??

make/test.mk
nurali-techie: `-oc` with hyphen to ignore error and continue.  This is required as host-operator and registration-service will be deployed on same namespace `$(HOST_NS)` and both will try to create HOST_NS.

I check history and found that `1>/dev/null` was added to clean output with PR https://github.com/codeready-toolchain/toolchain-e2e/pull/13

make/test.mk
nurali-techie: Discussing here https://github.com/codeready-toolchain/toolchain-e2e/pull/12#discussion_r336439458

make/test.mk
nurali-techie: Discussing here https://github.com/codeready-toolchain/toolchain-e2e/pull/12#discussion_r336439458

make/test.mk
nurali-techie: Discussing here https://github.com/codeready-toolchain/toolchain-e2e/pull/12#discussion_r336439458

make/test.mk
nurali-techie: Done https://github.com/codeready-toolchain/toolchain-e2e/pull/12/commits/1b50da335d5059ac5991579812ea9315b1e26beb for all clone we are making in e2e test.

alexeykazakov: We should be careful with that. The repos we clone are used in two use cases:
- Against master (then we do not need any history)
- Against particular branch. In this case we will switch to the PR branch (which can have multiple commits) of the repo and it may be broken if we do shallow clone.

Let's not introduce shallow clones in this PR. Let's discuss/address it separately.

nurali-techie: @alexeykazakov good input, reverted `--depth 1` changes  https://github.com/codeready-toolchain/toolchain-e2e/pull/12/commits/927732311359236e90bce45a6eac64c851823cdd

make/test.mk
nurali-techie: Done https://github.com/openshift/release/pull/5378

PR #26, opened 2019-10-28
add smoke test for the registration service

xcoulon:
waits for the registration service deployment to be ready
and for the route to be available (as part of the setup),
then during the test, just make sure that the homepage is
available (returning a '200 OK' response)

Fixes CRT-322

### DIFF
--- a/test/e2e/registration_service_test.go
+++ b/test/e2e/registration_service_test.go
@@ -35,7 +35,7 @@ func (s *registrationServiceTestSuite) SetupSuite() {
 	s.route = s.awaitility.RegistrationServiceURL
 }
 
-func (s *registrationServiceTestSuite) TestSmokeTest() {
+func (s *registrationServiceTestSuite) TestLandingPageReachable() {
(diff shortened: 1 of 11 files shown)


test/e2e/registration_service_test.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+func (s *registrationServiceTestSuite) SetupSuite() {
+	userSignupList := &v1alpha1.UserSignupList{}
+	s.testCtx, s.awaitility = doubles.WaitForDeployments(s.T(), userSignupList)
+	s.namespace = s.awaitility.RegistrationServiceNs
+	s.route = s.awaitility.RegistrationServiceURL
+}
+
+func (s *registrationServiceTestSuite) TestSmokeTest() {
```
alexeykazakov: `TestSmokeTest()` is too generic IMO.
```suggestion
func (s *registrationServiceTestSuite) TestLandingPageReachable() {
```

PR #28, opened 2019-10-30
rename pkg 'doubles' to 'testsupport'

xcoulon:


xcoulon:
I voluntarily assigned all team members as reviewers so you're all aware of this change when it is approved and merged ;)

PR #32, opened 2019-11-03
Switch to kube 1.14

alexeykazakov:
Requires:
https://github.com/codeready-toolchain/toolchain-common/pull/48 (Merged)
https://github.com/codeready-toolchain/toolchain-common/pull/52 (Merged)
https://github.com/codeready-toolchain/toolchain-e2e/pull/33 (Merged)

Paired with:
https://github.com/codeready-toolchain/host-operator/pull/98
https://github.com/codeready-toolchain/member-operator/pull/89

PR #33, opened 2019-11-03
Switch to go 1.13 in builder image

alexeykazakov:
There is checksum mismatch for operator-sdk binary between go 1.12 and go 1.13. To fix this issue we need to switch our builder image to go 1.13.

Required for https://github.com/codeready-toolchain/toolchain-e2e/pull/32

### DIFF
--- a/openshift-ci/Dockerfile.tools
+++ b/openshift-ci/Dockerfile.tools
@@ -9,7 +9,7 @@ ENV LANG=en_US.utf8 \
     PATH=$PATH:$GOPATH/bin \
     GIT_COMMITTER_NAME=devtools \
     GIT_COMMITTER_EMAIL=devtools@redhat.com \
-    OPERATOR_SDK_VERSION=v0.10.0
+    OPERATOR_SDK_VERSION=v0.11.0
 
 ARG GO_PACKAGE_PATH=github.com/codeready-toolchain/toolchain-e2e
 
@@ -18,7 +18,6 @@ RUN yum install epel-release -y \
     && yum install --enablerepo=centosplus install -y --quiet \
     findutils \
     git2u-all \
-    golang \
     make \
     procps-ng \
     tar \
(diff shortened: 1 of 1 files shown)


openshift-ci/Dockerfile.tools
```diff
@@ -45,7 +55,7 @@ RUN curl -L -s https://github.com/openshift/origin/releases/download/v3.11.0/ope
 # download, verify and install operator-sdk
 RUN curl -L -s https://github.com/operator-framework/operator-sdk/releases/download/${OPERATOR_SDK_VERSION}/operator-sdk-${OPERATOR_SDK_VERSION}-x86_64-linux-gnu -o operator-sdk \
     && curl -L -s https://github.com/operator-framework/operator-sdk/releases/download/${OPERATOR_SDK_VERSION}/operator-sdk-${OPERATOR_SDK_VERSION}-x86_64-linux-gnu.asc -o operator-sdk.asc \
-    && gpg --keyserver keyserver.ubuntu.com --recv-key 90D2B445 \
+    && gpg --keyserver keyserver.ubuntu.com --recv-key 78086003 \
```
xcoulon: just curious, why do we have to do this? In https://github.com/fabric8-services/fabric8-wit/blob/master/Makefile#L269 we had a more straightforward process. Is this one better, maybe? (asking to understand)

alexeykazakov: We don’t have the checksum for operator-sdk. They use asc file and gpg public key for verification instead. See https://github.com/operator-framework/operator-sdk/blob/master/doc/user/install-operator-sdk.md

xcoulon: ok, thanks for the clarification, @alexeykazakov

alexeykazakov:
> weird, I thought that go 1.13 is installed via yum :-/ we had an issue with that some time ago and we also verified it.

Yeah right.. I remember that too. But as you mentioned we don't really control what is being installed via yum and it definitely cause issues. We can/should revisit this later but for now downloading golang form google unblocks us.

PR #41, opened 2019-11-08
chore: don't clean cluster after successful test

MatousJobanek:
* removes `test-e2e-keep-namespaces` as it's not needed anymore
* don't call `clean-e2e-resources` target at the end of the e2e test and just print info
* use a unique name for `catalogsource` and `configmap`
* store the final `deploy_csv.yaml` and `install_operator.yaml` files in `/tmp` directory for better debugging


make/test.mk
```diff
@@ -264,18 +263,21 @@ ifeq ($(REPO_NAME),registration-service)
...
+		sed -e 's|REPLACE_NAMESPACE|${NAMESPACE}|g;s|^  source: .*|&-${DATE_SUFFIX}|' ${E2E_REPO_PATH}/hack/install_operator.yaml > /tmp/${REPO_NAME}_install_operator_${DATE_SUFFIX}.yaml
+		cat /tmp/${REPO_NAME}_install_operator_${DATE_SUFFIX}.yaml | oc apply -f -
 		while [[ -z `oc get sa ${REPO_NAME} -n ${NAMESPACE} 2>/dev/null` ]] || [[ -z `oc get crd kubefedclusters.core.kubefed.io 2>/dev/null` ]]; do \
 			if [[ $${NEXT_WAIT_TIME} -eq 300 ]]; then \
+			   CATALOGSOURCE_NAME=`oc get catalogsource --output=name -n openshift-marketplace | grep "codeready-toolchain-saas.*${DATE_SUFFIX}"`; \
+			   SUBSCRIPTION_NAME=`oc get subscription --output=name -n ${NAMESPACE} | grep "codeready-toolchain-saas"`; \
 			   echo "reached timeout of waiting for ServiceAccount ${REPO_NAME} to be available in namespace ${NAMESPACE} and CRD kubefedclusters.core.kubefed.io to be available in the cluster - see following info for debugging:"; \
 			   echo "================================ CatalogSource =================================="; \
-			   oc get ${CATALOGSOURCE_NAME} -n openshift-marketplace -o yaml; \
+			   oc get $${CATALOGSOURCE_NAME} -n openshift-marketplace -o yaml; \
```
xcoulon: just curious: why do you need to use `$${CATALOGSOURCE_NAME}` instead of `${CATALOGSOURCE_NAME}`?

make/test.mk
```diff
@@ -264,18 +263,21 @@ ifeq ($(REPO_NAME),registration-service)
...
 			   echo "================================ CatalogSource Pod Logs =================================="; \
-			   CATALOGSOURCE_NAME=${CATALOGSOURCE_NAME}; \
-			   oc logs `oc get pods -n openshift-marketplace -o name | grep $${CATALOGSOURCE_NAME#*/}` -n openshift-marketplace; \
+			   oc logs `oc get pods -l "olm.catalogSource=$${CATALOGSOURCE_NAME#*/}" -n openshift-marketplace -o name` -n openshift-marketplace; \
```
xcoulon: is this syntax correct? `$${CATALOGSOURCE_NAME#*/}`

make/test.mk
MatousJobanek: because it's inside of the bash script - If it contained only one dollar sign that it would be replaced by makefile, which would mean that it would be empty because that variable is not set inside of the makefile

MatousJobanek: I wanted to write that the declaration of the variable is inside of the bash script and then used in the same script. Originally, it was makefile variable, but for the simplicity I moved it to the script logic

make/test.mk
MatousJobanek: yes it is - two dollar signs are because it's inside of the bash script - see the explanation above.
the suffix `#*/` ensures that it takes the string after the slash. The `CATALOGSOURCE_NAME` variable contains:
`catalogsource.operators.coreos.com/codeready-toolchain-saas-toolchain-member-operator-1573220355`
and I need only the second part - the name

PR #46, opened 2019-11-15
Allow individual tests override wait timeouts and retry intervals 

alexeykazakov:
For example if we need to override the default timeout:
```
s.hostAwait.WithRetryOptions(TimeoutOption(time.Second * 10)).WaitForMasterUserRecord(...)
```

It's useful in test which wait for some resource not to be created and we don't want to wait for the entire default 120 seconds in such tests.

### DIFF
--- a/wait/awaitility.go
+++ b/wait/awaitility.go
@@ -33,8 +33,6 @@ const (
 	KubeFedClusterConditionTimeout = 10*defaults.DefaultClusterHealthCheckPeriod + 5*time.Second
 )
 
-type OperatorRetryIntervalOption time.Duration
-type OperatorTimeoutOption time.Duration
 type RetryIntervalOption time.Duration
 type TimeoutOption time.Duration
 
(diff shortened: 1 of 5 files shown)


wait/awaitility.go
```diff
@@ -54,10 +60,27 @@ type SingleAwaitility interface {
...
-	T               *testing.T
-	Client          framework.FrameworkClient
-	Ns              string
-	OtherOperatorNs string
+	T                     *testing.T
+	Client                framework.FrameworkClient
+	Ns                    string
+	OtherOperatorNs       string
+	OperatorRetryInterval time.Duration
+	OperatorTimeout       time.Duration
```
MatousJobanek: why do we have `OperatorRetryInterval` and `OperatorTimeout` when they are not used anywhere?

alexeykazakov: Good question :)
Let me double check and clean them up.

alexeykazakov: Done in https://github.com/codeready-toolchain/toolchain-e2e/pull/46/commits/44cc75736a74e317729701874ce1ed520d54208a

alexeykazakov:
I also did some cleanup. Removed the garbage collection test from UserSignup tests because it's already covered in the overall e2e_test.go.

PR #47, opened 2019-11-15
Cleanup UserSignup tests

alexeykazakov:


test/e2e/conditions.go
```diff
@@ -6,10 +6,53 @@ import (
...
 	return toolchainv1alpha1.Condition{
 		Type:   toolchainv1alpha1.ConditionReady,
 		Status: corev1.ConditionTrue,
 		Reason: "Provisioned",
 	}
 }
+
+func pendingApproval() []toolchainv1alpha1.Condition {
+	return []toolchainv1alpha1.Condition{
+		toolchainv1alpha1.Condition{
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
		{
```

alexeykazakov: Fixed in https://github.com/codeready-toolchain/toolchain-e2e/pull/47/commits/614901a87742d8804179991af671fe04ef8adf9d

test/e2e/conditions.go
```diff
@@ -6,10 +6,53 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
-func isProvisioned() toolchainv1alpha1.Condition {
+func provisioned() toolchainv1alpha1.Condition {
```
xcoulon: what's wrong with `isProvisioned`?

test/e2e/e2e_test.go
```diff
@@ -162,65 +163,31 @@ func TestE2EFlow(t *testing.T) {
...
 }
 
-func setup(ctx *framework.TestCtx, awaitility *wait.Awaitility, username string) (*toolchainv1alpha1.UserSignup, *toolchainv1alpha1.UserAccountSpec, error) {
+func setup(ctx *framework.TestCtx, awaitility *wait.Awaitility, username string) (*toolchainv1alpha1.UserSignup, *toolchainv1alpha1.UserAccountSpec) {
 	// 0. Verify that the `basic` NSTemplateTier resource exists (will be needed later)
 	revisions, err := getRevisions(awaitility)
-	if err != nil {
-		return nil, nil, err
-	}
+	require.NoError(awaitility.T, err)
```
xcoulon: I think we shouldn;t do that, because the `T` in `awaitility.T` refers to the parent `t` and in case of error, it may cause a panic

test/e2e/usersignup_test.go
```diff
@@ -45,662 +43,190 @@ func (s *userSignupIntegrationTest) TearDownTest() {
...
-				Type:   v1alpha1.UserSignupApproved,
-				Status: corev1.ConditionTrue,
-				Reason: "ApprovedByAdmin",
-			},
-			v1alpha1.Condition{
-				Type:   v1alpha1.UserSignupComplete,
-				Status: corev1.ConditionTrue,
-			}))
-		require.NoError(s.T(), err)
+		s.T().Run("approved set to false", func(t *testing.T) {
```
xcoulon:
```suggestion
		t.Run("approved set to false", func(t *testing.T) {
```

test/e2e/usersignup_test.go
```diff
@@ -45,662 +43,190 @@ func (s *userSignupIntegrationTest) TearDownTest() {
...
+			// UserSignup created with approved flag set to false but it should be approved anyway
+			s.createUserSignupAndAssertAutoApproval(false)
+		})
 
-		// Confirm the MUR was created
-		s.assertCreatedMUR(userSignup)
+		s.T().Run("approved set to true", func(t *testing.T) {
```
xcoulon:
```suggestion
		t.Run("approved set to true", func(t *testing.T) {
```

test/e2e/usersignup_test.go
```diff
@@ -45,662 +43,190 @@ func (s *userSignupIntegrationTest) TearDownTest() {
...
-		v1alpha1.Condition{
-			Type:    v1alpha1.UserSignupComplete,
-			Status:  corev1.ConditionFalse,
-			Reason:  "InvalidMURState",
-			Message: "multiple matching MasterUserRecord resources found",
-		}))
-	require.NoError(s.T(), err)
-}
+func (s *userSignupIntegrationTest) checkUserSignupManualApproval() {
+	s.T().Run("usersignup created first then manually approved", func(t *testing.T) {
```
MatousJobanek: I guess, that the same problem with the parent `t` being nill we can face here as well.

test/e2e/usersignup_test.go
```diff
@@ -45,662 +43,190 @@ func (s *userSignupIntegrationTest) TearDownTest() {
...
-	err = s.awaitility.Client.Create(context.TODO(), userSignup, testsupport.CleanupOptions(s.testCtx))
+func (s *userSignupIntegrationTest) createUserSignupAndAssertPendingApproval() *v1alpha1.UserSignup {
+	// Create a new UserSignup with approved flag set to false
+	id := uuid.NewV4().String()
+	username := "testuser" + id
+	userSignup := newUserSignup(s.T(), s.awaitility.Host(), id, username)
 
+	err := s.awaitility.Client.Create(context.TODO(), userSignup, testsupport.CleanupOptions(s.testCtx))
 	require.NoError(s.T(), err)
 	s.T().Logf("user signup '%s' created", userSignup.Name)
```
MatousJobanek: And in this function too

test/e2e/usersignup_test.go
```diff
@@ -45,662 +43,190 @@ func (s *userSignupIntegrationTest) TearDownTest() {
...
-	// Set the user approval policy to automatic
+	// Create user signup
 	s.setApprovalPolicyConfig("automatic")
-
-	// Create user signup - no approval set
-	s.T().Logf("Creating UserSignup with namespace %s", s.namespace)
-
-	userSignup, err := newUserSignup(s.awaitility.Host(), uuid.NewV4().String(), "reginald@alpha.com")
-	require.NoError(s.T(), err)
+	userSignup := newUserSignup(s.T(), s.awaitility.Host(), uuid.NewV4().String(), "reginald@alpha.com")
```
MatousJobanek: I see that in all cases when  `newUserSignup` function is called it generates the UUID every time in the same way `uuid.NewV4().String()` - could we move it directly into the function `newUserSignup`?

test/e2e/conditions.go
alexeykazakov: 1. `isProvisioned()` doesn't return a bool.
2. `provisioned()` looks better in the constructions where it's used. For example: `wait.UntilMasterUserRecordHasConditions(provisioned())`

It doesn't have to be a function btw. Could be a const. But IMO it's minor.

test/e2e/usersignup_test.go
```diff
@@ -45,662 +43,190 @@ func (s *userSignupIntegrationTest) TearDownTest() {
...
+		s.T().Run("approved set to false", func(t *testing.T) {
+			// UserSignup created with approved flag set to false but it should be approved anyway
+			s.createUserSignupAndAssertAutoApproval(false)
+		})
 
-		// Confirm the MUR was created
-		s.assertCreatedMUR(userSignup)
+		s.T().Run("approved set to true", func(t *testing.T) {
+			// UserSignup created with approved flag set to true and it should be approved
+			s.createUserSignupAndAssertManualApproval(true)
```
MatousJobanek:
```suggestion
			s.createUserSignupAndAssertAutoApproval(true)
```

test/e2e/usersignup_test.go
```diff
@@ -45,662 +43,190 @@ func (s *userSignupIntegrationTest) TearDownTest() {
...
-func (s *userSignupIntegrationTest) TestUserSignupWithAutoApprovalNoApprovalSet() {
-	// Set the user approval policy to automatic
-	s.setApprovalPolicyConfig("automatic")
-
-	// Create user signup - no approval set
-	s.T().Logf("Creating UserSignup with namespace %s", s.namespace)
+func (s *userSignupIntegrationTest) TestTransformUsername() {
+	// Create UserSignup with a username that we don't need to transform
+	userSignup, _ := s.createUserSignupWithNameAndAssertManualApproval(true, "paul-no-need-to-transform")
+	require.Equal(s.T(), "paul-no-need-to-transform", userSignup.Status.CompliantUsername)
```
MatousJobanek: To be honest, this test looks like a unit test. The only thing that is verified here is that the compliant username is set in the status of the usersignup - the rest is verified in other test cases. I believe that if we use verify the generation of the  compliant username in the previous test `TestTargetClusterSelectedAutomatically`, then we should be fine and we can drop this entire test

test/e2e/e2e_test.go
alexeykazakov: awaitility.T refers to the same t actually. But even if it didn't (but it does) I thought that testify now properly handle the parent-child T situation.

test/e2e/usersignup_test.go
alexeykazakov: No, no. In this case we will be using "manual" approval anyway. The `Spec.Approved` set to true. So, it doesn't matter if auto-approval is enabled or not. So, the status reason will be set to "ApprovedByAdmin". Not "ApprovedAutomatically". So, the check is correct here.

test/e2e/usersignup_test.go
alexeykazakov: It does test more. Not just the fact that compliant username is set. It tests all possible combinations:
- username == compliant username
- there is already a username with such compliant username
- there are multiple matches (so we should keep increasing the number)

We could merge it with other tests but it would make the test code a bit less readable IMO.

test/e2e/usersignup_test.go
MatousJobanek: ah, I see. It was very confusing that it's inside of the `"auto approval policy"` subtests, but using checks for manual approval - I didn't notice the different resulting status...

test/e2e/e2e_test.go
xcoulon: testify handles the parent-child change of T in https://github.com/stretchr/testify/pull/655/files#diff-da9b094bf431c6cb21dbc8467bc17c17R74 and you can see that `T` is set to the subtests' `t` in line 74. Is it the case for `awaitility`?

test/e2e/usersignup_test.go
alexeykazakov: Let me maybe add a comment to the test so it's clear why we still assert "manual" approval in auto-approval test :)

MatousJobanek: btw, could we merge the three functions:
`createUserSignupAndAssertManualApproval`
`createUserSignupWithNameAndAssertManualApproval`
`createUserSignupAndAssertAutoApproval`
into one:
```go
func (s *userSignupIntegrationTest) createAndCheckUserSignup(specApproved bool, conditions ...v1alpha1.Condition) *v1alpha1.UserSignup {
	// Create a new UserSignup with the given approved flag
	userSignup := newUserSignup(s.T(), s.awaitility.Host(), uuid.NewV4().String(), "testuser"+uuid.NewV4().String())
	userSignup.Spec.Approved = specApproved
	err := s.awaitility.Client.Create(context.TODO(), userSignup, testsupport.CleanupOptions(s.testCtx))
	require.NoError(s.T(), err)
	s.T().Logf("user signup '%s' created", userSignup.Name)

	// Check the UserSignup is approved now
	userSignup, err = s.hostAwait.WaitForUserSignup(userSignup.Name, wait.UntilUserSignupHasConditions(conditions...))
	require.NoError(s.T(), err)

	return userSignup
}
```
and then use it:
```go
s.T().Run("approved set to false", func(t *testing.T) {
	// UserSignup created with approved flag set to false but it should be approved anyway
	userSignup := s.createAndCheckUserSignup(false, approvedAutomatically()...)
	// Confirm the MUR was created
	s.assertCreatedMUR(userSignup)
})

s.T().Run("approved set to true", func(t *testing.T) {
	// UserSignup created with approved flag set to true and it should be approved
	userSignup := s.createAndCheckUserSignup(true, approvedByAdmin()...)
	// Confirm the MUR was created
	s.assertCreatedMUR(userSignup)
})
```
and in the `createUserSignupAndAssertPendingApproval` in this way:
```go
func (s *userSignupIntegrationTest) createUserSignupAndAssertPendingApproval() *v1alpha1.UserSignup {
	// Create a new UserSignup with approved flag set to false
	userSignup := s.createAndCheckUserSignup(false, pendingApproval()...)

	// Confirm the CompliantUsername has NOT been set
	require.Empty(s.T(), userSignup.Status.CompliantUsername)

	// Confirm that a MasterUserRecord wasn't created
	_, err := s.hostAwait.WithRetryOptions(wait.TimeoutOption(time.Second * 10)).WaitForMasterUserRecord(userSignup.Spec.Username)
	require.Error(s.T(), err)
	return userSignup
}
```
it would make the code shorter.

test/e2e/e2e_test.go
alexeykazakov: Let's move the awaitility.T discussion to a separate PR. For now I just added t param to setup().

alexeykazakov: https://github.com/codeready-toolchain/toolchain-e2e/pull/47/commits/4f1fbb01a91236fb4977a72435d1b0d7bd84864c

test/e2e/usersignup_test.go
alexeykazakov: Yeah... we could for sure.
Done in https://github.com/codeready-toolchain/toolchain-e2e/pull/47/commits/c99f78bfe01012cd5b9a996765a5f81c8e883b1e

test/e2e/usersignup_test.go
alexeykazakov: We can't auto-generate the username always (we need to use specific ones in username transformation tests) but I made the code shorter: https://github.com/codeready-toolchain/toolchain-e2e/pull/47/commits/381ca0d949494b7a7872c0e82539637b31fd5cbb

alexeykazakov:
@xcoulon @MatousJobanek thanks for the review! I've addressed your comments. Please review :)

test/e2e/conditions.go
xcoulon: ok, I see what you mean. I still believe that `isProvisioned` looks better, but it's cosmetic.

test/e2e/e2e_test.go
xcoulon: cool, thanks for taking this request into account, @alexeykazakov

xcoulon:
> @xcoulon @MatousJobanek thanks for the review! I've addressed your comments. Please review :)

thanks for taking my changes into account. I'm going to approve it, but let's wait for @MatousJobanek's review before merging it.

test/e2e/usersignup_test.go
MatousJobanek: Sure, when I was writing it I hoped that we could drop the transformation tests. Anyway, I see that we need it so passing the username is just one parameter in the methods. 
I still believe that we could consolidate it more according to my proposal (drop some methods and make `createUserSignupAndAssertPendingApproval` shorter), but your changes are sufficient.

alexeykazakov:
Thanks @xcoulon @MatousJobanek for the review!

PR #50, opened 2019-11-19
makefile goal to display the latest e2e project names

xcoulon:
The goal of `display-eval` is to provide a convenient way to retrieve the latest host and member operator namespaces on a cluster, and optionally, to export them as env vars:
```
$ make display-eval
export HOST_NS=host-operator-1574156115
export MEMBER_NS=member-operator-1574156115
export REGISTRATION_SERVICE_NS=$HOST_NS
# Run this command to configure your shell:
# eval $(make display-eval)
```

The `host-operator-xxx` and `member-operator-xxx` projects are now labelled with `app=host-operator` and `app=member-operator` respectively the latest occurrence is retained. 
By querying the cluster, we can use this goal alone, without interfering with the `HOST_NS` and `MEMBER_NS` variables set in the `test.mk` file.

### DIFF
--- a/make/test.mk
+++ b/make/test.mk
@@ -189,6 +189,7 @@ ifeq ($(MEMBER_REPO_PATH),)
 endif
 	@echo "Deploying member operator to $(MEMBER_NS)..."
 	-oc new-project $(MEMBER_NS) 1>/dev/null
+	-oc label ns $(MEMBER_NS) app=member-operator
 	-oc project $(MEMBER_NS)
 ifneq ($(IS_OS_3),)
 	oc apply -f ${MEMBER_REPO_PATH}/deploy/service_account.yaml
@@ -207,6 +208,7 @@ ifeq ($(HOST_REPO_PATH),)
 endif
 	@echo "Deploying host operator to $(HOST_NS)..."
 	-oc new-project $(HOST_NS) 1>/dev/null
+	-oc label ns $(MEMBER_NS) app=host-operator
 	-oc project $(HOST_NS)
 ifneq ($(IS_OS_3),)
 	# is using OS 3, so we need to deploy the manifests manually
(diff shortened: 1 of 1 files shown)


make/test.mk
```diff
@@ -304,3 +307,11 @@ else
 		sed -e 's|REPLACE_IMAGE|${IMAGE_NAME}|g' ${E2E_REPO_PATH}/deploy/operator.yaml | oc apply -f -
     endif
 endif
+
+.PHONY: display-eval
+display-eval:
+	@echo 'export HOST_NS=$(shell oc get projects -l app=host-operator --output=name --no-headers=true -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}{end}' | sort | tail -n 1)'
```
MatousJobanek: I think that you could make it shorter and drop the jsonpath as well as the --no-headers param:
```suggestion
	@echo 'export HOST_NS=$(shell oc get projects -l app=host-operator --output=name  | cut -f2 -d"/" | sort | tail -n 1)'
```
but it's only my proposal - you don't have to use it as your solution is fine too ;-)

make/test.mk
```diff
@@ -304,3 +307,11 @@ else
...
+	@echo 'export MEMBER_NS=$(shell oc get projects -l app=member-operator --output=name --no-headers=true -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}{end}' | sort | tail -n 1)'
```
MatousJobanek:
```suggestion
	@echo 'export MEMBER_NS=$(shell oc get projects -l app=member-operator --output=name | cut -f2 -d"/" | sort | tail -n 1)'
```

make/test.mk
xcoulon: well, I kinda like the jsonpath expression, so I'll keep it as-is for now if it's ok for you ;)

make/test.mk
xcoulon: same here, keeping the `jsonpath` for now, but indeed, the `--no-headers=true` is not needed, so I'll remove it.

xcoulon: done in 689670f

make/test.mk
xcoulon: removed the `--no-headers=true` arg in 689670f

xcoulon:
> WOW, this is very nice improvement 👍 🥇

thanks @MatousJobanek!

PR #52, opened 2019-11-26
Removed workaround for token timestamp issue

sbryzak:
Signed-off-by: Shane Bryzak <sbryzak@gmail.com>


test/e2e/registration_service_test.go
```diff
@@ -123,7 +123,7 @@ func (s *registrationServiceTestSuite) TestAuthConfig() {
 func (s *registrationServiceTestSuite) TestSignupFails() {
 	identity0 := authsupport.NewIdentity()
 	emailClaim0 := authsupport.WithEmailClaim(uuid.NewV4().String() + "@email.tld")
-	iatClaim0 := authsupport.WithIATClaim(time.Now().Add(-60 * time.Second))
+	//iatClaim0 := authsupport.WithIATClaim(time.Now().Add(-60 * time.Second))
```
tinakurian: maybe we can just remove this line as well?

tinakurian:
I guess we would need to update the common version to use this version: https://github.com/codeready-toolchain/toolchain-common/pull/63 that way we can see if the e2e-tests pass

PR #54, opened 2019-12-03
Remove deprecated TargetCluster from MUR status

alexeykazakov:
This PR is purely to verify that the changes in API and common do not break the member operator:
https://github.com/codeready-toolchain/api/pull/66
https://github.com/codeready-toolchain/toolchain-common/pull/65

When the PR from above are merged I will remove the replace and update the api and common dependencies properly.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -42,4 +42,9 @@ replace (
 
 replace github.com/Azure/go-autorest => github.com/Azure/go-autorest v13.0.0+incompatible
 
+replace (
+	github.com/codeready-toolchain/api => github.com/alexeykazakov/api v0.0.0-20191203022256-57871e5e6c39
+	github.com/codeready-toolchain/toolchain-common => github.com/alexeykazakov/toolchain-common v0.0.0-20191203023806-0035b5057ff0
+)
+
 go 1.13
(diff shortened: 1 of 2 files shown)


alexeykazakov:
> so shouldn't you make PRs to this repo and to the host repo from a branch with the same name, but a different one than to the other repositories? To ensure that the e2e tests are paired only with the PR in host repo... If I understand it correctly, you wanted to verify that the changes in api and common repos that are transitively available in the host repo PR will not break the other repositories (reg-service and member), correct? In that case, we should pair only with the PR in the host repo, shouldn't we?

Changes in the member and reg-service PRs are just about making sure that the code still compiles tbh and that I didn't miss anything :) There is no any code changes in these repos. So, I thought it was OK if the e2e test are paired with all the PRs from all repos.

MatousJobanek:
> Changes in the member and reg-service PRs are just about making sure that the code still compiles tbh and that I didn't miss anything :) There is no any code changes in these repos. So, I thought it was OK if the e2e test are paired with all the PRs from all repos.

Ah, right. It just updates the dependencies which means that it's almost the same as the master is. Yeah, that's fine ;-)

PR #56, opened 2019-12-06
feat(CRT-354): let host-operator deploy the registration service

MatousJobanek:
based on the changes proposed in this PR https://github.com/codeready-toolchain/host-operator/pull/117 it changes the way the registration-service is being deployed. As the deployment is moved to host-operator, the e2e setup only builds the image and sets the appropriate host deployment params

### DIFF
--- a/testsupport/init.go
+++ b/testsupport/init.go
@@ -53,7 +53,7 @@ func WaitForDeployments(t *testing.T, obj runtime.Object) (*framework.TestCtx, *
 	require.NoError(t, err, "failed while waiting for member operator deployment")
 
 	// wait for registration service to be ready
-	err = e2eutil.WaitForDeployment(t, f.KubeClient, registrationServiceNs, "registration-service", 3, wait.DefaultRetryInterval, wait.DefaultOperatorTimeout)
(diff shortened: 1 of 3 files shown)


testsupport/init.go
```diff
@@ -53,7 +53,7 @@ func WaitForDeployments(t *testing.T, obj runtime.Object) (*framework.TestCtx, *
 	require.NoError(t, err, "failed while waiting for member operator deployment")
 
 	// wait for registration service to be ready
-	err = e2eutil.WaitForDeployment(t, f.KubeClient, registrationServiceNs, "registration-service", 3, wait.DefaultRetryInterval, wait.DefaultOperatorTimeout)
+	err = e2eutil.WaitForDeployment(t, f.KubeClient, registrationServiceNs, "registration-service", 1, wait.DefaultRetryInterval, wait.DefaultOperatorTimeout)
```
alexeykazakov: I think we need to change it two at least 2 replicas: https://github.com/codeready-toolchain/host-operator/blob/master/deploy/env/e2e-tests.yaml#L3
So we can have our tests with multiple reg-service replicas to make verify it's a stateless service and there is no any issues with running multiple replicas.

We can do it in separate PRs though.

MatousJobanek: good point. But I can do it easily here. it's as simple as these two commits:
0483817 
https://github.com/codeready-toolchain/host-operator/pull/117/commits/0610d667fdc411e1eb974c98454f4ccc765771f2

PR #71, opened 2020-01-28
chore: use digest instead of tags

MatousJobanek:
To properly verify that the changes introduced here https://github.com/codeready-toolchain/api/commit/d37dc262950f4345055d85c3d7e2aa8ba88c819a are working properly (in other words, that CSV and OLM can use digests instead of tags) use digests instead of tags in e2e tests too

### DIFF
--- a/make/test.mk
+++ b/make/test.mk
@@ -271,6 +271,7 @@ endif
 
 .PHONY: build-operator
 build-operator:
+	mkdir ${IMAGE_NAMES_DIR} || true
 # when e2e tests are triggered from different repo - eg. as part of PR in host-operator repo - and the image of the operator is (not) provided
 ifeq ($(SET_IMAGE_NAME),)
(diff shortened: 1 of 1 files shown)


alexeykazakov:
> docker pull registry.svc.ci.openshift.org/ci-op-4m5xk0zy/stable:registration-service
/bin/bash: docker: command not found

MatousJobanek:
> docker pull registry.svc.ci.openshift.org/ci-op-4m5xk0zy/stable:registration-service
/bin/bash: docker: command not found

I know, I had to switch to API call: 818dadb 
I will do it also in the scripts ;-)

MatousJobanek:
and yeah, it's doing only for `quay.io`, which means that it's using it only for local testing, unfortunately, there is no other easy way. 
I can try to improve it later, but for now the local verification should be enough

PR #79, opened 2020-03-09
feat: verify that ChangeTierRequest changes the tier

MatousJobanek:
* modifies tests covering tier promotions so it uses the ChangeTierRequest CRs
* move the promotion-tier related tests to `nstemplatetier_test.go` file
* tests in `nstemplatetier_test.go` are now more generic so we won't miss a newly created tiers
* introduced `tiers` package that contains:
   * `checks.go` - for every tier there is an implementation of `TierChecks` that returns:
      * specific check functions that verify the existence and non-existence of inner objects
      * expected set of namespace types and revisions
   * `nstemplateset.go` - with the help of the checks it verifies that NsTemplateSet and namespaces contain the correct resources and configuration
   * `revisions.go` - moved `GetRevisions` function

paired PR: https://github.com/codeready-toolchain/host-operator/pull/160


test/e2e/e2e_test.go
```diff
@@ -229,6 +247,34 @@ func TestE2EFlow(t *testing.T) {
 	})
 }
 
+func newChangeTierRequest(namespace, tier, murName string) *toolchainv1alpha1.ChangeTierRequest {
+	return &toolchainv1alpha1.ChangeTierRequest{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: namespace,
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
			Namespace:    namespace,
```

wait/host.go
```diff
@@ -225,3 +225,43 @@ func HasNamespaceRevisions(r string) NSTemplateTierSpecMatcher {
...
+			if errors.IsNotFound(err) {
+				a.T.Logf("ChangeTierRequest is checked as deleted '%s'", name)
+				return true, nil
+			}
+			return false, err
+		}
+		a.T.Logf("waiting until ChangeTierRequest is deleted '%s'", name)
+		return false, nil
+	})
+}
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
}
```

tiers/checks.go
```diff
@@ -0,0 +1,157 @@
...
+	v1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	providerMatchingLabels = client.MatchingLabels(map[string]string{"provider": "codeready-toolchain"})
+)
+
+func NewTierChecks(tier string) (TierChecks, error) {
```
xcoulon: or just `NewChecks` ?

tiers/checks.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+func (a *basicTierChecks) GetExpectedRevisions(awaitility *wait.Awaitility) Revisions {
+	revisions, err := GetRevisions(awaitility, "basic", "code", "dev", "stage")
+	require.NoError(awaitility.T, err)
+	return revisions
+}
+
+type advancedTierChecks struct {
+	*basicTierChecks
```
xcoulon: it's a big weird to have `advancedTierChecks` composed of `*basicTierChecks`. What about having a `baseTierChecks` or something similar, instead?

tiers/checks.go
```diff
@@ -0,0 +1,157 @@
...
+func (a *advancedTierChecks) GetExpectedRevisions(awaitility *wait.Awaitility) Revisions {
+	revisions, err := GetRevisions(awaitility, "advanced", "code", "dev", "stage")
+	require.NoError(awaitility.T, err)
+	return revisions
+}
+
+type teamTierChecks struct {
+}
+
+func (a *teamTierChecks) GetInnerObjectChecks(nsType string) []innerObjectCheck {
```
xcoulon: do you need a pointer receiver?

tiers/revisions.go
```diff
@@ -0,0 +1,38 @@
...
+	toolchainv1alpha1 "github.com/codeready-toolchain/api/pkg/apis/toolchain/v1alpha1"
+	"github.com/codeready-toolchain/toolchain-e2e/wait"
+	"github.com/stretchr/testify/require"
+)
+
+type Revisions map[string]string
+
+func GetRevisions(awaitility *wait.Awaitility, tier string, nsTypes ...string) (Revisions, error) {
+	templateTier, err := awaitility.Host().WaitForNSTemplateTier(tier, wait.UntilNSTemplateTierSpec(wait.Not(wait.HasNamespaceRevisions("000000a"))))
+	if err != nil {
```
xcoulon: what about `required.NoError(awaitility.T, err)` instead?

tiers/checks.go
alexeykazakov: We don't have to use a pointer receiver in this case but it doesn't hurt, does it? :)
I personally prefer to use pointer receivers since it's IMO less error prone. Is there any disadvantages of using them?

tiers/checks.go
MatousJobanek: why not - changed in 959f24a

tiers/checks.go
MatousJobanek: you're right. This should be better: 1c5633f

tiers/checks.go
MatousJobanek: yes, it's my habit to use a pointer receiver in all methods and not use it only when there is a reason for it.

tiers/revisions.go
MatousJobanek: yeah, could be - it was copy-pasted from the previous location :-) 
changed in c7e22d6

PR #84, opened 2020-03-26
chore: verify role name rbac-edit

MatousJobanek:
verify that we use `rbac-edit` role name instead of `toolchain-*-edit` name

paired PR: https://github.com/codeready-toolchain/host-operator/pull/166

### DIFF
--- a/tiers/checks.go
+++ b/tiers/checks.go
@@ -62,8 +62,8 @@ func (a *basicTierChecks) GetInnerObjectChecks(nsType string) []innerObjectCheck
 func getDefaultChecks(nsType string) []innerObjectCheck {
 	if nsType == "code" {
 		return append(commonChecks,
-			toolchainUserEditRole("toolchain-che-edit"),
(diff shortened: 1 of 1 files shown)


tiers/checks.go
```diff
@@ -122,22 +121,22 @@ func userEditRoleBinding() innerObjectCheck {
 	}
 }
 
-func toolchainUserEditRole(roleName string) innerObjectCheck {
+func rbacEditRoleBinding() innerObjectCheck {
```
xcoulon: wouldn't it be more flexible if we keep the resource name as a parameter of these funcs?

MatousJobanek: Since we verify also the content of the resources, then I don't think that we could reuse it - I don't expect that we would have multiple resources with the same content but with different names.

MatousJobanek: And that is exactly what I'm trying to fix in these two PRs - have the resource names consistent

xcoulon: ok

PR #94, opened 2020-04-03
chore: verify new transformation of compliant username

MatousJobanek:
paired PR: https://github.com/codeready-toolchain/host-operator/pull/176


MatousJobanek:
good catch @alexeykazakov :+1: 
removed in ecedcb1 and I added it also to .gitignore

PR #95, opened 2020-04-03
feat(CRT-459): add cluster resource quotas on user namespaces

xcoulon:
E2E tests in relation with https://github.com/codeready-toolchain/member-operator/pull/143

### DIFF
--- a/make/clean.mk
+++ b/make/clean.mk
@@ -2,3 +2,11 @@
 clean:
 	$(Q)-rm -rf ${V_FLAG} $(OUT_DIR) ./vendor
 	$(Q)go clean ${X_FLAG} ./...
+
+.PHONY: clean-resources
+clean-resources:
+	$(Q)echo "cleaning resources in $(HOST_NS) and $(MEMBER_NS)..."
+	$(Q)oc get usersignups -n $(HOST_NS) -o name | xargs oc delete -n $(HOST_NS)
+	$(Q)oc get pods -n $(HOST_NS) -l name=host-operator -o name | xargs oc delete -n $(HOST_NS)
+	$(Q)oc get pods -n $(MEMBER_NS) -l name=member-operator -o name | xargs oc delete -n $(MEMBER_NS)
+	$(Q)oc get clusterresourcequotas -l "toolchain.dev.openshift.com/provider"=codeready-toolchain -o name| xargs oc delete
--- a/make/clean.mk
+++ b/make/clean.mk
@@ -1,12 +1,37 @@
 .PHONY: clean
+## Removes vendor directory and runs go clean command
 clean:
 	$(Q)-rm -rf ${V_FLAG} $(OUT_DIR) ./vendor
 	$(Q)go clean ${X_FLAG} ./...
 
-.PHONY: clean-resources
-clean-resources:
-	$(Q)echo "cleaning resources in $(HOST_NS) and $(MEMBER_NS)..."
-	$(Q)oc get usersignups -n $(HOST_NS) -o name | xargs oc delete -n $(HOST_NS)
-	$(Q)oc get pods -n $(HOST_NS) -l name=host-operator -o name | xargs oc delete -n $(HOST_NS)
-	$(Q)oc get pods -n $(MEMBER_NS) -l name=member-operator -o name | xargs oc delete -n $(MEMBER_NS)
-	$(Q)oc get clusterresourcequotas -l "toolchain.dev.openshift.com/provider"=codeready-toolchain -o name| xargs oc delete
+.PHONY: clean-e2e-resources
+## Deletes resources in the OpenShift cluster. The deleted resources are:
+##    * namespaces created during both the dev and e2e test setup (for both operators host and member)
+##    * all CatalogSources that were created as part of operator deployment
+clean-e2e-resources:
+	$(Q)-oc get projects --output=name | grep -E "${QUAY_NAMESPACE}-(toolchain\-)?(member|host)(\-operator)?(\-[0-9]+)?|${QUAY_NAMESPACE}-toolchain\-e2e\-[0-9]+" | xargs oc delete
+	$(Q)-oc get catalogsource --output=name -n openshift-marketplace | grep "source-toolchain-.*${QUAY_NAMESPACE}" | xargs oc delete -n openshift-marketplace
+
+.PHONY: clean-e2e-files
+## Removes files and directories used during e2e test setup
+clean-e2e-files:
+	rm -f ${WAS_ALREADY_PAIRED_FILE} 2>/dev/null || true
+	rm -rf ${IMAGE_NAMES_DIR} 2>/dev/null || true
+
+.PHONY: clean-toolchain-resources-in-dev
+## Deletes resources in the OpenShift cluster. The deleted resources are:
+##    * all toolchain.dev.openshift.com CRs in both host and member namespaces created during the dev setup
+##    * all ClusterresourceQuotas with the label toolchain.dev.openshift.com/provider=codeready-toolchain in member namespace
+##    * pods of both operators host and member
+clean-toolchain-resources-in-dev:
+	$(Q)echo "cleaning resources in $(DEV_HOST_NS) and $(DEV_MEMBER_NS)..."
+	$(Q)oc get usersignups -n $(DEV_HOST_NS) -o name | xargs -i oc delete -n $(DEV_HOST_NS) {}
+	$(Q)for CRD in `oc get crd -o name | grep toolchain`; do \
+		echo "deleting $${CRD} in $(DEV_HOST_NS) and $(DEV_MEMBER_NS)"; \
+		CRD_NAME=`oc get $${CRD} --template '{{.metadata.name}}'`; \
+		oc get $${CRD_NAME} -n $(DEV_HOST_NS) -o name | xargs -i oc delete {}; \
+		oc get $${CRD_NAME} -n $(DEV_MEMBER_NS) -o name | xargs -i oc delete {}; \
+	done
+	$(Q)oc get clusterresourcequotas -l "toolchain.dev.openshift.com/provider"=codeready-toolchain -o name -n $(DEV_MEMBER_NS) | xargs -i oc delete {}
+	$(Q)oc get pods -n $(DEV_HOST_NS) -l name=host-operator -o name | xargs -i oc delete -n $(DEV_HOST_NS) {}
+	$(Q)oc get pods -n $(DEV_MEMBER_NS) -l name=member-operator -o name | xargs -i oc delete -n $(DEV_MEMBER_NS) {}
--- a/make/test.mk
+++ b/make/test.mk
@@ -23,6 +23,7 @@ IMAGE_NAMES_DIR := /tmp/crt-e2e-image-names
 WAS_ALREADY_PAIRED_FILE := /tmp/${GO_PACKAGE_ORG_NAME}_${GO_PACKAGE_REPO_NAME}_already_paired
 
 .PHONY: test-e2e
+## Run the e2e tests
 test-e2e: deploy-e2e e2e-run
 	@echo "The tests successfully finished"
 	@echo "To clean the cluster run 'make clean-e2e-resources'"
--- a/make/test.mk
+++ b/make/test.mk
@@ -6,8 +6,8 @@
 
 QUAY_NAMESPACE ?= codeready-toolchain
 DATE_SUFFIX := $(shell date +'%d%H%M%S')
-MEMBER_NS ?= ${QUAY_NAMESPACE}-member-${DATE_SUFFIX}
-HOST_NS ?= ${QUAY_NAMESPACE}-host-${DATE_SUFFIX}
+MEMBER_NS := ${QUAY_NAMESPACE}-member-${DATE_SUFFIX}
+HOST_NS := ${QUAY_NAMESPACE}-host-${DATE_SUFFIX}
 REGISTRATION_SERVICE_NS := $(HOST_NS)
 TEST_NS := ${QUAY_NAMESPACE}-toolchain-e2e-${DATE_SUFFIX}
 AUTHOR_LINK := $(shell jq -r '.refs[0].pulls[0].author_link' <<< $${CLONEREFS_OPTIONS} | tr -d '[:space:]')
--- a/make/test.mk
+++ b/make/test.mk
@@ -100,11 +100,6 @@ setup-kubefed:
 	curl -sSL https://raw.githubusercontent.com/codeready-toolchain/toolchain-common/master/scripts/add-cluster.sh | bash -s -- -t member -mn $(MEMBER_NS) -hn $(HOST_NS) -s
 	curl -sSL https://raw.githubusercontent.com/codeready-toolchain/toolchain-common/master/scripts/add-cluster.sh | bash -s -- -t host -mn $(MEMBER_NS) -hn $(HOST_NS) -s
 
-.PHONY: clean-e2e-resources
-clean-e2e-resources:
-	$(Q)-oc get projects --output=name | grep -E "${QUAY_NAMESPACE}-(toolchain\-)?(member|host)(\-operator)?(\-[0-9]+)?|${QUAY_NAMESPACE}-toolchain\-e2e\-[0-9]+" | xargs oc delete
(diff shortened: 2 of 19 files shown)


make/clean.mk
```diff
@@ -1,4 +1,12 @@
 .PHONY: clean
 clean:
 	$(Q)-rm -rf ${V_FLAG} $(OUT_DIR) ./vendor
-	$(Q)go clean ${X_FLAG} ./...
\ No newline at end of file
+	$(Q)go clean ${X_FLAG} ./...
+
+.PHONY: clean-resources
+clean-resources:
```
alexeykazakov, line 7: It's a bit weird target... It deletes usersignups, all pods (including the operator pods, but not deployments) and cluster resource quotas only. What's the purpose of such target?
If you want to introduce a target to delete all resources created by usersignups (usersignup and all child resources) then shouldn't deleting the usersignups only be enough? Cluster resource qouatas are "child" objects too and should be deleted by GC. And why pods?

test/e2e/e2e_test.go
```diff
@@ -257,35 +256,35 @@ func createMultipleSignups(t *testing.T, ctx *framework.TestCtx, awaitility *wai
 		userSignup := newUserSignup(t, awaitility.Host(), fmt.Sprintf("multiple-signup-testuser-%d", i), fmt.Sprintf("multiple-signup-testuser-%d@test.com", i))
 		userSignup.Spec.Approved = true
 		err := awaitility.Host().Client.Create(context.TODO(), userSignup, testsupport.CleanupOptions(ctx))
-		awaitility.T.Logf("created UserSignup with username: '%s' and resource name: '%s'", userSignup.Spec.Username, userSignup.Name)
+		awaitility.T.Logf("created usersignupwith username: '%s' and resource name: '%s'", userSignup.Spec.Username, userSignup.Name)
```
alexeykazakov, line 259:
```suggestion
		awaitility.T.Logf("created usersignup with username: '%s' and resource name: '%s'", userSignup.Spec.Username, userSignup.Name)
```

test/e2e/e2e_test.go
```diff
@@ -257,35 +256,35 @@ func createMultipleSignups(t *testing.T, ctx *framework.TestCtx, awaitility *wai
...
 	memberAwait := wait.NewMemberAwaitility(awaitility)
-
+	revisions := tiers.GetRevisions(hostAwait, tier)
 	// Get the latest signup version
 	userSignup, err := awaitility.Host().WaitForUserSignup(signup.Name)
 	require.NoError(t, err)
 
 	// First, wait for the MasterUserRecord to exist, no matter what status
-	mur, err := hostAwait.WaitForMasterUserRecord(userSignup.Status.CompliantUsername)
+	mur, err := hostAwait.WaitForMasterUserRecord(userSignup.Status.CompliantUsername, wait.UntilMasterUserRecordHasConditions(provisioned()))
```
alexeykazakov, line 281: The comment above explicitly says that we don't care what status of the MUR at this point. The status is checked in line 317: https://github.com/codeready-toolchain/toolchain-e2e/pull/95/files#diff-6413e1dbafd6dd63b5d907ea3ae6f6f9L318

tiers/checks.go
```diff
@@ -73,26 +75,36 @@ func getDefaultChecks(nsType string) []innerObjectCheck {
...
 
 func (a *basicTierChecks) GetExpectedRevisions(awaitility *wait.Awaitility) Revisions {
-	revisions := GetRevisions(awaitility, "basic", "code", "dev", "stage")
-	return revisions
+	return GetRevisions(awaitility.Host(), "basic")
+}
+
+func (a *basicTierChecks) GetClusterObjectChecks() []clusterObjectsCheck {
+	return []clusterObjectsCheck{
+		numberOfClusterResourceQuotas(1),
```
alexeykazakov, line 83: Just checking the number of quotas is not enough. Let's also check the content. Similar to `limitRange()` check: https://github.com/codeready-toolchain/toolchain-e2e/blob/master/tiers/checks.go#L154

Also, different tiers will have different quotas. So the check should be parametrized.
See https://github.com/codeready-toolchain/host-operator/pull/177 and https://github.com/codeready-toolchain/toolchain-e2e/pull/97

tiers/nstemplateset.go
```diff
@@ -33,21 +33,18 @@ Revisions:
...
+		for _, check := range tierChecks.GetNamespaceObjectChecks(key) {
 			check(t, ns, memberAwait, userAccount.Name)
 		}
 	}
+	if expectedRevisions.ClusterResources != nil {
+		for _, check := range tierChecks.GetClusterObjectChecks() {
+			check(t, memberAwait, userAccount.Name)
+		}
+	}
+
```
alexeykazakov, line 49: Let's also add a check for deleted cluster quotas to our GC test: https://github.com/codeready-toolchain/toolchain-e2e/blob/master/test/e2e/e2e_test.go#L136

tiers/revisions.go
```diff
@@ -8,34 +8,31 @@ import (
...
-	require.Len(awaitility.T, templateTier.Spec.Namespaces, len(nsTypes))
-	revisions := Revisions{
-		Namespaces: make(map[string]string, len(nsTypes)),
-	}
-	for _, typ := range nsTypes {
-		r, found := namespaceRevision(*templateTier, typ)
-		require.True(awaitility.T, found, "unable to find revision for '%s' namespace in the '%s' NSTemplateTier", typ, tier)
-		revisions.Namespaces[typ] = r
+// GetRevisions returns the expected revisions for all the namespace templates and the optional cluster resources template for the given tier
+func GetRevisions(hostAwait *wait.HostAwaitility, tier string) Revisions {
```
alexeykazakov, line 20: Nice improvement.

xcoulon: thanks ;)

make/clean.mk
xcoulon: this is a target I added while doing trying things on a temp cluster. Deleting the pods meant I would start with fresh pods and hence, so it's a trick to clear the logs, actually ;)
but I can remove this target as it was really tailored for my own testing.

tiers/checks.go
xcoulon: ok, makes sense

tiers/nstemplateset.go
xcoulon: ok

test/e2e/e2e_test.go
xcoulon: ok, I'll revert this change.

make/clean.mk
alexeykazakov: But why not using `make clean-e2e-resources` for that?

MatousJobanek: > But why not using make clean-e2e-resources for that?

because `clean-e2e-resources` deletes the whole deployment of the operator including the `CatalogSource`. This target is just to delete the internal toolchain objects (via deleting `usersignups` and deleting `clusterresourcequotas` specifically) plus retriggering the deployment of the operators (via deleting the pod). In addition, this can be used only with the dev namespaces (namespace name without the time suffix) otherwise, it wouldn't work. 

I understand the purpose of the target, but I would do some additional changes there:
* rename for example to `clean-toolchain-resources-in-dev`
* change `HOST_NS` to `DEV_HOST_NS` and `MEMBER_NS` to `DEV_MEMBER_NS`
* when the usersignups are deleted then I would delete the rest of the toolchain objects (including BannedUser, NSTemplateTier, etc.)
* for `xargs oc delete` skip the empty values so it doesn't fail when there is no resource found.

I'll send you a suggestion in the next comment ;-)

make/clean.mk
```diff
@@ -1,4 +1,12 @@
...
+	$(Q)echo "cleaning resources in $(HOST_NS) and $(MEMBER_NS)..."
+	$(Q)oc get usersignups -n $(HOST_NS) -o name | xargs oc delete -n $(HOST_NS)
+	$(Q)oc get pods -n $(HOST_NS) -l name=host-operator -o name | xargs oc delete -n $(HOST_NS)
+	$(Q)oc get pods -n $(MEMBER_NS) -l name=member-operator -o name | xargs oc delete -n $(MEMBER_NS)
+	$(Q)oc get clusterresourcequotas -l "toolchain.dev.openshift.com/provider"=codeready-toolchain -o name| xargs oc delete
```
MatousJobanek, line 12: I wrote it quickly - there is maybe a way to make it nicer :-) 
```suggestion
.PHONY: clean-toolchain-resources-in-dev
clean-toolchain-resources-in-dev:
	$(Q)echo "cleaning resources in $(DEV_HOST_NS) and $(DEV_MEMBER_NS)..."
	$(Q)oc get usersignups -n $(DEV_HOST_NS) -o name | xargs -i oc delete -n $(DEV_HOST_NS) {}
	$(Q)for CRD in `oc get crd -o name -n $(DEV_HOST_NS) | grep toolchain`; do \
		echo "deleting $${CRD} in $(DEV_HOST_NS)"; \
		CRD_NAME=`oc get $${CRD} -n $(DEV_HOST_NS) --template '{{.metadata.name}}'`; \
		oc get $${CRD_NAME} -n $(DEV_HOST_NS) -o name | xargs -i oc delete {}; \
	done
	$(Q)for CRD in `oc get crd -o name -n $(DEV_MEMBER_NS) | grep toolchain`; do \
		echo "deleting $${CRD} in $(DEV_MEMBER_NS)"; \
		CRD_NAME=`oc get $${CRD} -n $(DEV_MEMBER_NS) --template '{{.metadata.name}}'`; \
		oc get $${CRD_NAME} -n $(DEV_MEMBER_NS) -o name | xargs -i oc delete {}; \
	done
	$(Q)oc get clusterresourcequotas -l "toolchain.dev.openshift.com/provider"=codeready-toolchain -o name -n $(DEV_MEMBER_NS) | xargs -i oc delete {}
	$(Q)oc get pods -n $(DEV_HOST_NS) -l name=host-operator -o name | xargs -i oc delete -n $(DEV_HOST_NS) {}
	$(Q)oc get pods -n $(DEV_MEMBER_NS) -l name=member-operator -o name | xargs -i oc delete -n $(DEV_MEMBER_NS) {}
```

tiers/checks.go
```diff
@@ -73,28 +81,38 @@ func getDefaultChecks(nsType string) []innerObjectCheck {
 }
 
 func (a *basicTierChecks) GetExpectedRevisions(awaitility *wait.Awaitility) Revisions {
-	revisions := GetRevisions(awaitility, "basic", "code", "dev", "stage")
-	return revisions
+	return GetRevisions(awaitility.Host(), "basic")
```
MatousJobanek, line 84: Since you refactored the `GetRevisions` function, then we don't have any place where the expected set of namespace types would be verified. In other words, with this change we don't verify that for `basic`/`advanced` tiers there should be `dev`, `stage` and `code` ns types and for `team` tier only `dev` and `stage` types (not less not more)

Could you please add such a check somewhere? For example here in the `GetExpectedRevisions` method or into the `nstemplatetier_test.go` file

tiers/checks.go
```diff
@@ -262,11 +287,24 @@ func numberOfLimitRanges(number int) innerObjectCheck {
...
 		assert.Len(t, nps.Items, number)
 	}
 }
+
+func numberOfClusterResourceQuotas(number int) clusterObjectsCheck {
+	return func(t *testing.T, memberAwait *wait.MemberAwaitility, userName string) {
+		quotas := &quotav1.ClusterResourceQuotaList{}
+		matchingLabels := client.MatchingLabels(map[string]string{ // make sure we only list the ClusterResourceQuota resources associated with the given "userName"
+			"toolchain.dev.openshift.com/provider": "codeready-toolchain",
+			"toolchain.dev.openshift.com/owner":    userName,
```
MatousJobanek, line 304: could we also verify the other labels - `toolchain.dev.openshift.com/type`,`toolchain.dev.openshift.com/tier` and `toolchain.dev.openshift.com/revision`?

make/clean.mk
alexeykazakov: This is a comment to the @MatousJobanek suggestion :)

- `clean-toolchain-resources-in-dev` is not an intuitive name. Let's add a detailed comment explaining what this target does. It might be worth adding it to readme too...
- You are deleting toolchain CRDs from the namespaces but CRDs are cluster scoped resources. So, that namespace iteration and `-n` params are redundant.
- Wouldn't deleting CRDs break the operators?
- What about deleting tiers?

I would move this target out of this PR to a separate one. Don't want to block this PR because of that.

tiers/checks.go
alexeykazakov: Good point @MatousJobanek. I would probably add such checks to `nstemplatetier_test.go`

make/clean.mk
MatousJobanek: crated PR for that: https://github.com/codeready-toolchain/toolchain-e2e/pull/100

make/test.mk
```diff
@@ -6,8 +6,8 @@
 
 QUAY_NAMESPACE ?= codeready-toolchain
 DATE_SUFFIX := $(shell date +'%d%H%M%S')
-MEMBER_NS := ${QUAY_NAMESPACE}-member-${DATE_SUFFIX}
-HOST_NS := ${QUAY_NAMESPACE}-host-${DATE_SUFFIX}
+MEMBER_NS ?= ${QUAY_NAMESPACE}-member-${DATE_SUFFIX}
+HOST_NS ?= ${QUAY_NAMESPACE}-host-${DATE_SUFFIX}
```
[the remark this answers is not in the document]
MatousJobanek: removed in 286596b

test/e2e/nstemplatetier_test.go
```diff
@@ -38,10 +37,6 @@ func TestNSTemplateTiers(t *testing.T) {
 	// all tiers to check - keep the basic as the last one, it will verify downgrade back to the default tier at the end of the test
 	tiersToCheck := []string{"advanced", "team", "basic"}
 
-	// when
-	// created on startup
-
-	// then
```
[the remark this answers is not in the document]
MatousJobanek: rewritten in 2771f05

test/e2e/nstemplatetier_test.go
```diff
@@ -61,12 +56,13 @@ func TestNSTemplateTiers(t *testing.T) {
 			UntilNSTemplateTierSpec(Not(HasClusterResources("000000a"))),
 		)
 		require.NoError(t, err)
-		tierChecks, err := tiers.NewChecks(tierToCheck)
+		// check resources before starting the tier promotions
+		mur, err := hostAwaitility.WaitForMasterUserRecord(testingTiersName, UntilMasterUserRecordHasConditions(provisioned()))
 		require.NoError(t, err)
+		verifyResourcesProvisionedForSignup(t, awaitility, johnSignup, mur.Spec.UserAccounts[0].Spec.NSTemplateSet.TierName)
```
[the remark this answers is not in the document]
MatousJobanek: changed in 28349f8

tiers/checks.go
MatousJobanek: added assertion to checks in 53bab50

test/e2e/e2e_test.go
MatousJobanek: done in 3e90269

test/e2e/e2e_test.go
```diff
@@ -163,21 +161,21 @@ func TestE2EFlow(t *testing.T) {
...
-		assert.NoError(t, err, "johnsmith-dev namnespace is not deleted")
+		assert.NoError(t, err, "johnsmith-dev namespace is not deleted")
 
 		err = memberAwait.WaitUntilNamespaceDeleted(johnsmithName, "stage")
-		assert.NoError(t, err, "johnsmith-stage namnespace is not deleted")
+		assert.NoError(t, err, "johnsmith-stage namespace is not deleted")
 
 		// also, verify that other user's resource are left intact
-		verifyResourcesProvisionedForSignup(t, awaitility, johnExtraSignup, revisions, "basic")
+		verifyResourcesProvisionedForSignup(t, awaitility, johnExtraSignup, "basic")
```
alexeykazakov, line 173: We also need to add a check that the ClusterQuota is deleted too.

tiers/nstemplateset.go
MatousJobanek: verified in 7b2eb76

test/e2e/e2e_test.go
MatousJobanek: it's already done - see 7b2eb76

alexeykazakov: It wasn't when I was adding that comment :) Thanks for addressing it!

PR #102, opened 2020-04-17
Extracting additional user information

tinakurian:
In order to customize notification messages to users, we require additional user information which we should store in the UserSignup CRD when the user first registers. The new fields being added for the notification service messages are givenName, familyName and company. This PR will test that the fields were set correctly in the usersignup CRD via invoking the http post in registration service.

related to: https://github.com/codeready-toolchain/registration-service/pull/102
jira: https://issues.redhat.com/browse/CRT-518

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -207,8 +207,11 @@ func createAndApproveSignup(t *testing.T, awaitility *wait.Awaitility, username 
 func postSignup(t *testing.T, route string, identity authsupport.Identity) {
 	// Call signup endpoint with a valid token.
 	emailClaim := authsupport.WithEmailClaim(uuid.NewV4().String() + "@email.tld")
+	companyClaim := authsupport.WithCompanyClaim("red hat")
+	givenNameClaim := authsupport.WithGivenNameClaim("jane")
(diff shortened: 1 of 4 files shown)


test/e2e/e2e_test.go
```diff
@@ -207,8 +210,11 @@ func createAndApproveSignup(t *testing.T, awaitility *wait.Awaitility, username
 func postSignup(t *testing.T, route string, identity authsupport.Identity) {
 	// Call signup endpoint with a valid token.
 	emailClaim := authsupport.WithEmailClaim(uuid.NewV4().String() + "@email.tld")
+	givenNameClaim := authsupport.WithGivenNameClaim("jane")
+	familyNameClaim := authsupport.WithFamilyNameClaim("doe")
+	companyClaim := authsupport.WithCompanyClaim("red hat")
```
alexeykazakov, line 215: Instead of hardcoding the same name for all the usersignups in the tests we could generate unique ones based on username or userID. Something like:
```
givenNameClaim := authsupport.WithGivenNameClaim(identity.Username + "-First-Name")
```

tinakurian: fixed here: bb2c80307a4f90e88a7cc77e78ee01a1a7dbbe78

PR #104, opened 2020-04-28
feat: create TierTemplate resources at startup

xcoulon:
verify that the new `TemplateRef` field is set
in all tiers (on namespaces and cluster resources
when applicable) and that the corresponding
TierTemplate resources exist when checking the
NSTemplateTiers

Fixes https://issues.redhat.com/browse/CRT-500


xcoulon:
> Elegant solution 👍 :-)
> 
> BTW, do you think that it would be good to also verify that there are only those TierTemplates created that should be - no more no less?

done in 00e1b70

MatousJobanek:
@xcoulon just please run `go mod tidy` before merging the PR - there are a lot of redundant dependencies

PR #108, opened 2020-05-05
feat: if the operator CSV already exists then update it

MatousJobanek:
* don't create new OLM resources for every `dev-deploy-e2e*` run - use the same name suffix
* if the operator namespace already contains the CSV then update it to force the redeployment

### DIFF
--- a/make/dev.mk
+++ b/make/dev.mk
@@ -9,14 +9,14 @@ dev-deploy-e2e: deploy-e2e-to-dev-namespaces print-reg-service-link
 
 .PHONY: deploy-e2e-to-dev-namespaces
 deploy-e2e-to-dev-namespaces:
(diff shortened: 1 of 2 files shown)


make/dev.mk
```diff
@@ -9,14 +9,14 @@ dev-deploy-e2e: deploy-e2e-to-dev-namespaces print-reg-service-link
 
 .PHONY: deploy-e2e-to-dev-namespaces
 deploy-e2e-to-dev-namespaces:
-	$(MAKE) deploy-e2e MEMBER_NS=${DEV_MEMBER_NS} HOST_NS=${DEV_HOST_NS} REGISTRATION_SERVICE_NS=${DEV_REGISTRATION_SERVICE_NS} ENVIRONMENT=${DEV_ENVIRONMENT}
+	$(MAKE) deploy-e2e MEMBER_NS=${DEV_MEMBER_NS} HOST_NS=${DEV_HOST_NS} REGISTRATION_SERVICE_NS=${DEV_REGISTRATION_SERVICE_NS} ENVIRONMENT=${DEV_ENVIRONMENT} RESOURCES_SUFFIX=dev
```
rajivnathan, line 12: We can replace `RESOURCES_SUFFIX=dev` with `RESOURCES_SUFFIX=$DEV_ENVIRONMENT`?

MatousJobanek: good idea. done in cad21ad

PR #110, opened 2020-05-14
chore(CRT-501): verify migration to templateRef in member-operator

MatousJobanek:
verifies migration to templateRef field in member-operator. It also verifies that already existing namespaces should be updated to the latest label set.

It drops `nstemplateset_test.go` since it was just a duplication and didn't cover a real scenario.

paired PR: https://github.com/codeready-toolchain/member-operator/pull/156


MatousJobanek:
> The test looks good. Not sure why it's failing though :)

because of this: 68346ae 
the namespaces shouln't be cleaned up by the framework but via a finalizer of the NSTemplateSet

PR #113, opened 2020-05-19
chore: Update prereq doc and checks

rajivnathan:
I recently went through the setup process again and thought these would be small but useful changes for newcomers to the project.


make/find-tools.mk
```diff
@@ -13,6 +13,22 @@ ifeq ($(VERBOSE),1)
...
+eg.
+brew install gnu-sed
+export PATH="/usr/local/opt/gnu-sed/libexec/gnubin:$$PATH
+
+endef
+
+# On MacOS, the default sed install will cause a malformed target YAML, use gnu-sed instead
+ifeq ($(shell uname -s),Darwin)
+$(if $(shell sed --version 2&>1 > /dev/null || echo "bad sed"),$(error ${SED_MSG}))
+endif
```
MatousJobanek, line 34: yes, or fix our scripts to make them POSIX compatible (I'm too lazy and I don't have MacOS :-))

rajivnathan: Yeah, I was planning to take a look at that at some point. I can do that in a follow-up PR if I get around to it.

PR #124, opened 2020-06-15
test: verify that TemplateUpdateRequest are created

xcoulon:
scenario:
- create users
- promote to new tier (to avoid messing with default ones)
- update the template refs in the tier
- verify that all TemplateUpdateRequests have
 been created (but not more)

(Note: let's use some funny names in the tests for once)


go.mod
```diff
@@ -2,14 +2,15 @@ module github.com/codeready-toolchain/toolchain-e2e
 
 require (
 	github.com/codeready-toolchain/api v0.0.0-20200608101253-f0555f65f63d
+	github.com/codeready-toolchain/host-operator v0.0.0-00010101000000-000000000000
```
alexeykazakov, line 5: We managed to avoid adding dependencies to the operator source code in our e2e tests so far. It seems that you only need it for `MaxPoolSize`. Let's hardcode the expected value in the tests instead. It would be better to hard code it in the tests anyway. Just to make sure we have the expected configuration (hardcoded atm) in the operator.

go.mod
```diff
@@ -25,4 +26,8 @@ replace (
 	k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20200204173128-addea2498afe // avoids case-insensitive import collision: "github.com/googleapis/gnostic/openapiv2" and "github.com/googleapis/gnostic/OpenAPIv2"
 )
 
+replace github.com/codeready-toolchain/host-operator => github.com/xcoulon/host-operator v0.0.0-20200611083634-f34811a7e366
```
alexeykazakov, line 29: A reminder.

go.mod
xcoulon: sure, let me do that

go.mod
xcoulon: yes, thanks ;)

go.mod
xcoulon: done in 42bea0b

test/e2e/masteruserrecord_test.go
```diff
@@ -0,0 +1,149 @@
...
+	templatesFromAdvanced := createBasicTierTemplate(t, ctx, awaitility, tierTemplatesByTier["advanced"], "123abcd")
+	templatesFromTeam := createBasicTierTemplate(t, ctx, awaitility, tierTemplatesByTier["team"], "456efgh")
+	templatesFromBasic := createBasicTierTemplate(t, ctx, awaitility, tierTemplatesByTier["basic"], "789ijkl")
+
+	// wait for the user to be provisioned for the first time
+	verifyResourcesProvisionedForSignup(t, awaitility, userSignup, "basic")
+
+	t.Run("update to refs basic-*-123abcd made out of the advanced TierTemplates", func(t *testing.T) {
+		// when update to basic-*-123abcd
+		replaceMurRefs(t, awaitility, murName, templatesFromAdvanced)
```
alexeykazakov, line 41: Do we actually support changing template refs directly in MUR right now? Wouldn't TemplateTier reconcile sync the MURs to the refs used by the tier?

test/e2e/nstemplatetier_test.go
```diff
@@ -87,47 +90,76 @@ func TestNSTemplateTiers(t *testing.T) {
...
+	}, advanced)
+	require.NoError(t, err)
+	// when
+	cheesecake.Spec.ClusterResources = advanced.Spec.ClusterResources
+	cheesecake.Spec.Namespaces = advanced.Spec.Namespaces
+	err = hostAwaitility.Client.Update(context.TODO(), cheesecake)
+	// then
+	require.NoError(t, err)
+	// and there should be `MaxPoolSize` TemplateUpdateRequests created
+	err = hostAwaitility.WaitForTemplateUpdateRequests(hostAwaitility.Ns, MaxPoolSize)
```
alexeykazakov, line 161: Also let's check that the corresponding NamespaceTemplateSet was actually updated and now both Cluster and Namespace templates are using the updated templates.

Also, I would add some "noise". Users/Murs with tiers we are not updating. It will test all our label/selector dance.

test/e2e/masteruserrecord_test.go
xcoulon: I moved this test in a separate file, but it was authored by @MatousJobanek in https://github.com/codeready-toolchain/toolchain-e2e/pull/119/files#diff-26ea7544c30692a2091a1b3ec8907eb7R110. 

The context for his PR is: verifying that the MUR is provisioned when it is changed. And yes, the change will be done by the TemplateUpdateRequestController which is coming in a subsequent PR. In other words, we are testing portions of the workflow and there are still a few missing pieces in the puzzle ;)

test/e2e/nstemplatetier_test.go
xcoulon: > Also let's check that the corresponding NamespaceTemplateSet was actually updated and now both Cluster and Namespace templates are using the updated templates.

where, in the MasterUserRecords? That's not implemented yet. This PR is for https://issues.redhat.com/browse/CRT-683 (which deals with creating all the `TemplateUpdateRequest` resources), whereas the update of the `MasterUserRecord` resources will be covered in https://issues.redhat.com/browse/CRT-685

alexeykazakov: Ah, right.

test/e2e/masteruserrecord_test.go
alexeykazakov: Ok. I’m just concerned about creating a flaky test when we “manually” change the template refs in MUR which could be overridden by a Tier controller reconcile.

PR #125, opened 2020-06-18
addressing pr comments

tinakurian:
E2E tests for notification. 

Due to the notification being cleaned up now in the notfication controller, we can no longer verify the notification as part of `verifyResourcesProvisionedForSignup.` This is because we call `createMultipleSignups` and `createAndApproveSignup` at the start of the test. when the notifications are created `verifyResourcesProvisionedForSignup` will pass for the first notification. However, for the subsequent notifications this would fail because the notification controller would have already deleted the other notifications since the deletion time would have passed. This would cause us to be stuck attempting to retrieve notifications. 

In order to work around this issue, i've moved the notification tests into a new file called `notification_test.` If there is a more preferred way of testing notifications, please do let me know - i'm open to ideas!

Related PRs: 
https://github.com/codeready-toolchain/host-operator/pull/228
https://github.com/codeready-toolchain/api/pull/157

### DIFF
--- a/test/e2e/notification_test.go
+++ b/test/e2e/notification_test.go
@@ -43,36 +43,23 @@ func (s *notificationTestSuite) TestNotificationCleanup() {
 
 	// Create and approve "janedoe"
 	janedoeName := "janedoe"
-	janeSignup := createAndApproveSignup(s.T(), s.awaitility, janedoeName)
+	createAndApproveSignup(s.T(), s.awaitility, janedoeName)
 
 	s.T().Run("notification created and deleted", func(t *testing.T) {
-
 		hostAwait := wait.NewHostAwaitility(s.awaitility)
-		memberAwait := wait.NewMemberAwaitility(s.awaitility)
 
-		// Get the latest signup version
-		userSignup, err := s.awaitility.Host().WaitForUserSignup(janeSignup.Name)
-		require.NoError(t, err)
-
-		// First, wait for the MasterUserRecord to exist, no matter what status
-		mur, err := hostAwait.WaitForMasterUserRecord(userSignup.Status.CompliantUsername)
-		require.NoError(t, err)
-
-		// Then wait for the associated UserAccount to be provisioned
-		userAccount, err := memberAwait.WaitForUserAccount(mur.Name,
-			wait.UntilUserAccountHasConditions(provisioned()))
-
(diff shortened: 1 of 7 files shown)


test/e2e/notification_test.go
```diff
@@ -0,0 +1,78 @@
...
+		// First, wait for the MasterUserRecord to exist, no matter what status
+		mur, err := hostAwait.WaitForMasterUserRecord(userSignup.Status.CompliantUsername)
+		require.NoError(t, err)
+
+		// Then wait for the associated UserAccount to be provisioned
+		userAccount, err := memberAwait.WaitForUserAccount(mur.Name,
+			wait.UntilUserAccountHasConditions(provisioned()))
+
+		_, err = hostAwait.WaitForMasterUserRecord(mur.Name,
+			wait.UntilMasterUserRecordHasConditions(provisioned(), provisionedNotificationCRCreated()))
```
alexeykazakov, line 66: We don't really need to wait for userSignup (it's done when we do `createAndApproveSignup()` and UserAccounts. Waiting for the mur to be provisioned is enough. UserID, username, etc are also available in the MUR.
```suggestion
		// Then wait for the MUR to be provisioned
		mur, err = hostAwait.WaitForMasterUserRecord(mur.Name,
			wait.UntilMasterUserRecordHasConditions(provisioned(), provisionedNotificationCRCreated()))
```

test/e2e/notification_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+		notification, err := hostAwait.WaitForNotification(userAccount.Name + "-provisioned")
```
alexeykazakov, line 68: Let's wait for the notification with the `Sent` status.

test/e2e/notification_test.go
tinakurian: @alexeykazakov makes sense - is this more what you were looking for: e81dddd9dc778bdd808eecba8e30335074db22cb?

test/e2e/notification_test.go
tinakurian: fixed as part of: e81dddd9dc778bdd808eecba8e30335074db22cb

test/e2e/notification_test.go
alexeykazakov: yep

PR #129, opened 2020-07-03
fix(flaky test): test should use latest MUR when comparing with UA

MatousJobanek:
When we compare UserAccount with MUR we need to use the latest version of MUR, otherwise, the test of NSTempalteTier updates can fail: https://storage.googleapis.com/origin-ci-test/pr-logs/pull/codeready-toolchain_host-operator/239/pull-ci-codeready-toolchain-host-operator-master-e2e/1278939436381376512/build-log.txt
What happened?
1. User `cheesecakelover01` was provisioned and promoted to `cheesecake` tier containing basic templates
2. Test updated `cheesecake` Tier to use advanced template
3. There was synchronization between UserAccount and MUR of the `cheesecakelover01` user that changed the `SyncIndex`
4. Test loaded MUR with changed `SyncIndex` and in `Provisioned` state and saved the MUR (still with basic templates) in a var
5. Test started waiting for UserAccount having advanced templates but expecting that the NSTemplateSet will contain the same data as is in MUR - containing basic tempalte.
6. Test fails because the condition won't be ever met

### DIFF
--- a/wait/member.go
+++ b/wait/member.go
@@ -52,8 +52,8 @@ func UntilUserAccountHasSpec(expected toolchainv1alpha1.UserAccountSpec) UserAcc
 	}
 }
 
-// UntilUserAccountMatchesMur returns a `UserAccountWaitCriterion` which checks that the given
-// MasterUserRecordSpec and UserAccountSpecEmbedded are the expected specs
+// UntilUserAccountMatchesMur returns a `UserAccountWaitCriterion` which loads the existing MUR
+// and compares the first UserAccountSpecEmbedded in the MUR with the actual UserAccount spec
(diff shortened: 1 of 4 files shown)


wait/member.go
```diff
@@ -54,18 +54,24 @@ func UntilUserAccountHasSpec(expected toolchainv1alpha1.UserAccountSpec) UserAcc
 
 // UntilUserAccountMatchesMur returns a `UserAccountWaitCriterion` which checks that the given
 // MasterUserRecordSpec and UserAccountSpecEmbedded are the expected specs
```
alexeykazakov, line 56: The description needs to be updated too.

MatousJobanek: updated in 52a7d96

PR #135, opened 2020-07-25
Tests for Idler

alexeykazakov:
Tests for https://github.com/codeready-toolchain/host-operator/pull/245

These tests only checks that Idler resources are created during user provisioning. The Idler controller tests are more tricky to implement.
First at all there is no easy way of configuring a shorter timeout for idling (vs. 8 hours which are specified in all tiers).
Also, the tests might take awhile to execute because we would want to create many pods (with different controllers like Deployment, ReplicaSet, etc). Wait for them to start. And then wait for them to be killed.
Idler controller is pretty well covered by unit tests. So, not sure if it's that critical for us to introduce complex and slow e2e tests for full e2e coverage right now.

Any thoughts?

### DIFF
--- a/tiers/checks.go
+++ b/tiers/checks.go
@@ -79,6 +79,7 @@ func (a *basicTierChecks) GetClusterObjectChecks() []clusterObjectsCheck {
 	return []clusterObjectsCheck{
 		clusterResourceQuota("4000m", "1750m", "7Gi"),
 		numberOfClusterResourceQuotas(1),
+		idlers("code", "dev", "stage"),
 	}
 }
 
@@ -109,6 +110,7 @@ func (a *advancedTierChecks) GetClusterObjectChecks() []clusterObjectsCheck {
 	return []clusterObjectsCheck{
 		clusterResourceQuota("4000m", "1750m", "7Gi"),
 		numberOfClusterResourceQuotas(1),
+		idlers("code", "dev", "stage"),
 	}
 }
 
@@ -152,6 +154,7 @@ func (a *teamTierChecks) GetClusterObjectChecks() []clusterObjectsCheck {
 	return []clusterObjectsCheck{
 		clusterResourceQuota("4000m", "2000m", "15Gi"),
 		numberOfClusterResourceQuotas(1),
+		idlers("dev", "stage"),
 	}
 }
 
@@ -305,6 +308,17 @@ func networkPolicyIngress(name, group string) namespaceObjectsCheck {
 		}
 
 		assert.Equal(t, expected.Spec, np.Spec)
+	}
+}
+
+func idlers(namespaceTypes ...string) clusterObjectsCheck {
+	return func(t *testing.T, memberAwait *wait.MemberAwaitility, userName string) {
+		for _, nt := range namespaceTypes {
+			idler, err := memberAwait.WaitForIdler(fmt.Sprintf("%s-%s", userName, nt))
+			require.NoError(t, err)
+			assert.Equal(t, userName, idler.ObjectMeta.Labels["toolchain.dev.openshift.com/owner"])
+			assert.Equal(t, int32(28800), idler.Spec.TimeoutSeconds)
+		}
 	}
 }
 
--- a/tiers/checks.go
+++ b/tiers/checks.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/codeready-toolchain/api/pkg/apis/toolchain/v1alpha1"
+
 	"github.com/codeready-toolchain/toolchain-e2e/wait"
 	quotav1 "github.com/openshift/api/quota/v1"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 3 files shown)


rajivnathan:
That sounds reasonable to me.

Not sure if there'd be problems but another option I thought of that may be doable without increasing the e2e test time by much:
- cover just one or a two of the most used types? eg. Deployment and an individual pod
- the e2e test could modify the idler resource's timeoutSeconds to a few seconds

alexeykazakov:
> cover just one or a two of the most used types? eg. Deployment and an individual pod

Yes, we could cover just a few of them

> the e2e test could modify the idler resource's timeoutSeconds to a few seconds before creating the pod/deployment

This is actually a bigger problem. Right now we can modify the created Idlers (or any resource defined in tiers and created during user provisioning). If such resources are modified after creation then our NSTemplateSet controller won't re-store them. But it should. We just haven't implemented it yet but we should and will eventually. If any of such resources are modified by anything but the NSTemplateSet controller (Idler timeout is changed by some admin manually for example, or a namespace role is modified) then NSTemplateSet controller should roll-back such changes to make sure the state of those resources matches what is defined in the tier.
So, I would avoid creating tests which will be broken as soon as we introduce that functionality.

MatousJobanek:
> If such resources are modified after creation then our NSTemplateSet controller won't re-store them. But it should. We just haven't implemented it yet but we should and will eventually. If any of such resources are modified by anything but the NSTemplateSet controller (Idler timeout is changed by some admin manually for example, or a namespace role is modified) then NSTemplateSet controller should roll-back such changes to make sure the state of those resources matches what is defined in the tier.

This is already implemented - don't forget that this is cluster-scoped resource - we watch for changes of these resources. That's why you added: https://github.com/codeready-toolchain/member-operator/blob/master/pkg/controller/nstemplateset/cluster_resources.go#L92-L109

> First at all there is no easy way of configuring a shorter timeout for idling (vs. 8 hours which are specified in all tiers).

There is an easy way:
1. create a new tier with low values (just reuse an already existing tier, find the Idler objects and replace the `28800` values)
2. promote a user to this new tier
3. test the idler functionality

We already have a test that creates new tiers https://github.com/codeready-toolchain/toolchain-e2e/blob/master/test/e2e/nstemplatetier_test.go#L135-L140

tiers/checks.go
```diff
@@ -308,6 +311,17 @@ func networkPolicyIngress(name, group string) namespaceObjectsCheck {
...
 }
 
+func idlers(namespaceTypes ...string) clusterObjectsCheck {
+	return func(t *testing.T, memberAwait *wait.MemberAwaitility, userName string) {
+		for _, nt := range namespaceTypes {
+			idler, err := memberAwait.WaitForIdler(fmt.Sprintf("%s-%s", userName, nt))
+			require.NoError(t, err)
+			assert.Equal(t, userName, idler.ObjectMeta.Labels["toolchain.dev.openshift.com/owner"])
+			assert.Equal(t, int32(28800), idler.Spec.TimeoutSeconds)
+		}
```
MatousJobanek, line 321: Correct me if I'm wrong, but you don't verify the number of the `Idler` objects anywhere. 
We should check that to verify the case when a user is promoted to `team` tier, then the controller should remove the redundant `*-code` Idler object

alexeykazakov:
> This is already implemented - don't forget that this is cluster-scoped resource - we watch for changes of these resources. That's why you added: https://github.com/codeready-toolchain/member-operator/blob/master/pkg/controller/nstemplateset/cluster_resources.go#L92-L109

We do watch the changes but we still don't compare old and new objects properly. We currently use https://github.com/codeready-toolchain/member-operator/blob/master/pkg/controller/nstemplateset/cluster_resources.go#L329 for to check if the object needs to be updates which simply check the template ref label but not the actual object/spec state.
It's in our TODO list to make proper object comparison.
See, we still use `compareNotSupported`: https://github.com/codeready-toolchain/member-operator/blob/master/pkg/controller/nstemplateset/cluster_resources.go#L102
It's pretty easy to start using proper comparison now though.

> There is an easy way:
- create a new tier with low values (just reuse an already existing tier, find the Idler objects and replace the 28800 values)
- promote a user to this new tier
- test the idler functionality
We already have a test that creates new tiers https://github.com/codeready-toolchain/toolchain-e2e/blob/master/test/e2e/nstemplatetier_test.go#L135-L140

It's not that simple, unfortunately. This was the first thing I tried. We do already create new tiers and tests but we re-use the existing TierTemplates in the new tiers. The tricky part is to actually create a new TierTemplate. The interesting part is seating in OpenShift Template object which is basically a raw byte array. We would need to unmarshal it, modify, marshal back. Create a new template and refer it in the new tier. Or introduce a new tier yaml and load it as a brand new tier.

tiers/checks.go
alexeykazakov: Good point. Added it here: https://github.com/codeready-toolchain/toolchain-e2e/pull/135/commits/833004e32de39c1f40995cde2000f12f77094283

MatousJobanek:
> We do watch the changes but we still don't compare old and new objects properly. 

right, stuped me :man_facepalming: 
I forgot that "small" detail :man_facepalming:  :smile: 

>  The interesting part is seating in OpenShift Template object which is basically a raw byte array. We would need to unmarshal it, modify, marshal back.

Yeah, this is a tricky part. I thought that it could be similar to this:
https://github.com/codeready-toolchain/member-operator/blob/master/pkg/controller/nstemplateset/nstemplateset_controller_test.go#L1077-L1080
Then replace the object inside of the template and create a new TierTemplate.

But if you're saying that you tried that, it's too complicated and it's not worth it, then I believe you. 
We can also create a jira task as a stretch goal - just for a case when someone is bored and has nothing to do :laughing:

alexeykazakov:
> But if you're saying that you tried that, it's too complicated and it's not worth it, then I believe you.

It's not like it's not worth it or it's impossible to do. I actually think we should do it eventually :) I would want to cover more tier resources flows in our e2e tests. But I've been already working on this Idler thing for more than a month in my spare time. And won't spend another month to make the e2e tests perfect. So, my thoughts here is that the current tests are good enough for introducing the Idler now rather than postponing it even further.

> We can also create a jira task as a stretch goal - just for a case when someone is bored and has nothing to do

Yep. Let's do it.

alexeykazakov:
> We can also create a jira task as a stretch goal - just for a case when someone is bored and has nothing to do

https://issues.redhat.com/projects/CRT/issues/CRT-747

PR #141, opened 2020-08-11
refactor: move setup and assertions in 'testsupport'

xcoulon:
Moved the setup and assertions funcs in `testsupport` so we can later introduce a `test/perf` package and reuse the funcs.

### DIFF
--- a/testsupport/user_assertions.go
+++ b/testsupport/user_assertions.go
@@ -1,0 +1,149 @@
+package testsupport
+
+import (
+	"crypto/tls"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"testing"
+	"time"
+
+	"github.com/codeready-toolchain/api/pkg/apis/toolchain/v1alpha1"
+	toolchainv1alpha1 "github.com/codeready-toolchain/api/pkg/apis/toolchain/v1alpha1"
+	"github.com/codeready-toolchain/toolchain-common/pkg/cluster"
+	"github.com/codeready-toolchain/toolchain-e2e/tiers"
+	"github.com/codeready-toolchain/toolchain-e2e/wait"
+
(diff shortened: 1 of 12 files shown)


testsupport/user_assertions.go
```diff
@@ -0,0 +1,149 @@
...
+
+func VerifyMemberStatus(t *testing.T, memberAwait *wait.MemberAwaitility) {
+	_, err := memberAwait.WaitForMemberStatus(wait.UntilMemberStatusHasConditions(ToolchainStatusReady()))
+	require.NoError(t, err, "failed while waiting for MemberStatus")
+}
+
+func VerifyToolchainStatus(t *testing.T, hostAwait *wait.HostAwaitility) {
+	_, err := hostAwait.WaitForToolchainStatus(wait.UntilToolchainStatusHasConditions(ToolchainStatusReady()))
+	require.NoError(t, err, "failed while waiting for ToolchainStatus")
+}
```
MatousJobanek, line 113: You have put it into the the `user_assertion.go` file, but this is not related to user management verifications

xcoulon: ah right, let me move them in a `toolchain_status.go` file (or something similar)

xcoulon:
> Awesome idea 👍
> btw, you could use the "dot import" to make the lines of the code a bit shorter, but it's a detail :-)

good idea, let me check that

xcoulon:
> > Awesome idea 👍
> > btw, you could use the "dot import" to make the lines of the code a bit shorter, but it's a detail :-)
> 
> good idea, let me check that

done in 4d1424c

testsupport/user_assertions.go
xcoulon: done in 4d1424c: funcs moved into `testsupport/toolchain_status_assertions.go`

PR #145, opened 2020-08-20
WIP - fix: user account deleted but not recreated

xcoulon:
Note: may need some rework: the test verifies that the UserAccount was deleted and then waits for the UserAccount to exist again. However, there may be some race conditions (if the UserAccount is recreated too quickly, the test will fail while waiting for its deletion). Instead, we may want to check that the resource's UID changed.


xcoulon:
replaced by https://github.com/codeready-toolchain/toolchain-e2e/pull/146

PR #157, opened 2020-09-03
README to explain the performance tests

xcoulon:
a `README.adoc` file to explain how the performance tests work and what is measured. Living in the repository instead of in a Google Docs, as per @MatousJobanek's request ;)

### DIFF
--- a/test/perf/README.adoc
+++ b/test/perf/README.adoc
@@ -1,0 +1,66 @@
+= Performance Testing
+
+Before running the performance tests locally, ensure that you have one or more clusters. 
+See the https://github.com/codeready-toolchain/toolchain-e2e/blob/master/multicluster_setup.adoc[multi-cluster setup] doc for instructions on how to configure multiple clusters.
+
+== How performance is measured
+
+All metrics collected during the performance tests come from the `/metrics` endpoint provided by the metrics server running within the operator pod.
+Each controller automatically exposes its own metrics via the Operator SDK, so we don't need to add anything in our code, unless we want to expose extra data.
+
+While the metrics evolve with time, the `/metrics` endpoint only provides "instantaneous" values so we need to poll the data from this endpoint to track changes (eg: memory or CPU) or wait until a certain value or threshold is reached (eg: number of resources processed).
+
+A `Service` is associated with the `/metrics` endpoint on each operator pod, and the performance test creates a `Route` to expose the service. This is a requirement since the test runs outside of the cluster.
+
+Here's an example of the metrics for the Host operator:
+
+----
+...
+
+# HELP process_resident_memory_bytes Resident memory size in bytes.
+# TYPE process_resident_memory_bytes gauge
+process_resident_memory_bytes 6.1227008e+07
+# HELP process_virtual_memory_bytes Virtual memory size in bytes.
+# TYPE process_virtual_memory_bytes gauge
+process_virtual_memory_bytes 7.62011648e+08
+# HELP process_virtual_memory_max_bytes Maximum amount of virtual memory available in bytes.
+# TYPE process_virtual_memory_max_bytes gauge
+process_virtual_memory_max_bytes -1
+
+...
+
+# HELP controller_runtime_reconcile_total Total number of reconciliations per controller
+# TYPE controller_runtime_reconcile_total counter
+controller_runtime_reconcile_total{controller="masteruserrecord-controller",result="success"} 24
+controller_runtime_reconcile_total{controller="notification-controller",result="success"} 18
+controller_runtime_reconcile_total{controller="nstemplatetier-controller",result="success"} 3
+controller_runtime_reconcile_total{controller="registrationservice-controller",result="success"} 2
+controller_runtime_reconcile_total{controller="templateupdaterequest-controller",result="success"} 30
+controller_runtime_reconcile_total{controller="toolchaincluster-controller",result="success"} 16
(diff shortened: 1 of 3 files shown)


test/perf/perf_test.go
```diff
@@ -30,8 +30,11 @@ func TestPerformance(t *testing.T) {
 	// host metrics should become available at this point
 	metricsService, err := hostAwait.WaitForMetricsService("host-operator-metrics")
 	require.NoError(t, err, "failed while waiting for the 'host-operator-metrics' service")
+	// host metrics should become available again at this point
+	metricsRoute, err := hostAwait.SetupRouteForService(metricsService, "/metrics")
+	require.NoError(t, err, "failed while setting up or waiting for the route to the 'host-operator-metrics' service to be available")
 
-	count := 1000
+	count := 10
```
rajivnathan, line 37: Did you mean to reduce the count to 10? Maybe we need to somehow make this value configurable by environment? I often change this to 10 when developing and then change back to 1000 too.

test/perf/README.adoc
```diff
@@ -0,0 +1,66 @@
...
+== What is measured
+
+=== Processing existing resources after a new deployment
+
+To measure how much time a controller takes to process existing resources after a new deployment, the test does the following steps:
+
+1. Provision the user accounts.
+2. Create (if needed) a route to expose the `/metrics` endpoint to the outside world.
+3. Delete the operator pod. The Deployment controller will immediately trigger a new pod.
+4. Wait until the `/metrics` endpoint is available. At this point, the controller has fully bootstrapped, so the test can start the stopwatch.
```
rajivnathan, line 56: I guess this step is no longer accurate since you moved the `hostAwait.SetupRouteForService(metricsService, "/metrics")` before deleting the operator pod?

test/perf/README.adoc
```diff
@@ -0,0 +1,66 @@
...
+5. Wait until the `controller_runtime_reconcile_total(controller,usersignup-controller)` metric reaches the number of user accounts provisioned in step 1. Stop the stopwatch and log the results.
+
+
+Note:: With a limit of 5000 namespaces per member cluster and 3 namespaces per user provisioned with the `basic` tier, multiple member clusters may be needed to support the test case. With 1000 users, a
+single member cluster is enough.
```
rajivnathan, line 61: nit: is the newline here intentional?

rajivnathan (commented):
Nice to have this stuff documented 😃

test/perf/perf_test.go
xcoulon: rahhh, once again, this change was not supposed to be part of the PR! I wanted to run a quick test to capture some metrics, actually :(

test/perf/README.adoc
xcoulon: no, we want to wait until the endpoint is available *again*, ie, when the pod has finished bootstrapping after the deletion in step 3. 
Maybe this sentence should be reworded to clarify?

test/perf/README.adoc
xcoulon: it was not really intention, and it should not cause any issue with the rendering, but I will remove it nonetheless

xcoulon:
> Nice to have this stuff documented 😃

thanks for your review and feedback, @rajivnathan! 🙌

test/perf/perf_test.go
xcoulon: reverted in https://github.com/codeready-toolchain/toolchain-e2e/pull/157/commits/01e294cb1b77b636143209c8dd9e38145c5f8107

test/perf/README.adoc
rajivnathan: I think the sentence is correct, I just don't think the code is reflecting that anymore since you moved the `hostAwait.SetupRouteForService(metricsService, "/metrics")` up? 

Before this PR we captured the start time `start := time.Now()` after the metrics endpoint is available (after the operator pod was deleted and comes back up) but with the changes in this PR `start := time.Now()` would be called right after we delete the operator pod so the metrics endpoint may not be available when we capture the start time?

sbryzak (approved):
Looks good, well explained!

test/perf/README.adoc
xcoulon: @rajivnathan sorry for the late response, this PR totally slipped out of my mind. I reworded the steps in fabffcc. Please let me know if it's ok for you like that now.

rajivnathan: Yeah, that sounds accurate now. Thanks! One minor thing though is it incorrectly says the member operator pod when it's actually the host operator pod that is deleted.

test/perf/README.adoc
```diff
@@ -0,0 +1,66 @@
...
+
+== What is measured
+
+=== Processing existing resources after a new deployment
+
+To measure how much time a controller takes to process existing resources after a new deployment, the test does the following steps:
+
+1. Create (if needed) a route to expose the `/metrics` endpoint to the outside world.
+2. Provision the user accounts.
+3. Delete the member operator pod. The Deployment controller will immediately trigger a new one.
```
rajivnathan, line 55:
```suggestion
3. Delete the host operator pod. The Deployment controller will immediately trigger a new one.
```

rajivnathan (commented):
Thanks for updating it @xcoulon , I noticed after pulling in the latest changes from master there are a bunch of other changes showing under the Files changed tab, I was expecting to see only the perf README file.

rajivnathan (approved):
Nice, looks great! Thanks again!

PR #165, opened 2020-09-22
Add test for user deactivation notification

rajivnathan:
Adds tests for user deactivation notifications. The test checks that the deactivation notification resources is created when the user is deactivated and is deleted once the notification service processes the notification.

Related PR: https://github.com/codeready-toolchain/host-operator/pull/289


test/e2e/notification_test.go
```diff
@@ -55,4 +57,22 @@ func (s *notificationTestSuite) TestNotificationCleanup() {
 		err = s.hostAwait.WaitUntilNotificationDeleted(mur.Name + "-provisioned")
 		require.NoError(t, err)
 	})
+
+	s.T().Run("deactivation notification created and deleted", func(t *testing.T) {
```
alexeykazakov, line 61: I wonder if we actually would want to move this tests to our provisioning and deactivation flows.
Provisioning notification test IMO should be part of https://github.com/codeready-toolchain/toolchain-e2e/blob/master/testsupport/user_assertions.go#L27:6
And deactivation notification tests should be part of https://github.com/codeready-toolchain/toolchain-e2e/blob/master/testsupport/user_assertions.go#L27:6
Just to be clear. I'm not talking about adding a new test method to the existing suites. I'm talking about adding the notification status and resource checks in the existing corresponding tests. So it's tested as part of the existing flows instead of creating a new flow dedicated to the notification tests.

alexeykazakov (commented):
Looks good but do we have a test for user re-activation?

test/e2e/notification_test.go
rajivnathan: So I couldn't move the provisioning notification test into `VerifyResourcesProvisionedForSignup` because we do multiple creations:

eg.
```
	johnsmithName := "johnsmith"
	johnSignup := CreateAndApproveSignup(t, hostAwait, johnsmithName)
	extrajohnName := "extrajohn"
	johnExtraSignup := CreateAndApproveSignup(t, hostAwait, extrajohnName)

	VerifyResourcesProvisionedForSignup(t, hostAwait, memberAwait, johnSignup, "basic")
	VerifyResourcesProvisionedForSignup(t, hostAwait, memberAwait, johnExtraSignup, "basic")
```

So the first call to `VerifyResourcesProvisionedForSignup` will work but the second call will fail because this function will return an error
 `notification, err := hostAwait.WaitForNotification(mur.Name+"-provisioned", wait.UntilNotificationHasConditions(Sent()))` 

This is because the second creation is already processed (notification created and deleted) by the time it's called so I simply moved it into `CreateAndApproveSignup` instead so that we're always verifying it whenever we're using that function to create the usersignup so it is part of the existing flows like we wanted.

rajivnathan:
@alexeykazakov yep, we had an existing one here: https://github.com/codeready-toolchain/toolchain-e2e/blob/aa7fe7387570b4e2a42a7805b9d6b17ad8c96f68/test/e2e/user_management_test.go#L62

And it works as expected (verifies the deactivation notification status) because I updated the conditions in conditions.go to include the deactivation notification status.

alexeykazakov:
OK. Makes sense.

alexeykazakov (approved):

alexeykazakov:
> @alexeykazakov yep, we had an existing one here: https://github.com/codeready-toolchain/toolchain-e2e/blob/aa7fe7387570b4e2a42a7805b9d6b17ad8c96f68/test/e2e/user_management_test.go#L62
> 
> 
> 
> And it works as expected (verifies the deactivation notification status) because I updated the conditions in conditions.go to include the deactivation notification status.

I see... Everything looks good then :)

MatousJobanek (approved):
Nice :+1: :heart:

PR #167, opened 2020-09-25
Added e2e test for phone verification

sbryzak:
This PR adds a test for a successful phone verification workflow.

Fixes https://issues.redhat.com/browse/CRT-792


test/e2e/registration_service_test.go
```diff
@@ -429,6 +431,177 @@ func (s *registrationServiceTestSuite) TestSignupOK() {
...
+	require.True(s.T(), mpStatus["verificationRequired"].(bool))
+
+	// Approve usersignup.
+	userSignup.Spec.Approved = true
+	err = s.hostAwait.Client.Update(context.TODO(), userSignup)
+	require.NoError(s.T(), err)
+
+	// Ensure the Master User Record is NOT provisioned
+	_, err = s.hostAwait.WaitForMasterUserRecord(identity0.Username)
+	require.Error(s.T(), err)
```
alexeykazakov, line 497: We are waiting for the default timeout here. So, this test will be always too slow.
Let's maybe do the following instead:
- Wait for StatusVerificationRequired: https://github.com/codeready-toolchain/host-operator/blob/9dc20a3c6c4522f916dd7db045421e7d3bc638bf/pkg/controller/usersignup/usersignup_controller.go#L281
- Check (without wait) that no MUR created/exist

test/e2e/registration_service_test.go
```diff
@@ -429,6 +431,177 @@ func (s *registrationServiceTestSuite) TestSignupOK() {
...
+
+	// Initiate the verification process
```
alexeykazakov, line 499: Before that, let's also call the reg service /signup endpoint returns the expected status.

test/e2e/registration_service_test.go
```diff
@@ -429,6 +431,177 @@ func (s *registrationServiceTestSuite) TestSignupOK() {
...
+	require.NoError(s.T(), err)
+	req.Header.Set("Authorization", "Bearer "+token0)
+	req.Header.Set("content-type", "application/json")
+	resp, err = httpClient.Do(req)
+	require.NoError(s.T(), err)
+	defer close(s.T(), resp)
+
+	body, err = ioutil.ReadAll(resp.Body)
+	require.NoError(s.T(), err)
+	require.NotNil(s.T(), body)
```
alexeykazakov, line 512: What body do we expect? Is it some JSON or it should be empty? Let's be specific and verify the body content if we expect any.

test/e2e/registration_service_test.go
```diff
@@ -429,6 +431,177 @@ func (s *registrationServiceTestSuite) TestSignupOK() {
...
+	userSignup, err = s.hostAwait.WaitForUserSignup(identity0.ID.String())
+	require.NoError(s.T(), err)
+
+	// Check attempts has been incremented
+	require.NotEmpty(s.T(), userSignup.Annotations[v1alpha1.UserVerificationAttemptsAnnotationKey])
+
+	// Verify with the correct code
+	// Attempt to verify with an incorrect verification code
+	req, err = http.NewRequest("GET", s.route + fmt.Sprintf("/api/v1/signup/verification/%s",
+		userSignup.Annotations[v1alpha1.UserSignupVerificationCodeAnnotationKey]),
```
alexeykazakov, line 550: Let's also verify that the initial verification code from line 520 is still the same in UserSignup. Let's keep the code from line 520 and check that the same code in the update usersignup from line 541.
It would test that we do not update the code if verification failed.

alexeykazakov (commented):
This test looks good overall (I have a few suggestions though).
But what about other use cases?
- When we exceed the max number of attempts.
- When the phone number already in use (including banned users).
- Invalid phone number.
- ...

test/e2e/registration_service_test.go
```diff
@@ -429,6 +431,177 @@ func (s *registrationServiceTestSuite) TestSignupOK() {
...
+	assert.Equal(s.T(), identity0.Username, mp["username"])
+	require.IsType(s.T(), false, mpStatus["ready"])
+	assert.False(s.T(), mpStatus["ready"].(bool))
+	assert.Equal(s.T(), "PendingApproval", mpStatus["reason"])
+	require.True(s.T(), mpStatus["verificationRequired"].(bool))
+
+	// Approve usersignup.
+	userSignup.Spec.Approved = true
+	err = s.hostAwait.Client.Update(context.TODO(), userSignup)
+	require.NoError(s.T(), err)
```
MatousJobanek, line 493: you know what is weird to me - that we firstly approve and then wait for the verification. Shouldn't that be vice versa? Approve only when it's verified?

test/e2e/registration_service_test.go
MatousJobanek: For checking that it's not created you could use:
```
        err := s.hostAwait.WaitUntilMasterUserRecordDeleted(identity0.Username)
        require.NoError(t, err)
```

test/e2e/registration_service_test.go
```diff
@@ -429,6 +431,177 @@ func (s *registrationServiceTestSuite) TestSignupOK() {
 	assert.Equal(s.T(), ExpectedConsoleURL(s.T(), s.memberAwait, memberCluster), mp["consoleURL"])
 }
 
+func (s *registrationServiceTestSuite) TestPhoneVerification() {
```
MatousJobanek, line 434: The test has 170 lines... This is a pain to follow/understand - I got to the middle of the test and lost the whole context. Could you please break it down to either multiple subtests or multiple helper functions?

testsupport/user_setup.go
```diff
@@ -108,7 +108,7 @@ var HttpClient = &http.Client{
 func postSignup(t *testing.T, route string, identity authsupport.Identity) {
 	require.NotEmpty(t, route)
 	// Call signup endpoint with a valid token.
-	emailClaim := authsupport.WithEmailClaim(uuid.NewV4().String() + "@email.tld")
+	emailClaim := authsupport.WithEmailClaim(uuid.NewV4().String() + "@acme.com")
```
MatousJobanek, line 111: Just out of curiosity - what is the purpose of this change?

test/e2e/registration_service_test.go
alexeykazakov: Yeah I guess that would be less ambiguous...

test/e2e/registration_service_test.go
alexeykazakov: `WaitUntilMasterUserRecordDeleted()` is good for waiting for the MUR to be deleted:
T1. MUR exists.
T2. Do something which should delete the MUR.
T3. Wait for the MUR to not exist.

But it won't work properly for making sure that the MUR has never created:
T1. MUR does not exist.
T2. Do something which should prevent the MUR to be created
T3. Wait for the MUR to not exist (can pass even if the MUR is created with a delay).

testsupport/user_setup.go
sbryzak: We can configure certain email domains to not require phone verification - for testing purposes, we've configured both `redhat.com` and `acme.com` as excluded domains (see https://github.com/codeready-toolchain/host-operator/pull/291/files#diff-f9db715595ded24a97b8ca746f684769L7).  This change sets the email domain to one of the configured domains in order for the existing tests to still pass, as they would otherwise fail due to requiring verification.

test/e2e/registration_service_test.go
sbryzak: That workflow actually seems a little weird to me, if I'm a user and I'm confronted with a phone verification requirement which I go to the effort to complete by providing my personal phone number, I can imagine being more than a little annoyed if after that I'm *still* required to wait for someone to manually approve me.  If I know that I need to wait for approval in the first place, then the phone verification step won't be a blocker for me to start using the product once someone has approved me.

test/e2e/registration_service_test.go
sbryzak: It was a monumental effort to actually get the entire phone verification workflow tested and working successfully so yes, the size of the test did blow out a bit - I totally agree though that it's a bit too large to follow, so I'll make some changes to turn some of the duplicated code into reusable functions.

test/e2e/registration_service_test.go
alexeykazakov: Phone verification is basically a prove that the user is not a bot. In many systems you can't even login without going through some verification. Including 2FA.
So, asking for phone verification (so the user is not a bot) before we even consider the user to be granted access to the system is IMO justified. We can double check with Parag though.

test/e2e/registration_service_test.go
sbryzak: I've refactored it down to 100 lines, let me know if that looks ok now.

test/e2e/registration_service_test.go
sbryzak: Done.

test/e2e/registration_service_test.go
sbryzak: This has been refactored, now where a body is expected in the response the contents are checked.

test/e2e/registration_service_test.go
sbryzak: That would basically require all of this code yet again, when all we have done is manually approve the UserSignup.  We already have existing tests that check the status after a manual approval (see `TestSignupOK()`), do we really need to test this behaviour again here?

```	// Call get signup endpoint with a valid token and make sure it's pending approval
	body := invokeEndpoint(s.T(), "GET", s.route + "/api/v1/signup", token0, "", http.StatusOK)

	// Convert the signup response into a map so that we can examine its values
	mp := make(map[string]interface{})
	err = json.Unmarshal([]byte(body), &mp)
	require.NoError(s.T(), err)

	// Check that the response looks fine
	mpStatus, ok := mp["status"].(map[string]interface{})
	assert.True(s.T(), ok)
	assert.Equal(s.T(), "", mp["compliantUsername"])
	assert.Equal(s.T(), identity0.Username, mp["username"])
	require.IsType(s.T(), false, mpStatus["ready"])
	assert.False(s.T(), mpStatus["ready"].(bool))
	assert.Equal(s.T(), "PendingApproval", mpStatus["reason"])
	require.True(s.T(), mpStatus["verificationRequired"].(bool))

	// Now approve the usersignup.
	err = s.hostAwait.Client.Update(context.TODO(), userSignup)
	require.NoError(s.T(), err)```

File: test/e2e/registration_service_test.go (hunk 1 above)
Comment (sbryzak) [reply test/e2e/registration_service_test.go]: I've already updated this code to what Matous suggested - is it the consensus that this won't actually work?

File: test/e2e/registration_service_test.go (hunk 5 above)
Comment (sbryzak) [reply test/e2e/registration_service_test.go]: I would definitely like to double check on this, I still feel uncomfortable about forcing a user to provide their phone number with the possibility that they still might potentially need to wait for an extended period of time to be approved.  What about situations where we have run out of cluster resources - what is the expected process here and what will the delay be for the user?  What if there are thousands of users in the queue waiting to be approved?

Comment (sbryzak) [reply test/e2e/registration_service_test.go]: Also in response to the comment about many existing systems requiring verification, I am struggling to think of a single example in which there might still be a required manual approval step after already verifying.

File: test/e2e/registration_service_test.go (hunk 2 above)
Comment (alexeykazakov) [reply test/e2e/registration_service_test.go]: No, I think we are good. I see that we do check `require.True(s.T(), mpStatus["verificationRequired"].(bool))` in the snippet above. So we are good.

File: test/e2e/registration_service_test.go (hunk 1 above)
Comment (alexeykazakov) [reply test/e2e/registration_service_test.go]: `s.hostAwait.WaitUntilMasterUserRecordDeleted(identity0.Username) is unfortunately useless in this case. Take a look at my  previous message. So I think we have to implement it as a suggested in https://github.com/codeready-toolchain/toolchain-e2e/pull/167#discussion_r495219691.

File: test/e2e/registration_service_test.go (hunk 5 above)
Comment (MatousJobanek) [reply test/e2e/registration_service_test.go]: We have manual approval step now for CRT cluster. I guess that we won't have it as soon as we go live. But it can happen that the clusters are full, then the user would need to wait.

Anyway, if I compare two types of the user experience:

First one that requires inputs from the user in two phases:
1. Sign up for the Dev Sandbox (that may require also creating a new user account at redhat.com)
2. wait until my account is approved (can be minutes, hours or days)
3. add additional information for phone verification (with the fact that the phone can be rejected since it was already used for some other user account)
4. Finally, enter Dev Sandbox

or provide all information at the beginning and then just wait until it is approved:
1. Sign up for the Dev Sendbox (that may require also creating a new user account at redhat.com) & verify the phone
2. Wait until my account is approved (can be minutes, hours or days)
3. Enter Dev Sandbox

To be honest, the second one seems to be more logical and fluent. I as a user give all necessary information about myself (including phone verification) at the beginning, then I just wait until I'm approved.

File: test/e2e/registration_service_test.go (hunk 1 above)
Comment (MatousJobanek) [reply test/e2e/registration_service_test.go]: Yeah, I thought that checking that the MUR does not exist should be enough - for this particular use-case the method is sufficient - it verifies that he MUR does not exist. I know that ideally we should wait some short time and check that MUR is not really created, but since we are going to wait for StatusVerificationRequired (as suggested by Alexey), then we should be fine.

File: testsupport/user_setup.go (hunk 7 above)
Comment (MatousJobanek) [reply testsupport/user_setup.go:127]: Ok, I missed the connection. Thanks a lot for the explanation.

File: test/e2e/registration_service_test.go (hunk 1 above)
Comment (alexeykazakov) [reply test/e2e/registration_service_test.go]: There is a couple of cases where we don't have a choice and have to wait for some resource to be not be created. For example in https://github.com/codeready-toolchain/toolchain-e2e/blob/4996af7b68436dfe74338ef560c8e0cb60804060/test/e2e/usersignup_test.go#L133
But we introduced `WithRetryOptions()` to override the default timeout (which is too long) for such cases:
```
s.hostAwait.WithRetryOptions(wait.TimeoutOption(time.Second * 10)).WaitForMasterUserRecord(username)
require.Error(s.T(), err)
```
It still slows down the test but not for the entire default timeout (10 sec vs 60 sec per call). But in this particular case I think we could just wait for StatusVerificationRequired and don't wait for MUR at all. So we can save those 10 seconds.

File: test/e2e/registration_service_test.go (hunk 5 above)
Comment (alexeykazakov) [reply test/e2e/registration_service_test.go]: +1 @MatousJobanek

File: test/e2e/registration_service_test.go (hunk 1 above)
Comment (sbryzak) [reply test/e2e/registration_service_test.go]: I've removed the MUR check here and replaced it with a UserSignup status check, plus a check the confirm that `verificationRequired` is correctly set to true when calling the getSignup endpoint

File: test/e2e/registration_service_test.go (hunk 5 above)
Comment (sbryzak) [reply test/e2e/registration_service_test.go]: Actually looking at the code, this conversation is moot.... if verification is required then the `verificationRequired` property will be set to true regardless of whether the usersignup is approved or not, meaning that a new user should immediately have to perform phone verification after signing up anyway.  I've modified the test workflow to do the approval after the verification has been completed, in order to reflect this.

File: test/e2e/registration_service_test.go (hunk 1 above)
Comment (alexeykazakov) [reply test/e2e/registration_service_test.go]: But we still need to check that the MUR is not created. Why not checking that it does not exist?

Comment (sbryzak) [reply test/e2e/registration_service_test.go]: Your very last comment was "But in this particular case I think we could just wait for StatusVerificationRequired and don't wait for MUR at all." What did you mean by this?

Comment (alexeykazakov) [reply test/e2e/registration_service_test.go]: If we wait for StatusVerificationRequired then we don't have to **wait** for the MUR to. In other words we don't have to use any `WaitFor...()` function and can check if the MUR exists or not without any timeout/retries/waits. Sorry for not being clear.

Comment (alexeykazakov) [reply test/e2e/registration_service_test.go]: Just try to load the MUR from the client. It should not found any.

Comment (sbryzak) [reply test/e2e/registration_service_test.go]: No problem, your use of the word "just" in that context made me think you only cared about the status check.  I'll add a lookup then for the MUR and expect an error there.

File: test/e2e/registration_service_test.go [hunk 8]
```diff
@@ -429,6 +431,133 @@ func (s *registrationServiceTestSuite) TestSignupOK() {
...
+	require.IsType(s.T(), false, mpStatus["ready"])
+	assert.False(s.T(), mpStatus["ready"].(bool))
+	assert.Equal(s.T(), "PendingApproval", mpStatus["reason"])
+	require.True(s.T(), mpStatus["verificationRequired"].(bool))
+
+	// Confirm the status of the UserSignup is correct
+	_, err = s.hostAwait.WaitForUserSignup(identity0.ID.String(), wait.UntilUserSignupHasConditions(NotApprovedAndVerificationRequired()...))
+
+	// Initiate the verification process
+	invokeEndpoint(s.T(), "PUT", s.route + "/api/v1/signup/verification", token0,
```
tinakurian, line 464: /@sbryzak it looks like there isn't a test for verifying that the user can request one too many new verification codes? this should probably be added

tinakurian: @sbryzak it would be nice if we could better organize the tests as well - it's a bit hard to follow

test/e2e/registration_service_test.go
```diff
@@ -429,6 +431,133 @@ func (s *registrationServiceTestSuite) TestSignupOK() {
...
+	// Retrieve the updated UserSignup
+	userSignup, err = s.hostAwait.WaitForUserSignup(identity0.ID.String())
+	require.NoError(s.T(), err)
+
+	// Confirm there is a verification code annotation value, and store it in a variable
+	verificationCode := userSignup.Annotations[v1alpha1.UserSignupVerificationCodeAnnotationKey]
+	require.NotEmpty(s.T(), verificationCode)
+
+	// Confirm the expiry time has been set
+	require.NotEmpty(s.T(), userSignup.Annotations[v1alpha1.UserVerificationExpiryAnnotationKey])
```
tinakurian, line 476: @sbryzak shouldn't we also test that when the expiry happens we get the correct result?

test/e2e/registration_service_test.go
```diff
@@ -429,6 +431,133 @@ func (s *registrationServiceTestSuite) TestSignupOK() {
...
+	assert.False(s.T(), mpStatus["ready"].(bool))
+	assert.Equal(s.T(), "PendingApproval", mpStatus["reason"])
+	require.True(s.T(), mpStatus["verificationRequired"].(bool))
+
+	// Confirm the status of the UserSignup is correct
+	_, err = s.hostAwait.WaitForUserSignup(identity0.ID.String(), wait.UntilUserSignupHasConditions(NotApprovedAndVerificationRequired()...))
+
+	// Initiate the verification process
+	invokeEndpoint(s.T(), "PUT", s.route + "/api/v1/signup/verification", token0,
+		`{ "country_code":"+61", "phone_number":"408999999" }`, http.StatusNoContent)
```
tinakurian, line 465: @sbryzak after looking at the tests i guess we don't test any invalid phone number entries - maybe we should add that?

test/e2e/registration_service_test.go
sbryzak: Doesn't this test in registration service already do this?

https://github.com/codeready-toolchain/registration-service/blob/master/pkg/controller/signup_test.go#L365

test/e2e/registration_service_test.go
sbryzak: I thought we had this covered already in registration service, if not we should probably add a test for it there.

test/e2e/registration_service_test.go
sbryzak: Same as above, this test doesn't require user interaction beyond registration service so doesn't need to be an end to end test, if we don't test this in registration service already then we should probably add it there.

test/e2e/registration_service_test.go
MatousJobanek: Yes, you are right. :+1: 
I'm going to swap the verification & approval phases in my PR anyway: https://github.com/codeready-toolchain/host-operator/pull/288/files#diff-5b4b18a0fc2e0d708e4e6334cdb6bc69L285-L294 so I'm glad that we agreed on the flow

test/e2e/registration_service_test.go
```diff
@@ -429,6 +432,138 @@ func (s *registrationServiceTestSuite) TestSignupOK() {
...
+func (s *registrationServiceTestSuite) TestPhoneVerification() {
+	// Create a token and identity to sign up with
+	identity0 := authsupport.NewIdentity()
+	emailValue := uuid.NewV4().String() + "@some.domain"
+	emailClaim0 := authsupport.WithEmailClaim(emailValue)
+	token0, err := authsupport.GenerateSignedE2ETestToken(*identity0, emailClaim0)
+	require.NoError(s.T(), err)
+
+	// Call the signup endpoint
+	invokeEndpoint(s.T(), "POST", s.route + "/api/v1/signup", token0, "", http.StatusAccepted)
```
MatousJobanek, line 444: I have maybe a dumb question - couldn't we use the `postSignup` function for that - it looks like it's doing almost the same https://github.com/codeready-toolchain/toolchain-e2e/blob/master/testsupport/user_setup.go#L124-L148
We could technically reuse the `invokeEndpoint` function there.

test/e2e/registration_service_test.go
```diff
@@ -429,6 +432,138 @@ func (s *registrationServiceTestSuite) TestSignupOK() {
...
+	assert.Equal(s.T(), "PendingApproval", mpStatus["reason"])
+	require.True(s.T(), mpStatus["verificationRequired"].(bool))
+
+	// Confirm the status of the UserSignup is correct
+	_, err = s.hostAwait.WaitForUserSignup(identity0.ID.String(), wait.UntilUserSignupHasConditions(NotApprovedAndVerificationRequired()...))
+
+	// Confirm that a MUR hasn't been created
+	obj := &v1alpha1.MasterUserRecord{}
+	err = s.hostAwait.Client.Get(context.TODO(), types.NamespacedName{Namespace: s.hostAwait.Namespace, Name: identity0.Username}, obj)
+	require.Error(s.T(), err)
```
MatousJobanek, line 467: It's just a detail, but we should check that the error is `NotFound` - something like this:
```suggestion
	require.Error(s.T(), err)
	require.True(s.T(), errors.IsNotFound(err))
```

test/e2e/registration_service_test.go
```diff
@@ -429,6 +432,138 @@ func (s *registrationServiceTestSuite) TestSignupOK() {
...
+
+	// Initiate the verification process
+	invokeEndpoint(s.T(), "PUT", s.route + "/api/v1/signup/verification", token0,
+		`{ "country_code":"+61", "phone_number":"408999999" }`, http.StatusNoContent)
+
+	// Retrieve the updated UserSignup
+	userSignup, err = s.hostAwait.WaitForUserSignup(identity0.ID.String())
+	require.NoError(s.T(), err)
```
MatousJobanek, line 475: If I understand it correctly, we send `PUT` request to the registration service with the country code and phone number and the reg-service then updates the UserSignup. I know that it may look like a detail, but it could happen that we get the UserSignup in the test earlier than registration service updates it - I was just fixing such flaky tests https://github.com/codeready-toolchain/toolchain-e2e/pull/172 so I'd like to avoid it for the future. 
WDYT? Could you create a `UserSignupWaitCriterion` for that?

test/e2e/registration_service_test.go
```diff
@@ -429,6 +432,138 @@ func (s *registrationServiceTestSuite) TestSignupOK() {
...
+	require.NotEmpty(s.T(), verificationCode)
+
+	// Confirm the expiry time has been set
+	require.NotEmpty(s.T(), userSignup.Annotations[v1alpha1.UserVerificationExpiryAnnotationKey])
+
+	// Attempt to verify with an incorrect verification code
+	invokeEndpoint(s.T(), "GET", s.route+"/api/v1/signup/verification/invalid", token0, "", http.StatusForbidden)
+
+	// Retrieve the updated UserSignup
+	userSignup, err = s.hostAwait.WaitForUserSignup(identity0.ID.String())
```
MatousJobanek, line 488: The same case is also here.

test/e2e/registration_service_test.go
```diff
@@ -429,6 +432,138 @@ func (s *registrationServiceTestSuite) TestSignupOK() {
...
+	require.NoError(s.T(), err)
+
+	// Check attempts has been incremented
+	require.NotEmpty(s.T(), userSignup.Annotations[v1alpha1.UserVerificationAttemptsAnnotationKey])
```
MatousJobanek, line 492: btw, do you think that we could verify the expected number (and also in other cases of the annotations)? or at least verify that the number has increased since the last time...?

test/e2e/registration_service_test.go
```diff
@@ -429,6 +432,138 @@ func (s *registrationServiceTestSuite) TestSignupOK() {
...
+
+	// Confirm the verification code has not changed
+	require.Equal(s.T(), verificationCode, userSignup.Annotations[v1alpha1.UserSignupVerificationCodeAnnotationKey])
+
+	// Verify with the correct code
+	invokeEndpoint(s.T(), "GET", s.route + fmt.Sprintf("/api/v1/signup/verification/%s",
+		userSignup.Annotations[v1alpha1.UserSignupVerificationCodeAnnotationKey]), token0, "", http.StatusOK)
+
+	// Retrieve the updated UserSignup
+	userSignup, err = s.hostAwait.WaitForUserSignup(identity0.ID.String())
```
MatousJobanek, line 502: same case with the wait also here

test/e2e/registration_service_test.go
```diff
@@ -429,6 +432,138 @@ func (s *registrationServiceTestSuite) TestSignupOK() {
...
+	assert.Equal(s.T(), "PendingApproval", mpStatus["reason"])
+	require.False(s.T(), mpStatus["verificationRequired"].(bool))
+
+	// Now approve the usersignup.
+	userSignup.Spec.Approved = true
+	err = s.hostAwait.Client.Update(context.TODO(), userSignup)
+	require.NoError(s.T(), err)
+
+	// Confirm the MasterUserRecord is provisioned
+	_, err = s.hostAwait.WaitForMasterUserRecord(identity0.Username)
```
MatousJobanek, line 528: if we want to check that it's provisioned then we should verify the existence of the condition:
```suggestion
	_, err = s.hostAwait.WaitForMasterUserRecord(identity0.Username, wait.UntilMasterUserRecordHasCondition(Provisioned()))
```

MatousJobanek (commented):
Thanks a lot for changing the structure of the tests - this is much better.
I have a few questions/proposals, but nothing significant. Good job :+1:

test/e2e/registration_service_test.go
sbryzak: I could use it, but it's not going to reduce the lines of code - the email claim and the token are still required as they are used further down in the test.

test/e2e/registration_service_test.go
sbryzak: No problem, added.

test/e2e/registration_service_test.go
sbryzak: I'm not quite sure what you mean here.  The init verification (i.e. the PUT request) is synchronous, it will not return a response until after the UserSignup has been updated.  The `invokeEndpoint()` function is also synchronous, it will not return until it has fully read the body of the response.  How is it possible that we could get the `UserSignup` in the test earlier than the registration service updates it?

test/e2e/registration_service_test.go
sbryzak: We could add this test to registration service, I'll make a note to do it when I take care of the architectural refactor.  I discussed this very issue (of trying to cover every possible scenario in the e2e tests) with Alexey today and he agreed with me that the e2e tests was not the right place to do this.  If there's certain conditions that we don't currently test for in registration service then we should add a new test there, not in the e2e tests.

test/e2e/registration_service_test.go
sbryzak: Good idea, updated.

test/e2e/registration_service_test.go
sbryzak: For reference - I've added a reminder to add a test to cover this here: https://issues.redhat.com/browse/CRT-825

testsupport/conditions.go
```diff
@@ -112,6 +112,26 @@ func VerificationRequired() []toolchainv1alpha1.Condition {
 	}
 }
 
+
+func NotApprovedAndVerificationRequired() []toolchainv1alpha1.Condition {
+	return []toolchainv1alpha1.Condition{
+		{
+			Type:   toolchainv1alpha1.UserSignupApproved,
+			Status: corev1.ConditionFalse,
+		},
```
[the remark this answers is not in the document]
sbryzak: Good catch, I've added it.

test/e2e/registration_service_test.go
MatousJobanek: ah, right - I totally forgot that it is synchronous (I'm still living in the asynchronous world :grimacing:) Thanks for explanation :+1:

test/e2e/registration_service_test.go
MatousJobanek: cool, thanks :+1:

MatousJobanek (approved):
Thanks for addressing my comments :+1:

test/e2e/registration_service_test.go
tinakurian: yep it does, but the tests in registration-service are heavily mocked - it would be nice to have some that aren't mocked imo

PR #176, opened 2020-10-09
chore: verify creation of new SA when joining a cluster

MatousJobanek:
Verify that the creation of new SA with limited Role & ClusterRole when joining clusters works.
It also creates a new SA for the e2e tests and uses the `add-cluster.sh` script also for copying the token and creating separated `ToolchainCluster` resource.

Relies on changes from: https://github.com/codeready-toolchain/toolchain-common/pull/133


alexeykazakov:
Hm...
```
deployments.apps "member-operator" is forbidden: User "system:serviceaccount:codeready-toolchain-member-09145129:toolchaincluster-member-operator" cannot get resource "deployments" in API group "apps" in the namespace "codeready-toolchain-member-09145129"
```

MatousJobanek:
@alexeykazakov  yeah, I know. The original code was wrong. I rewrote the script again and it should be working fine now.

alexeykazakov (approved):
Overall looks good but after a quick chat with Matous we agreed on separating host, member and e2e Roles/ClusterRoles. So we will have:
- host Roles (already defined in https://github.com/codeready-toolchain/host-operator/tree/master/deploy)
- member Roles (already defined in https://github.com/codeready-toolchain/member-operator/tree/master/deploy)
- e2e roles (can be two roles, one for the member cluster and one for the host, or just one pair of role/clusterrole used in both e2e SAs (host & member)) with all the permissions the e2e pod could need instead of reusing the OLM SAs for that.

PR #181, opened 2020-10-14
Create pods in a loop until success

alexeykazakov:
Our Idler test is currently flaky because pod (or pod controller) creation fails from time to time due to this upstream issue: https://github.com/kubernetes/kubernetes/issues/67761

In this PR we create pods in a loop until success.

### DIFF
--- a/wait/member.go
+++ b/wait/member.go
@@ -6,6 +6,8 @@ import (
 	"reflect"
 	"testing"
 
+	"k8s.io/apimachinery/pkg/runtime"
+
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/pkg/apis/toolchain/v1alpha1"
 	"github.com/codeready-toolchain/toolchain-common/pkg/cluster"
 	"github.com/codeready-toolchain/toolchain-common/pkg/test"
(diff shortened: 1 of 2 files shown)


wait/member.go
```diff
@@ -430,6 +432,19 @@ func (a *MemberAwaitility) UpdateIdlerSpec(idler *toolchainv1alpha1.Idler) (*too
 	return result, err
 }
 
+// Create tries to create the object until success
+// Workaround for https://github.com/kubernetes/kubernetes/issues/67761
+func (a *MemberAwaitility) Create(obj runtime.Object) error {
+	err := wait.Poll(a.RetryInterval, a.Timeout, func() (done bool, err error) {
+		if err := a.Client.Create(context.TODO(), obj); err != nil {
```
alexeykazakov, line 439: I'm not sure but maybe we need to use a deep copy here (`obj.DeepCopy()`) instead of the original `obj`. We may need that if `Create()` actually modifies `obj` even if it failed with an error. Let's see how it works and if we see that issue we can start using a deep copy instead.

MatousJobanek (approved):
Looks good :+1:

PR #187, opened 2020-11-02
fix: ensure auto-approval is disabled when starting TestE2EFlow

xcoulon:


MatousJobanek (approved):
Just out of curiosity - what is this good for? We already have tests for automatic approval and those tests should not affect this one...

xcoulon:
> Just out of curiosity - what is this good for? We already have tests for automatic approval and those tests should not affect this one...

because I got an issue where the user approval was configured to `automatic` after a previous run, and when running the e2e tests with the current deployment/config, some assertions would fail

PR #190, opened 2020-11-06
load activity

xcoulon:
- fix: ensure auto-approval is disabled when starting TestE2EFlow
- WIP - invalid openshift/client-go version
- feat: CLI utility to load users


xcoulon:
and with a README! https://github.com/codeready-toolchain/toolchain-e2e/blob/17f9902a6df739ccb1034df1a2a7ac51f2628fe1/test/load/README.adoc

test/load/cmd/root_cmd.go
```diff
@@ -0,0 +1,302 @@
...
+	for {
+		select {
+		case <-signupChan:
+			signups++
+		case <-deactivationChan:
+			deactivations++
+		case <-banChan:
+			bans++
+		case <-errorChan:
+			bans++
```
rajivnathan, line 298: Should be `errors++`?

test/load/cmd/client.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/mitchellh/go-homedir"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/client-go/rest"
+	restclient "k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+func newClient() (*restclient.RESTClient, *restclient.Config, error) {
+	kubeconfig, err := contextConfig()
```
rajivnathan, line 21: Should we rename this `kubeconfig` function variable to something else since there's a global variable also named `kubeconfig`? Maybe `clientconfig` or something?

rajivnathan (changes requested):
This will be very handy!! 🙌  Added a couple comments. Also, since this load package doesn't actually run any tests should we maybe move it out of the test package and have it as a top-level package?

test/load/cmd/root_cmd.go
```diff
@@ -0,0 +1,302 @@
...
+			Labels: map[string]string{
+				toolchainv1alpha1.UserSignupUserEmailHashLabelKey: emailHash,
+			},
+		},
+		Spec: toolchainv1alpha1.UserSignupSpec{
+			Username:             username,
+			Approved:             true,
+			VerificationRequired: false,
+		},
+	}
```
MatousJobanek, line 164: Looking at this I think that we should start using some unified function for creating the `UserSignups` - it's yet another place where we create the `UserSignup` with all annotations & labels and it's pretty hard to keep it consistent. 
We could maybe place it in api repo, WDYT?
Can be done as separate PR(effort) of course :wink:

PR #197, opened 2020-11-24
e2e for usersignup reactivation in reg-service

alexeykazakov:
E2e tests for https://github.com/codeready-toolchain/registration-service/pull/144

Also I refactored the reg-service tests a bit to get rid of a lot of code duplication.


MatousJobanek (approved):
Nice cleanup :star_struck: 
and good solution with the reactivation test :+1:

PR #207, opened 2020-12-02
Update tests for basic-no-deactivation tier

rajivnathan:
Tests for https://github.com/codeready-toolchain/host-operator/pull/329

- Updates the nstemplatetier tests to include the new `basic-no-deactivation` tier.
- Adds a deactivation test to check that users in tiers without deactivation are indeed not deactivated.

### DIFF
--- a/test/e2e/user_management_test.go
+++ b/test/e2e/user_management_test.go
@@ -128,9 +128,9 @@ func (s *userManagementTestSuite) TestUserDeactivation() {
 		})
 	})
 
-	s.T().Run("tests for basic tier with deactivation disabled", func(t *testing.T) {
+	s.T().Run("tests for tiers with automatic deactivation disabled", func(t *testing.T) {
 		// Let's create a tier with deactivation disabled
-		noDeactivationTier := CreateNSTemplateTier(t, s.ctx, s.hostAwait, "deactivation-tier")
+		noDeactivationTier := CreateNSTemplateTier(t, s.ctx, s.hostAwait, "no-deactivation-tier")
 
 		// Move the user to the new tier without deactivation enabled
 		murSyncIndex := MoveUserToTier(t, s.hostAwait, userSignup.Spec.Username, *noDeactivationTier).Spec.UserAccounts[0].SyncIndex
--- a/test/e2e/user_management_test.go
+++ b/test/e2e/user_management_test.go
@@ -172,9 +172,9 @@ func (s *userManagementTestSuite) TestUserDeactivation() {
 			// Only the user with domain not on the exclusion list should be auto-deactivated
 			s.hostAwait.WaitUntilMetricHasValue("sandbox_user_signups_total", baseUserSignups+2)
 			s.hostAwait.WaitUntilMetricHasValue("sandbox_user_signups_approved_total", baseUserSignupsApproved+3)
-			s.hostAwait.WaitUntilMetricHasValue("sandbox_master_user_record_current", baseCurrentMURs+1)
-			s.hostAwait.WaitUntilMetricHasValue("sandbox_user_signups_deactivated_total", baseUserSignupsDeactivated+2)
-			s.hostAwait.WaitUntilMetricHasValue("sandbox_user_signups_auto_deactivated_total", baseUserSignupsAutoDeactivated+1)
+			s.hostAwait.WaitUntilMetricHasValue("sandbox_master_user_record_current", baseCurrentMURs+2)
+			s.hostAwait.WaitUntilMetricHasValue("sandbox_user_signups_deactivated_total", baseUserSignupsDeactivated+1)
+			s.hostAwait.WaitUntilMetricHasValue("sandbox_user_signups_auto_deactivated_total", baseUserSignupsAutoDeactivated)
 		})
 	})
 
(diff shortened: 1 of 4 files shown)


alexeykazakov (commented):
Looks good! But we need more tests :)
We need to provision a user with this new tier and make sure it creates all required resources.
See https://github.com/codeready-toolchain/toolchain-e2e/blob/47f5284792e3a285484f002fec449c67e35145e8/tiers/checks.go#L31 and how it's used.

tiers/checks.go
```diff
@@ -33,6 +33,10 @@ func NewChecks(tier string) (TierChecks, error) {
 	case "basic":
 		return &basicTierChecks{}, nil
 
+	case "basic-no-deactivation":
+		// we want the basic-no-deactivation tier to have the same resources as the basic tier with the only difference being auto deactivation disabled
```
rajivnathan, line 37: @alexeykazakov is it safe to assume we want the 'basic-no-deactivation' tier to be consistent with the 'basic' tier in terms of the resources so using the same checks is what we want?

alexeykazakov: yes

alexeykazakov: this is why it's called `basic-no-deactivation` ;)

test/e2e/user_management_test.go
```diff
@@ -128,6 +128,55 @@ func (s *userManagementTestSuite) TestUserDeactivation() {
 		})
 	})
 
+	s.T().Run("tests for basic tier with deactivation disabled", func(t *testing.T) {
```
MatousJobanek, line 131: hmm... shouldn't it be rather in `nstemplatetier_test.go` file - we want to verify the functionality of the new tier, right?

test/e2e/user_management_test.go
```diff
@@ -128,6 +128,55 @@ func (s *userManagementTestSuite) TestUserDeactivation() {
...
+		// Let's create a tier with deactivation disabled
+		noDeactivationTier := CreateNSTemplateTier(t, s.ctx, s.hostAwait, "deactivation-tier")
```
MatousJobanek, line 133: But why? You just added a tier `basic-no-deactivation` that has the deactivation disabled - you don't have to create a new one.

tiers/checks.go
MatousJobanek: btw, I'm a bit surprised that we verify everything from the tier, but not the deactivation information. 
Shouldn't we verify somewhere also the field `DeactivationTimeoutDays` of the created NSTemplateTiers?

alexeykazakov: Well.. the current tier checks verify namespace and cluster resources created by the tiers only.
Deactivation configuration is part of the tier resource itself and tested indirectly in our deactivation tests.
But I guess we should add a new tier check type. Currently we have only `namespaceObjectsCheck` and `clusterObjectsCheck`. Let's add something like `tierObjectsCheck` which will verify the Tier resource itself so we can make sure only `basic` tier has deactivation enabled and the rest of the tiers do not have it.

test/e2e/user_management_test.go
rajivnathan: Not sure, I think you could argue it both ways. The way I thought about it is tiers are more like part of the configuration, the tiers contain values we can configure but the actual behaviour is handled by the controllers that react on those values. So I figured the tier tests validate that the values we're setting are within some expected parameters and the other tests are testing the functionality. So for user management which includes Banning, Deactivating and Disabling users I figured it made sense to test the automatic deactivation here as well.

test/e2e/user_management_test.go
rajivnathan: Somewhat related to the comment above, I figured we wanted to test the general functionality and not have it dependent on the existence of any tiers?

alexeykazakov: Well... since it's e2e tests the main goal of these tests is the actual functionality available to the users. We don't generally care about the implementation details.
Basically some test which verifies what happens to the user who is provisioned with the basic-no-deactivation tier (the user should be deactivated) is more valuable than testing an abstract deactivation mechanism of some non-existing tier.
It's not always possible though and sometimes we have to do some tricks like copying tiers. But those tricks are just workarounds of technical limitations we face in the e2e environment.

It's different in unit tests.

test/e2e/user_management_test.go
MatousJobanek: ok, fair enough :+1: keep it here :-)

tiers/checks.go
rajivnathan: Thanks for the good point, I've added `tierObjectCheck` to validate the `DeactivationTimeoutDays`

test/e2e/user_management_test.go
rajivnathan: Sounds good. I've updated the test to just use the real tiers `basic` and `basicdeactivationdisabled`

rajivnathan:
@alexeykazakov @MatousJobanek Thanks for the review! 🙌  I think I've addressed the comments and finally got the tests passing again!

test/e2e/user_management_test.go
MatousJobanek: I agree with @alexeykazakov 
In this care, we need to verify the actual configuration of the Tier, not the general deactivation logic.
In addition, we can easily use the tier without creating a new one - in other words, the workaround mentioned by Alexey is not needed

alexeykazakov (approved):
Nice work!

PR #208, opened 2020-12-03
fix: improve logging & checks of member operator webhook

MatousJobanek:
* print the member-operator-webhook log when tests fail
* print info while checking resources
* wait until the webhook deployment is ready

### DIFF
--- a/wait/member.go
+++ b/wait/member.go
@@ -6,6 +6,7 @@ import (
 	"reflect"
 	"testing"
 
+	"github.com/codeready-toolchain/toolchain-common/pkg/status"
 	"github.com/stretchr/testify/assert"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/intstr"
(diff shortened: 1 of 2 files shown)


wait/member.go
```diff
@@ -966,6 +986,7 @@ func (a *MemberAwaitility) waitForSecret() []byte {
 }
 
 func (a *MemberAwaitility) waitForWebhookConfig(ca []byte) {
+	a.T.Logf("checking prensence of MutatingWebhookConfiguration resource '%s'", "sandbox-users-pods")
```
alexeykazakov, line 989: Isn't it `member-operator-webhook`?
```suggestion
	a.T.Logf("checking prensence of MutatingWebhookConfiguration resource '%s'", "member-operator-webhook")
```
or even:
```suggestion
	a.T.Logf("checking prensence of MutatingWebhookConfiguration resource 'member-operator-webhook'")
```

wait/member.go
```diff
@@ -901,16 +903,20 @@ func (a *MemberAwaitility) waitForResource(namespace, name string, object runtim
...
+			a.T.Logf("unexpected error when looking for resource '%s' in namesapace '%s'", name, namespace)
 			return false, err
 		}
+		a.T.Logf("resource '%s' in namesapace '%s' found", name, namespace)
 		return true, nil
 	})
 	require.NoError(a.T, err)
 }
 
 func (a *MemberAwaitility) waitForService() {
```
alexeykazakov, line 918: Btw I think we should rename all these web hook specific waits to `waitForWebHookService()`, `waitForWebHookDeployment()`, etc instead of `waitForDeployment()` to avoid confusion since it's not the only deployment which is running in the member.

alexeykazakov (approved):

PR #219, opened 2021-01-08
Fix: Adding GenerateName to Notification Name

tinakurian:
Relates to: https://github.com/codeready-toolchain/host-operator/pull/356
Jira :https://issues.redhat.com/browse/CRT-893

As a user, if I get deactivated twice in less than 24hrs, my account gets stuck in a deactivating status and the usersignup has the following status condition:

- lastTransitionTime: "2020-12-09T13:23:24Z"
    message: notifications.toolchain.dev.openshift.com "xcoulon-deactivated" already exists
We should add a timestamp suffix on the notification resource name to avoid such conflicts.

To solve this problem, I have changed the Name field to GenerateName. I have also added two new labels to the api repo for notification_types.go. Subsequently, I had to also change GET notifications to LIST notifications since more than one notification of a specific type is now permitted.

### DIFF
--- a/test/e2e/user_management_test.go
+++ b/test/e2e/user_management_test.go
@@ -44,6 +44,53 @@ func (s *userManagementTestSuite) TearDownTest() {
 func (s *userManagementTestSuite) TestUserDeactivation() {
 	s.hostAwait.UpdateHostOperatorConfig(test.AutomaticApproval().Enabled())
 
+	s.T().Run("create two deactivation notifications", func(t *testing.T) {
+
+		userSignup, mur := s.createAndCheckUserSignup(true, "twodeactivationnotifications", "twodeactivationnotifications@redhat.com", true, ApprovedByAdmin()...)
+
+		userSignup, err := s.hostAwait.UpdateUserSignupSpec(userSignup.Name, func(us *v1alpha1.UserSignup) {
+			us.Spec.Deactivated = true
+		})
+		require.NoError(s.T(), err)
+		s.T().Logf("user signup '%s' set to deactivated", userSignup.Name)
+		err = s.hostAwait.WaitUntilMasterUserRecordDeleted(mur.Name)
+		require.NoError(s.T(), err)
+
+		userSignup, err = s.hostAwait.UpdateUserSignupSpec(userSignup.Name, func(us *v1alpha1.UserSignup) {
+			us.Spec.Deactivated = false
+		})
+		require.NoError(s.T(), err)
+		s.T().Logf("user signup '%s' set to reactivated", userSignup.Name)
+		_, err = s.hostAwait.WaitForMasterUserRecord(mur.Name)
+		require.NoError(s.T(), err)
+
+		userSignup, err = s.hostAwait.UpdateUserSignupSpec(userSignup.Name, func(us *v1alpha1.UserSignup) {
+			us.Spec.Deactivated = true
+		})
+		require.NoError(s.T(), err)
+		s.T().Logf("user signup '%s' set to deactivated", userSignup.Name)
+		err = s.hostAwait.WaitUntilMasterUserRecordDeleted(mur.Name)
+		require.NoError(s.T(), err)
+
+		// "deactivated"
+		notifications, err := s.hostAwait.WaitForNotifications(userSignup.Status.CompliantUsername, "-deactivated", wait.UntilNotificationHasConditions(Sent()))
+		require.NoError(t, err)
+		require.NotEmpty(t, notifications)
+		require.Equal(t, 2, len(notifications))
+
+		notification0 := notifications[0]
+		assert.Contains(t, notification0.Name, userSignup.Status.CompliantUsername+"-deactivated")
+		assert.Equal(t, userSignup.Namespace, notification0.Namespace)
+		assert.Equal(t, "userdeactivated", notification0.Spec.Template)
+		assert.Equal(t, userSignup.Name, notification0.Spec.UserID)
+
+		notification1 := notifications[1]
+		assert.Contains(t, notification1.Name, userSignup.Status.CompliantUsername+"-deactivated")
+		assert.Equal(t, userSignup.Namespace, notification1.Namespace)
+		assert.Equal(t, "userdeactivated", notification1.Spec.Template)
+		assert.Equal(t, userSignup.Name, notification0.Spec.UserID)
+	})
+
 	s.T().Run("deactivate a user", func(t *testing.T) {
 		// Initialize metrics assertion counts
 		metricsAssertion := InitMetricsAssertion(s.T(), s.hostAwait)
@@ -70,12 +117,12 @@ func (s *userManagementTestSuite) TestUserDeactivation() {
 		notifications, err := s.hostAwait.WaitForNotifications(userSignup.Status.CompliantUsername, "-deactivated", wait.UntilNotificationHasConditions(Sent()))
 		require.NoError(t, err)
 		require.NotEmpty(t, notifications)
-		for _, notification := range notifications {
-			assert.Contains(t, notification.Name, userSignup.Status.CompliantUsername+"-deactivated")
-			assert.Equal(t, userSignup.Namespace, notification.Namespace)
-			assert.Equal(t, "userdeactivated", notification.Spec.Template)
-			assert.Equal(t, userSignup.Name, notification.Spec.UserID)
-		}
+		require.Equal(t, 1, len(notifications))
+		notification := notifications[0]
+		assert.Contains(t, notification.Name, userSignup.Status.CompliantUsername+"-deactivated")
+		assert.Equal(t, userSignup.Namespace, notification.Namespace)
+		assert.Equal(t, "userdeactivated", notification.Spec.Template)
+		assert.Equal(t, userSignup.Name, notification.Spec.UserID)
 
 		err = s.hostAwait.WaitUntilNotificationDeleted(userSignup.Status.CompliantUsername, "-deactivated")
 		require.NoError(t, err)
--- a/test/e2e/user_management_test.go
+++ b/test/e2e/user_management_test.go
@@ -67,7 +67,7 @@ func (s *userManagementTestSuite) TestUserDeactivation() {
 		require.NoError(s.T(), err)
 
 		// "deactivated"
-		notifications, err := s.hostAwait.WaitForNotifications(userSignup.Status.CompliantUsername, "-deactivated", wait.UntilNotificationHasConditions(Sent()))
+		notifications, err := s.hostAwait.WaitForNotifications(userSignup.Status.CompliantUsername, "deactivated", wait.UntilNotificationHasConditions(Sent()))
 		require.NoError(t, err)
 		require.NotEmpty(t, notifications)
 		require.Equal(t, 1, len(notifications))
--- a/test/e2e/user_management_test.go
+++ b/test/e2e/user_management_test.go
@@ -67,12 +67,12 @@ func (s *userManagementTestSuite) TestUserDeactivation() {
 		require.NoError(s.T(), err)
 
 		// "deactivated"
-		notifications, err := s.hostAwait.WaitForNotifications(userSignup.Status.CompliantUsername, v1alpha1.NotificationTypeDeactivated, wait.UntilNotificationHasConditions(Sent()))
+		notifications, err := s.hostAwait.WaitForNotifications(userSignup.Status.CompliantUsername, v1alpha1.NotificationTypeDeactivated, 1, wait.UntilNotificationHasConditions(Sent()))
 		require.NoError(t, err)
 		require.NotEmpty(t, notifications)
 		require.Equal(t, 1, len(notifications))
 		notification := notifications[0]
-		assert.Contains(t, notification.Name, userSignup.Status.CompliantUsername+"-deactivated")
+		assert.Contains(t, notification.Name, userSignup.Status.CompliantUsername+"-deactivated-")
 		assert.Equal(t, userSignup.Namespace, notification.Namespace)
 		assert.Equal(t, "userdeactivated", notification.Spec.Template)
 		assert.Equal(t, userSignup.Name, notification.Spec.UserID)
--- a/test/e2e/user_management_test.go
+++ b/test/e2e/user_management_test.go
@@ -67,7 +67,7 @@ func (s *userManagementTestSuite) TestUserDeactivation() {
 		require.NoError(s.T(), err)
 
 		// "deactivated"
-		notifications, err := s.hostAwait.WaitForNotifications(userSignup.Status.CompliantUsername, "deactivated", wait.UntilNotificationHasConditions(Sent()))
+		notifications, err := s.hostAwait.WaitForNotifications(userSignup.Status.CompliantUsername, v1alpha1.NotificationTypeDeactivated, wait.UntilNotificationHasConditions(Sent()))
 		require.NoError(t, err)
 		require.NotEmpty(t, notifications)
 		require.Equal(t, 1, len(notifications))
@@ -77,7 +77,7 @@ func (s *userManagementTestSuite) TestUserDeactivation() {
 		assert.Equal(t, "userdeactivated", notification.Spec.Template)
 		assert.Equal(t, userSignup.Name, notification.Spec.UserID)
 
-		err = s.hostAwait.WaitUntilNotificationDeleted(userSignup.Status.CompliantUsername, "deactivated")
+		err = s.hostAwait.WaitUntilNotificationDeleted(userSignup.Status.CompliantUsername, v1alpha1.NotificationTypeDeactivated)
 		require.NoError(t, err)
 
 		userSignup, err = s.hostAwait.WaitForUserSignup(userSignup.Name,
--- a/test/e2e/user_management_test.go
+++ b/test/e2e/user_management_test.go
@@ -44,53 +44,6 @@ func (s *userManagementTestSuite) TearDownTest() {
 func (s *userManagementTestSuite) TestUserDeactivation() {
 	s.hostAwait.UpdateHostOperatorConfig(test.AutomaticApproval().Enabled())
 
-	s.T().Run("create two deactivation notifications", func(t *testing.T) {
-
-		userSignup, mur := s.createAndCheckUserSignup(true, "twodeactivationnotifications", "twodeactivationnotifications@redhat.com", true, ApprovedByAdmin()...)
-
-		userSignup, err := s.hostAwait.UpdateUserSignupSpec(userSignup.Name, func(us *v1alpha1.UserSignup) {
-			us.Spec.Deactivated = true
-		})
-		require.NoError(s.T(), err)
-		s.T().Logf("user signup '%s' set to deactivated", userSignup.Name)
-		err = s.hostAwait.WaitUntilMasterUserRecordDeleted(mur.Name)
-		require.NoError(s.T(), err)
-
-		userSignup, err = s.hostAwait.UpdateUserSignupSpec(userSignup.Name, func(us *v1alpha1.UserSignup) {
-			us.Spec.Deactivated = false
-		})
-		require.NoError(s.T(), err)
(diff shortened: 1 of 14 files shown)


test/e2e/user_management_test.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"io/ioutil"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
xcoulon, line 7: this `import` statement shouldn't have been moved in the first section (usually reserved for standard packages)

wait/host.go
```diff
@@ -587,24 +587,32 @@ func (a *HostAwaitility) WaitForTemplateUpdateRequests(namespace string, count i
 type NotificationWaitCriterion func(a *HostAwaitility, mur *toolchainv1alpha1.Notification) bool
 
 // WaitForNotification waits until there is a Notification available with the given name and the optional conditions
-func (a *HostAwaitility) WaitForNotification(name string, criteria ...NotificationWaitCriterion) (*toolchainv1alpha1.Notification, error) {
-	var notification *toolchainv1alpha1.Notification
+func (a *HostAwaitility) WaitForNotification(name string, criteria ...NotificationWaitCriterion) ([]*toolchainv1alpha1.Notification, error) {
```
xcoulon, line 590: since this func will return a `[]*toolchainv1alpha1.Notification`, shouldn't we rename it to `WaitForNotifications` (plural)?

wait/host.go
```diff
@@ -587,24 +587,32 @@ func (a *HostAwaitility) WaitForTemplateUpdateRequests(namespace string, count i
...
+	var notification []*toolchainv1alpha1.Notification
```
xcoulon, line 591:
```suggestion
	var notifications []*toolchainv1alpha1.Notification
```
(change to `notifications` will need to be applied further down in this func)

wait/host.go
```diff
@@ -587,24 +587,32 @@ func (a *HostAwaitility) WaitForTemplateUpdateRequests(namespace string, count i
...
 	err := wait.Poll(a.RetryInterval, a.Timeout, func() (done bool, err error) {
-		obj := &toolchainv1alpha1.Notification{}
-		if err := a.Client.Get(context.TODO(), types.NamespacedName{Namespace: a.Namespace, Name: name}, obj); err != nil {
+		obj := &toolchainv1alpha1.NotificationList{}
+		if err := a.Client.List(context.TODO(), obj); err != nil {
```
xcoulon, line 594: what about adding a label on the notifications so we can associated them with the user?

wait/host.go
```diff
@@ -587,24 +587,32 @@ func (a *HostAwaitility) WaitForTemplateUpdateRequests(namespace string, count i
...
 				a.T.Logf("waiting for availability of notification '%s'", name)
 				return false, nil
 			}
 			return false, err
 		}
-		for _, match := range criteria {
-			if !match(a, obj) {
-				return false, nil
+
+		for _, n := range obj.Items {
```
xcoulon, line 602: with a label, we would not need such an iteration loop

wait/host.go
```diff
@@ -613,16 +621,23 @@ func (a *HostAwaitility) WaitForNotification(name string, criteria ...Notificati
 // WaitUntilNotificationDeleted waits until the Notification with the given name is deleted (ie, not found)
 func (a *HostAwaitility) WaitUntilNotificationDeleted(name string) error {
 	return wait.Poll(a.RetryInterval, a.Timeout, func() (done bool, err error) {
-		notification := &toolchainv1alpha1.Notification{}
-		if err := a.Client.Get(context.TODO(), types.NamespacedName{Namespace: a.Namespace, Name: name}, notification); err != nil {
+		obj := &toolchainv1alpha1.NotificationList{}
+		if err := a.Client.List(context.TODO(), obj); err != nil {
```
xcoulon, line 625: same question here about adding a label on the notifications.

wait/host.go
tinakurian: a label is a good idea, i'll go ahead and add that! thanks for the suggestion Xavier!

test/e2e/user_management_test.go
tinakurian: yeah, the editor was trying to be helpful and i didn't catch it before pushing! will fix this!

wait/host.go
```diff
@@ -586,43 +586,60 @@ func (a *HostAwaitility) WaitForTemplateUpdateRequests(namespace string, count i
...
 			}
 			return false, err
 		}
-		for _, match := range criteria {
-			if !match(a, obj) {
-				return false, nil
+
+		for _, n := range notificationList.Items {
+			for _, match := range criteria {
+				if match(a, &n) {
```
MatousJobanek, line 606: Just a detail - do we really need to pass it as a pointer? This is the only usage of the `NotificationWaitCriterion` type right?

wait/host.go
```diff
@@ -586,43 +586,60 @@ func (a *HostAwaitility) WaitForTemplateUpdateRequests(namespace string, count i
...
 				a.T.Logf("Notification has been deleted '%s'", name)
 				return true, nil
 			}
 			return false, err
 		}
-		a.T.Logf("waiting until Notification is deleted '%s'", name)
-		return false, nil
+
+		if len(notificationList.Items) > 0 {
+			a.T.Logf("waiting until Notification is deleted '%s'", notificationList.Items[0].Name)
```
MatousJobanek, line 636: I would rather print all of them, not only the first one

testsupport/user_setup.go
```diff
@@ -78,15 +78,19 @@ func CreateAndApproveSignup(t *testing.T, hostAwait *wait.HostAwaitility, userna
 	require.NoError(t, err)
 
 	// Wait for the the notification CR to be created & sent
-	notification, err := hostAwait.WaitForNotification(mur.Name+"-provisioned", wait.UntilNotificationHasConditions(Sent()))
+	// "-provisioned"
```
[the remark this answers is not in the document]
tinakurian: thanks for catching that!

wait/host.go
tinakurian: Fixed here: 258e10b3c4259d21be1b08d46f8b59ed0cced242 and as part of: 72d3c5c66de1c858dfee03ed84878e0fe86996de

wait/host.go
tinakurian: Yeah, not too sure why it was a pointer, was just re-using what was already there. But i've changed it as part of this commit: 72d3c5c66de1c858dfee03ed84878e0fe86996de

wait/host.go
tinakurian: Good point! Updated it as part of this commit: fd66ead0da803a9206a4e0326c56ae73dec1d2a5

wait/host.go
tinakurian: Good point! Updated it as part of this commit: fd66ead0da803a9206a4e0326c56ae73dec1d2a5

wait/host.go
tinakurian: Much cleaner! Updated it as part of this commit: 155dc96be86fca8f5b90c428ae30601a1c0cf8a4

wait/host.go
tinakurian: Updated as part of this commit: fa327b0e280198b8ac1cc89c5a1053a4763aba95

testsupport/user_setup.go
```diff
@@ -78,15 +78,18 @@ func CreateAndApproveSignup(t *testing.T, hostAwait *wait.HostAwaitility, userna
...
+	notifications, err := hostAwait.WaitForNotifications(mur.Name, toolchainv1alpha1.NotificationTypeProvisioned, wait.UntilNotificationHasConditions(Sent()))
 	require.NoError(t, err)
-	require.NotNil(t, notification)
-	assert.Equal(t, mur.Name+"-provisioned", notification.Name)
-	assert.Equal(t, mur.Namespace, notification.Namespace)
-	assert.Equal(t, "userprovisioned", notification.Spec.Template)
-	assert.Equal(t, mur.Spec.UserID, notification.Spec.UserID)
+	require.NotEmpty(t, notifications)
+	for _, notification := range notifications {
+		assert.Contains(t, notification.Name, mur.Name+"-provisioned")
```
alexeykazakov, line 85:
```suggestion
		assert.Contains(t, notification.Name, mur.Name+"-provisioned-")
```

wait/host.go
```diff
@@ -584,51 +584,69 @@ func (a *HostAwaitility) WaitForTemplateUpdateRequests(namespace string, count i
...
 
 // NotificationWaitCriterion checks if a Notification meets the given condition
-type NotificationWaitCriterion func(a *HostAwaitility, mur *toolchainv1alpha1.Notification) bool
+type NotificationWaitCriterion func(a *HostAwaitility, mur toolchainv1alpha1.Notification) bool
 
-// WaitForNotification waits until there is a Notification available with the given name and the optional conditions
-func (a *HostAwaitility) WaitForNotification(name string, criteria ...NotificationWaitCriterion) (*toolchainv1alpha1.Notification, error) {
-	var notification *toolchainv1alpha1.Notification
+// WaitForNotifications waits until there is a Notification available with the given name and the optional conditions
+func (a *HostAwaitility) WaitForNotifications(notificationName, notificationType string, criteria ...NotificationWaitCriterion) ([]*toolchainv1alpha1.Notification, error) {
```
alexeykazakov, line 590: The function name and description is confusing. It returns an empty slice if there is no notifications at all. In that case it won't wait for anything.
IMO we should add a new parameter for the number of expected notifications. Something like:
```suggestion
// WaitForNotifications waits until there is an expected number of Notifications available for the provided user and with the notification type and which match the conditions (if provided).
func (a *HostAwaitility) WaitForNotifications(username, numberOfNotifications int, notificationType string, criteria ...NotificationWaitCriterion) ([]*toolchainv1alpha1.Notification, error) {
```

wait/host.go
```diff
@@ -584,51 +584,69 @@ func (a *HostAwaitility) WaitForTemplateUpdateRequests(namespace string, count i
...
+	var notifications []*toolchainv1alpha1.Notification
 	err := wait.Poll(a.RetryInterval, a.Timeout, func() (done bool, err error) {
-		obj := &toolchainv1alpha1.Notification{}
-		if err := a.Client.Get(context.TODO(), types.NamespacedName{Namespace: a.Namespace, Name: name}, obj); err != nil {
+		labels := map[string]string{toolchainv1alpha1.NotificationUserNameLabelKey: notificationName, toolchainv1alpha1.NotificationTypeLabelKey: notificationType}
+		opts := client.MatchingLabels(labels)
+		notificationList := &toolchainv1alpha1.NotificationList{}
+		if err :=  a.Client.List(context.TODO(), notificationList, opts);  err != nil {
 			if errors.IsNotFound(err) {
```
alexeykazakov, line 597: I don't think `List()` can return NotFound. Check the number of the found notifications (which match the condition if provided! If not provided the check the overall number of the found notifications).

wait/host.go
```diff
@@ -584,51 +584,69 @@ func (a *HostAwaitility) WaitForTemplateUpdateRequests(namespace string, count i
...
+		}
 		return true, nil
 	})
-	return notification, err
+	return notifications, err
 }
 
 // WaitUntilNotificationDeleted waits until the Notification with the given name is deleted (ie, not found)
-func (a *HostAwaitility) WaitUntilNotificationDeleted(name string) error {
+func (a *HostAwaitility) WaitUntilNotificationDeleted(notificationName, notificationType string) error {
```
alexeykazakov, line 622:
```suggestion
func (a *HostAwaitility) WaitUntilNotificationDeleted(username, notificationType string) error {
```

wait/host.go
```diff
@@ -584,51 +584,69 @@ func (a *HostAwaitility) WaitForTemplateUpdateRequests(namespace string, count i
...
+		if len(notifications) <= 0 {
+			return false, nil
+		}
 		return true, nil
 	})
-	return notification, err
+	return notifications, err
 }
 
 // WaitUntilNotificationDeleted waits until the Notification with the given name is deleted (ie, not found)
```
alexeykazakov, line 621:
```suggestion
// WaitUntilNotificationDeleted waits until the Notification for the given user is deleted (ie, not found)
```

wait/host.go
```diff
@@ -584,51 +584,69 @@ func (a *HostAwaitility) WaitForTemplateUpdateRequests(namespace string, count i
...
-func (a *HostAwaitility) WaitUntilNotificationDeleted(name string) error {
+func (a *HostAwaitility) WaitUntilNotificationDeleted(notificationName, notificationType string) error {
 	return wait.Poll(a.RetryInterval, a.Timeout, func() (done bool, err error) {
-		notification := &toolchainv1alpha1.Notification{}
-		if err := a.Client.Get(context.TODO(), types.NamespacedName{Namespace: a.Namespace, Name: name}, notification); err != nil {
+		labels := map[string]string{toolchainv1alpha1.NotificationUserNameLabelKey: notificationName, toolchainv1alpha1.NotificationTypeLabelKey: notificationType}
+		opts := client.MatchingLabels(labels)
+		notificationList := &toolchainv1alpha1.NotificationList{}
+		if err :=  a.Client.List(context.TODO(), notificationList, opts);  err != nil {
 			if errors.IsNotFound(err) {
```
alexeykazakov, line 628: List doesn't return NotFound. Check if the list is empty instead.

alexeykazakov (changes requested):

testsupport/user_setup.go
tinakurian: Done as part of: 1ddb1a3c4a441b524ea8e21f03903f6dce26147e

wait/host.go
tinakurian: Fixed as part of: 1ddb1a3c4a441b524ea8e21f03903f6dce26147e

wait/host.go
tinakurian: Updated here: 1ddb1a3c4a441b524ea8e21f03903f6dce26147e

wait/host.go
tinakurian: Updated as part of: 1ddb1a3c4a441b524ea8e21f03903f6dce26147e

wait/host.go
tinakurian: Fixed as part of: 1ddb1a3c4a441b524ea8e21f03903f6dce26147e

test/e2e/user_management_test.go
```diff
@@ -66,15 +66,18 @@ func (s *userManagementTestSuite) TestUserDeactivation() {
...
-		notification, err := s.hostAwait.WaitForNotification(userSignup.Status.CompliantUsername+"-deactivated", wait.UntilNotificationHasConditions(Sent()))
+		// "deactivated"
+		notifications, err := s.hostAwait.WaitForNotifications(userSignup.Status.CompliantUsername, v1alpha1.NotificationTypeDeactivated, wait.UntilNotificationHasConditions(Sent()))
 		require.NoError(t, err)
-		require.NotNil(t, notification)
-		assert.Equal(t, userSignup.Status.CompliantUsername+"-deactivated", notification.Name)
+		require.NotEmpty(t, notifications)
+		require.Equal(t, 1, len(notifications))
+		notification := notifications[0]
+		assert.Contains(t, notification.Name, userSignup.Status.CompliantUsername+"-deactivated")
```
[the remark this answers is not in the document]
tinakurian: Updated here: 1ddb1a3c4a441b524ea8e21f03903f6dce26147e

wait/host.go
tinakurian: Updated comment here: 6857dc3b128382b4fc1feef6eaaa85232b22c933
Updated parameter here: 1ddb1a3c4a441b524ea8e21f03903f6dce26147e

wait/host.go
```diff
@@ -584,51 +585,77 @@ func (a *HostAwaitility) WaitForTemplateUpdateRequests(namespace string, count i
...
+		notificationList := &toolchainv1alpha1.NotificationList{}
+		if err :=  a.Client.List(context.TODO(), notificationList, opts);  err != nil {
 			return false, err
 		}
-		for _, match := range criteria {
-			if !match(a, obj) {
-				return false, nil
+
+		actualNotificationCount := len(notificationList.Items)
+		if numberOfNotifications < actualNotificationCount{
```
xcoulon, line 602: instead of having 2 blocks to compare the actual number of notifications vs the expected one, you could probably just do:
```
		if numberOfNotifications != actualNotificationCount {
			a.T.Logf("expected '%d' notifications, but found '%d' notifications", numberOfNotifications, actualNotificationCount)
			return false, errs.New(fmt.Sprintf("expected %d notifications, but found %d notifications", numberOfNotifications, actualNotificationCount))
		}
```

wait/host.go
```diff
@@ -584,51 +585,77 @@ func (a *HostAwaitility) WaitForTemplateUpdateRequests(namespace string, count i
...
+					notifications = append(notifications, &n)
+				}
 			}
 		}
-		a.T.Logf("found notification '%s'", name)
-		notification = obj
+
+		// The criteria hasn't matched yet
+		if len(notifications) <= 0 {
+			return false, nil
```
xcoulon, line 626: it's probably worth having a message here as well, to explain that some of the notifications did not match the criteria

wait/host.go
```diff
@@ -584,51 +585,77 @@ func (a *HostAwaitility) WaitForTemplateUpdateRequests(namespace string, count i
...
+					a.T.Logf("found notification '%s'", username)
+					notifications = append(notifications, &n)
+				}
 			}
 		}
-		a.T.Logf("found notification '%s'", name)
-		notification = obj
+
+		// The criteria hasn't matched yet
+		if len(notifications) <= 0 {
```
xcoulon, line 625: you can't have `len(notifications) <= 0`, can you? ;)

wait/host.go
```diff
@@ -584,51 +585,77 @@ func (a *HostAwaitility) WaitForTemplateUpdateRequests(namespace string, count i
...
 		return true, nil
 	})
-	return notification, err
+	return notifications, err
 }
 
-// WaitUntilNotificationDeleted waits until the Notification with the given name is deleted (ie, not found)
-func (a *HostAwaitility) WaitUntilNotificationDeleted(name string) error {
+// WaitUntilNotificationDeleted waits until the Notification for the given user is deleted (ie, not found)
+func (a *HostAwaitility) WaitUntilNotificationDeleted(username, notificationType string) error {
```
xcoulon, line 635: you could rename this func to `WaitUntilNotificationsDeleted` (plural on `Notification)

xcoulon (changes requested):

alexeykazakov (commented):
All my comments were addressed. But there are valid comments from Xavier. I'm fine with merging it as soon as the Xavier's comments are addressed.

wait/host.go
tinakurian: @xcoulon Hmm, I'm not sure this would work? I think we would need to know whether the expected notifications is less than the number we actually got in the list - in this case, we aren't waiting for more notifications. I was thinking this would probably be an error in our expectation or the test failing for some reason. On the other hand, if we have less notifications than what we expect, we should be waiting to see if more appear I guess? What do you think?

xcoulon (approved):
Thanks for taking my comments into account, @tinakurian 🙌

wait/host.go
tinakurian: Fixed here: b32b8a788930c4cfc225b932d8c5396daed3b270 to wait until we get an exact match or a timeout will occur

wait/host.go
tinakurian: Fixed as part of: b32b8a788930c4cfc225b932d8c5396daed3b270

wait/host.go
tinakurian: Fixed here: 9836e96a3fe2039ac20a77112f55042428ea443f

wait/host.go
tinakurian: Fixed as part of: b32b8a788930c4cfc225b932d8c5396daed3b270

alexeykazakov (approved):

PR #227, opened 2021-01-28
CRT-915 add test for encoded User IDs

sbryzak:
Associated tests for https://github.com/codeready-toolchain/registration-service/pull/167


alexeykazakov (approved):
Nice!

MatousJobanek (approved):
Cool :+1: 
However, it looks like you need to either change the way we create the `MyClaims` type or fix the assertion so you expect the subclaim and not the identity ID.

PR #233, opened 2021-02-09
chore: drop QUAY_NAMESPACE prefix

MatousJobanek:
We don't need it anymore and it will make the names shorter and more user-friendly

### DIFF
--- a/make/test.mk
+++ b/make/test.mk
@@ -7,10 +7,10 @@
 QUAY_NAMESPACE ?= codeready-toolchain
 DATE_SUFFIX := $(shell date +'%d%H%M%S')
 RESOURCES_SUFFIX := ${DATE_SUFFIX}
-HOST_NS ?= ${QUAY_NAMESPACE}-host-${DATE_SUFFIX}
-MEMBER_NS ?= ${QUAY_NAMESPACE}-member-${DATE_SUFFIX}
+HOST_NS ?= host-${DATE_SUFFIX}
+MEMBER_NS ?= member-${DATE_SUFFIX}
 REGISTRATION_SERVICE_NS := $(HOST_NS)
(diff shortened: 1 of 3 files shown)


alexeykazakov (approved):

make/test.mk
```diff
@@ -7,10 +7,10 @@
 QUAY_NAMESPACE ?= codeready-toolchain
 DATE_SUFFIX := $(shell date +'%d%H%M%S')
 RESOURCES_SUFFIX := ${DATE_SUFFIX}
-HOST_NS ?= ${QUAY_NAMESPACE}-host-${DATE_SUFFIX}
-MEMBER_NS ?= ${QUAY_NAMESPACE}-member-${DATE_SUFFIX}
+HOST_NS ?= host-${DATE_SUFFIX}
```
xcoulon, line 10: I found this prefix to be useful, as it would emphasise on the fact that these namespaces "belonged" to me.

MatousJobanek: Really? TBH, I don't have a string opinion on this... 
We can keep it there or remove - it should not have any effect on the functionality... I just thought that it could be a bit shorter and cleaner :-)

xcoulon: it's no big deal. Just that I could easily filter these namespaces by their prefix. But we can remove it, it's fine

xcoulon: by filtering, I meant something like
```
oc get projects -o name | grep xcoulon | xargs oc delete
```

xcoulon (approved):

make/test.mk
MatousJobanek: Yeah, I know. I only use the make targets for cleaning the environment. 
Anyway, If you still use it, then let's keep it there. I don't have any objection :wink:

MatousJobanek:
Closing for now

PR #237, opened 2021-02-16
test for crtadmin

tinakurian:
Preventing the registration-service from creating signup resources for users who signup which contains the term crtadmin in their username.

Jira: https://issues.redhat.com/browse/CRT-913

Related PRs: codeready-toolchain/toolchain-common#151, https://github.com/codeready-toolchain/registration-service/pull/170, https://github.com/codeready-toolchain/host-operator/pull/373

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -28,6 +28,7 @@ require (
 
 replace (
 	github.com/Azure/go-autorest => github.com/Azure/go-autorest v13.3.2+incompatible // Required by OLM
+	github.com/codeready-toolchain/toolchain-common v0.0.0-20200930070243-e124e69e7a0d => github.com/tinakurian/toolchain-common v0.0.0-20210216124316-080275ce4c11
 	github.com/openshift/api => github.com/openshift/api v0.0.0-20200821140346-b94c46af3f2b // Using 'github.com/openshift/api@release-4.5'
 	k8s.io/client-go => k8s.io/client-go v0.18.3 // Required by prometheus-operator
 	k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20200410145947-61e04a5be9a6 // avoids case-insensitive import collision: "github.com/googleapis/gnostic/openapiv2" and "github.com/googleapis/gnostic/OpenAPIv2"
--- a/go.mod
+++ b/go.mod
@@ -28,7 +28,7 @@ require (
 
 replace (
 	github.com/Azure/go-autorest => github.com/Azure/go-autorest v13.3.2+incompatible // Required by OLM
-	github.com/codeready-toolchain/toolchain-common v0.0.0-20210209125948-7a9e8c9e4ee6 => github.com/tinakurian/toolchain-common v0.0.0-20210216124316-080275ce4c11
+	github.com/codeready-toolchain/toolchain-common v0.0.0-20210209125948-7a9e8c9e4ee6 => github.com/tinakurian/toolchain-common v0.0.0-20210217164217-8325ca100257
 	github.com/openshift/api => github.com/openshift/api v0.0.0-20200821140346-b94c46af3f2b // Using 'github.com/openshift/api@release-4.5'
 	k8s.io/client-go => k8s.io/client-go v0.18.3 // Required by prometheus-operator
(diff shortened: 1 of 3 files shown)


test/e2e/registration_service_test.go
```diff
@@ -262,6 +262,24 @@ func (s *registrationServiceTestSuite) TestSignupFails() {
...
+		emailClaim := authsupport.WithEmailClaim(emailValue)
+		usernameClaim := authsupport.WithPreferredUsernameClaim("test-crtadmin")
+		token, err := authsupport.GenerateSignedE2ETestToken(*identity, emailClaim, usernameClaim)
+		require.NoError(s.T(), err)
+
+		// Call signup endpoint with a valid token to initiate a signup process
+		response := invokeEndpoint(s.T(), "POST", s.route+"/api/v1/signup", token, "", http.StatusInternalServerError)
+		require.Equal(s.T(), "failed to create usersignup for test-crtadmin:cannot create usersignup for crtadmin", response["message"])
+		require.Equal(s.T(), "error creating UserSignup resource", response["details"])
+		require.Equal(s.T(), float64(500), response["code"])
```
MatousJobanek, line 281: could you please add also a check that the `UserSignup` CR wasn't created?

tinakurian: done here: 532500657c77f81efb3a67b9e64effdefabcfaab

go.mod
```diff
@@ -28,6 +28,7 @@ require (
 
 replace (
 	github.com/Azure/go-autorest => github.com/Azure/go-autorest v13.3.2+incompatible // Required by OLM
+	github.com/codeready-toolchain/toolchain-common v0.0.0-20210209125948-7a9e8c9e4ee6 => github.com/tinakurian/toolchain-common v0.0.0-20210217164217-8325ca100257
```
MatousJobanek, line 31: a reminder

test/e2e/registration_service_test.go
```diff
@@ -262,6 +262,29 @@ func (s *registrationServiceTestSuite) TestSignupFails() {
...
```
MatousJobanek, line 281: shouldn't it return `403` as `Forbidden`?

MatousJobanek (commented):
Could you please check the code that is returned from the reg-service - it's weird that it returns 500 - I would rather expect 403 in this case

test/e2e/registration_service_test.go
tinakurian: Yeah.. so in reg-service code, i do return 403 forbidden, however, the handler just does a catch all and returns internal server error at the end: https://github.com/codeready-toolchain/registration-service/blob/master/pkg/controller/signup.go#L39-L43.. @alexeykazakov should this be changed?

xcoulon: @tinakurian I would suggest that you also fix the return code to 403 in https://github.com/codeready-toolchain/registration-service/pull/170 for such a kind of username, indeed.

MatousJobanek: I think that the error code should be changed/fixed to reflect the actual state/reason for the failure.
But feel free to do it as separate task/PR. @tinakurian Please create a jira story for that - as part of it go thought all the cases where we return error to check if there is any other use-case to be fixed.

MatousJobanek (approved):

test/e2e/registration_service_test.go
tinakurian: @xcoulon thanks for the review! yeah.. it needs to be fixed here though: https://github.com/codeready-toolchain/registration-service/blob/master/pkg/controller/signup.go#L39-L43.. the reg-service PR i created does return 403. It just isn't reflected in signup.go because there is a catch all which returns 500

tinakurian: @MatousJobanek sounds good, i think i saw at least one other place where a 400s error is converted into a 500.. will do this in a separate PR and create a new Jira for it. also, thanks for the review!

tinakurian: Jira for addressing the http error code catch-all: https://issues.redhat.com/browse/CRT-966

PR #241, opened 2021-02-22
Fix 500 catch all for PostHandler

tinakurian:
Currently, the registration-service signup-service returns other errors such as 403. However, the PostHandler does a catch-all and returns the http code 500 for all errors.  This should be changed to propagate and return the error which the service actually returned.

Corresponding PR: https://github.com/codeready-toolchain/registration-service/pull/171
Jira: https://issues.redhat.com/browse/CRT-966


test/e2e/registration_service_test.go
```diff
@@ -306,7 +306,7 @@ func (s *registrationServiceTestSuite) TestSignupOK() {
 		s.assertGetSignupStatusPendingApproval(identity.Username, token)
 
 		// Attempt to create same usersignup by calling post signup with same token should return an error
-		mp := invokeEndpoint(s.T(), "POST", s.route+"/api/v1/signup", token, "", http.StatusInternalServerError)
+		mp := invokeEndpoint(s.T(), "POST", s.route+"/api/v1/signup", token, "", http.StatusBadRequest)
```
alexeykazakov, line 309: Should be 409, right?

tinakurian: Yeah, didn't get to fixing it as fast because i noticed another error, for some reason the `re-registering new user with pre-existing ban` test now expects a "BadRequest" error instead of "Forbidden"... i'm not too sure what caused that change when looking at my code change, but i'll investigate that. Here is the change for the conflict: 6d59900c16a2d2577d05e07ce7577dd0d04e0054

alexeykazakov (approved):

xcoulon (approved):

PR #244, opened 2021-02-25
Added test to ensure users may not sign up with phone number already in use

sbryzak:
Added e2e test for https://github.com/codeready-toolchain/registration-service/pull/168

Fixes https://issues.redhat.com/browse/CRT-883

This PR also includes an increase to the timeout of client requests to the host API, to allow a greater threshold when running in CRC.

Signed-off-by: Shane Bryzak <shane@localhost.localdomain>

### DIFF
--- a/test/e2e/registration_service_test.go
+++ b/test/e2e/registration_service_test.go
@@ -519,10 +519,15 @@ func (s *registrationServiceTestSuite) TestPhoneVerification() {
 	responseMap := invokeEndpoint(s.T(), "PUT", s.route+"/api/v1/signup/verification", otherToken,
 		`{ "country_code":"+61", "phone_number":"408999999" }`, http.StatusForbidden)
 
-	responseCode, err := strconv.Atoi(responseMap["code"].(string))
-	require.NoError(s.T(), err)
+	require.NotEmpty(s.T(), responseMap)
+	var ok bool
+	if val, ok := responseMap["code"]; ok {
+		responseCode, err := strconv.Atoi(val.(string))
+		require.NoError(s.T(), err)
+		require.Equal(s.T(), http.StatusForbidden, responseCode)
+	}
+	require.True(s.T(), ok)
 
-	require.Equal(s.T(), http.StatusForbidden, responseCode)
 	require.Equal(s.T(), "Forbidden", responseMap["status"])
 	require.Equal(s.T(), "phone number already in use", responseMap["message"])
 	require.Equal(s.T(), "cannot register using phone number: +61408999999", responseMap["details"])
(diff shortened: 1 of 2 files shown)


test/e2e/registration_service_test.go
```diff
@@ -495,6 +495,35 @@ func (s *registrationServiceTestSuite) TestPhoneVerification() {
...
+	// Wait for the UserSignup to be created
+	otherUserSignup, err := s.hostAwait.WaitForUserSignup(otherIdentity.ID.String(),
+		wait.UntilUserSignupHasConditions(VerificationRequired()...),
+		wait.UntilUserSignupHasStateLabel(v1alpha1.UserSignupStateLabelValueNotReady))
+	require.NoError(s.T(), err)
+	otherEmailAnnotation := userSignup.Annotations[v1alpha1.UserSignupUserEmailAnnotationKey]
+	assert.Equal(s.T(), otherEmailValue, otherEmailAnnotation)
+
+	// Initiate the verification process using the same phone number as previously
+	invokeEndpoint(s.T(), "PUT", s.route+"/api/v1/signup/verification", otherToken,
```
alexeykazakov, line 518: Let's also check the error message returned by this endpoint.

alexeykazakov (commented):
Looks good overall. But let's check the error message too.

test/e2e/registration_service_test.go
sbryzak: Done

alexeykazakov (commented):
The test looks good!
But let's also test the case when there is a deactivated signup with the same phone and it's ignored.

test/e2e/registration_service_test.go
```diff
@@ -495,6 +495,65 @@ func (s *registrationServiceTestSuite) TestPhoneVerification() {
...
+	// Confirm there is no verification code annotation value
+	require.Empty(s.T(), otherUserSignup.Annotations[v1alpha1.UserSignupVerificationCodeAnnotationKey])
+
+	// Retrieve the current UserSignup
+	err = s.hostAwait.Client.Get(context.TODO(), types.NamespacedName{Namespace: s.hostAwait.Namespace, Name: userSignup.Name}, userSignup)
+	require.NoError(s.T(), err)
+
+	// Now mark the original UserSignup as deactivated
+	userSignup.Spec.Deactivated = true
+	userSignup.Labels[v1alpha1.UserSignupStateLabelKey] = v1alpha1.UserSignupStateLabelValueDeactivated
```
alexeykazakov, line 541: No need to set this label. It will be set by the controller.

alexeykazakov (approved):
Looks good. Just one minor comment.

PR #247, opened 2021-03-01
Auto approval tests using multiple member clusters

rajivnathan:
Related issue: https://issues.redhat.com/browse/CRT-949

Tests the automatic approval logic with two clusters to ensure that when cluster user account limits are set then when one cluster is full, remaining users should be provisioned to the other cluster.

Also includes a fix for the `test-perf-local` make target that was recently introduced.


test/e2e/usersignup_test.go
```diff
@@ -109,6 +109,63 @@ func (s *userSignupIntegrationTest) TestAutomaticApproval() {
...
+		}
+		require.Len(s.T(), memberLimits, 2)
+
+		userAccounts := &v1alpha1.UserAccountList{}
+		err = s.memberAwait.Client.List(context.TODO(), userAccounts)
+		require.NoError(s.T(), err)
+
+		userAccounts = &v1alpha1.UserAccountList{}
+		err = s.memberAwait2.Client.List(context.TODO(), userAccounts)
+		require.NoError(s.T(), err)
```
MatousJobanek, line 133: what is this good for?

test/e2e/usersignup_test.go
```diff
@@ -109,6 +109,63 @@ func (s *userSignupIntegrationTest) TestAutomaticApproval() {
...
+		expectedCluster1 := s.memberAwait
+		expectedCluster2 := s.memberAwait2
+		if mur1.Spec.UserAccounts[0].TargetCluster == s.memberAwait2.ClusterName {
+			expectedCluster1 = s.memberAwait2
+			expectedCluster2 = s.memberAwait
+		}
+		t.Logf("TargetCluster1: %s", mur1.Spec.UserAccounts[0].TargetCluster)
+		t.Logf("TargetCluster2: %s", mur2.Spec.UserAccounts[0].TargetCluster)
+		require.Equal(s.T(), mur1.Spec.UserAccounts[0].TargetCluster, expectedCluster1.ClusterName)
+		require.Equal(s.T(), mur2.Spec.UserAccounts[0].TargetCluster, expectedCluster2.ClusterName)
```
MatousJobanek, line 153: wouldn't be easier just to verify that they are provisioned to different clusters?
```suggestion
		require.NotEqual(s.T(), mur1.Spec.UserAccounts[0].TargetCluster,mur2.Spec.UserAccounts[0].TargetCluster)
```

BTW, when using `Equal` - the first param is the expected one:
```
require.Equal(s.T(), expectedCluster1.ClusterName, mur1.Spec.UserAccounts[0].TargetCluster)
```

test/e2e/usersignup_test.go
```diff
@@ -109,6 +109,63 @@ func (s *userSignupIntegrationTest) TestAutomaticApproval() {
...
+
+		// verify each usersignup is provisioned only on the cluster it was assigned
+		VerifyResourcesProvisionedForSignup(s.T(), s.hostAwait, *userSignup1, "basic", expectedCluster1)
+		VerifyResourcesProvisionedForSignup(s.T(), s.hostAwait, *userSignup2, "basic", expectedCluster2)
```
MatousJobanek, line 157: This check is already part of the `createAndCheckUserSignup` function

MatousJobanek (commented):
Thanks for adding the test for this scenario. 
The test looks good, but I think that the code could be simplified

test/e2e/usersignup_test.go
rajivnathan: Absolutely nothing! I had used them for debugging when the tests weren't working and I was too tired when I was doing the cleanup so I missed it 😆  I'll remove it.

test/e2e/usersignup_test.go
rajivnathan: Good point, thanks!

test/e2e/usersignup_test.go
rajivnathan: That's true. There's a slight difference here, in that I was only passing in the cluster it was expected to be in. But, since we do the cluster check before this I guess that should be good enough.

rajivnathan: Done bbae633047f6053a890249136999257f3a55d6a0

test/e2e/usersignup_test.go
rajivnathan: Done bbae633047f6053a890249136999257f3a55d6a0

test/e2e/usersignup_test.go
```diff
@@ -109,6 +109,42 @@ func (s *userSignupIntegrationTest) TestAutomaticApproval() {
...
+			}
+		}
+		require.Len(s.T(), memberLimits, 2)
+
+		s.hostAwait.UpdateHostOperatorConfig(test.AutomaticApproval().Enabled().MaxUsersNumber(0, memberLimits...))
+
+		// when
+		_, mur1 := s.createAndCheckUserSignup(false, "multimember-1", "multi1@redhat.com", nil, ApprovedAutomatically()...)
+		// we need to sleep one second to create UserSignup with different creation time
+		time.Sleep(time.Second)
```
MatousJobanek, line 132: This is actually not needed in this use-case - I used this hack to know which UserSignup won't be approved/provisioned and will end-up in the waiting list. But since we do expect that both of them should be provisioned, we don't need to put this time gap between them:
```suggestion
```

MatousJobanek (approved):
Nice :+1:

test/e2e/usersignup_test.go
rajivnathan: Ah, didn't know that. Thanks for clarifying

PR #250, opened 2021-03-05
onboard operator setup

xcoulon:
- refactor/cleanup packages
- restore 'make build' goal
- WIP feat: CLI to provision users with custom idlers


demo.cast
```diff
@@ -0,0 +1,1559 @@
+{"version": 2, "width": 117, "height": 70, "timestamp": 1614956408, "env": {"SHELL": "/bin/zsh", "TERM": "xterm-256color"}}
```
alexeykazakov, line 1: I guess you didn't want to commit this file?

setup/main.go
```diff
@@ -0,0 +1,22 @@
+/*
+Copyright © 2021 NAME HERE <EMAIL ADDRESS>
```
alexeykazakov, line 2: Looks like a template rather than an actual licence. The repo already includes the licence and I don't think we have to add it here.

demo.cast
rajivnathan: I'm not sure, that was there in @xcoulon original commit.

xcoulon: Indeed, it was not intended. Although, one may consider it an historical piece of art!
Joke aside, I will remove it

setup/main.go
xcoulon: right, that was generated by the Cobra CLI and I will remove it

demo.cast
xcoulon: removed in b7a9f4f

setup/main.go
xcoulon: removed in b307d9e

setup/user/user.go
```diff
@@ -0,0 +1,109 @@
...
+				return true, nil
+			}
+		}
+		return false, nil
+	})
+	return memberCluster.Name, err
+}
+
+func (s Setup) WaitForNamespace(namespace string) error {
+	ns := &corev1.Namespace{}
```
xcoulon, line 84: seems like a duplicate of the `func waitForNamespace(c *kubernetes.Clientset, namespace string) error` func in `setup/user/resource.go`

setup/user/user.go
```diff
@@ -0,0 +1,109 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type Setup struct {
+	Config *rest.Config
+	Client client.Client
+}
+
+func (s Setup) UpdateUserIdlersTimeout(username string, timeout time.Duration) error {
+	for _, suffix := range []string{"code", "dev", "stage"} { // TODO: hard coded suffixes, we could probably get them from the tier instead
```
xcoulon, line 30: I thought we said we would create a new tier with a shorter idler timeout? or did we decide to postpone it?

setup/user/resources.go
```diff
@@ -0,0 +1,175 @@
...
+	yamlutil "k8s.io/apimachinery/pkg/util/yaml"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/discovery/cached/memory"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/restmapper"
+)
+
+type templateProcessor struct {
```
xcoulon, line 28: what about reusing the template processor in https://github.com/codeready-toolchain/toolchain-common/blob/master/pkg/template/processor.go instead?

setup/user/user.go
rajivnathan: yeah, that's still the plan. Just haven't go around to it yet.

MatousJobanek (approved):
Great job @xcoulon @rajivnathan

PR #251, opened 2021-03-10
Experimental test to see if we can set creation date

sbryzak:
Signed-off-by: sbryzak <sbryzak@gmail.com>

### DIFF
--- a/test/e2e/usersignup_test.go
+++ b/test/e2e/usersignup_test.go
@@ -3,6 +3,7 @@ package e2e
 import (
 	"context"
 	"fmt"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


test/e2e/usersignup_test.go
```diff
@@ -295,6 +296,22 @@ func (s *userSignupIntegrationTest) TestTransformUsername() {
 	}
 }
 
+func (s *userSignupIntegrationTest) TestUserSignupModifiedCreationDate() {
+	userSignup := NewUserSignup(s.T(), s.hostAwait, "tango", "tango@xray.com")
+	userSignup.Spec.Approved = true
+	past := time.Now().AddDate(-1, 0, 0)
+	userSignup.ObjectMeta.CreationTimestamp = v1.Time{Time: past}
+
+	err := s.hostAwait.FrameworkClient.Create(context.TODO(), userSignup, CleanupOptions(s.ctx))
```
alexeykazakov, line 305: I guess Create will generate a new timestamp. Try to update the existing resource.

sbryzak: My guess is the same :)  I'll wait for this test to fail first then I'll try the update existing resource method.

PR #253, opened 2021-03-11
Add setup instructions for onboarding

rajivnathan:
Setup instructions for onboarding operators to Dev Sandbox.


setup/README.adoc
```diff
@@ -0,0 +1,85 @@
...
+* Log in to the target OpenShift 4.2+ cluster with cluster admin privileges using `oc login`
+
+=== Dev Sandbox Setup
+
+🚧 This section is still a work in progress 🚧
+
+1. Clone this repository
+`git clone git@github.com:codeready-toolchain/toolchain-e2e.git`
+2. Run the following to install the Dev Sandbox operators
+`make deploy-e2e HOST_NS=toolchain-host-operator MEMBER_NS=toolchain-member-operator MEMBER_NS_2=toolchain-member2-operator`
```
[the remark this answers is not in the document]
rajivnathan: The default ones when using the `deploy-e2e` target use the naming pattern `${QUAY_NAMESPACE}-host-${DATE_SUFFIX}`. eg. `rsenthil-host-11085720` So it looks a bit random because of the DATE_SUFFIX, I figured it's better to have something more constant.

I changed the instructions to use `dev-deploy-e2e` target instead and then the default naming pattern is `${QUAY_NAMESPACE}-host-operator` eg. `rsenthil-host-operator` which is a little better.

setup/README.adoc
```diff
@@ -0,0 +1,85 @@
+= Dev Sandbox Setup
+
+This document describes how to setup a Dev Sandbox environment that is similar to production and can be used as part of an onboarding process for new operators.
+
+=== Prereqs
+
+1. Provision an OCP cluster with sufficient resources.
```
alexeykazakov, line 7:
```suggestion
1. Provision an OCP cluster on AWS with sufficient resources: 4xlarge master nodes and at least 3 2xlarge worker nodes.
```

MatousJobanek (commented):
There is already a README for the related to the perf test-suite: https://github.com/codeready-toolchain/toolchain-e2e/tree/master/test/perf so I guess that we could merge them. 
Apparently, the perf README is hidden, so could you please add a ref link into the main README?

setup/README.adoc
```diff
@@ -0,0 +1,85 @@
...
+With the cluster now under load, it's time to evaluate the environment.
+
+1. Use your operators as a user would and evaluate the performance.
+2. Monitor the cluster's performance using the Monitoring view in the OpenShift Console.
+
+=== Clean up
+
+==== Remove Only Users and Their Namespaces
+
+Run `oc get usersignup -n toolchain-host-operator -o jsonpath='{.items..metadata.name}' | xargs oc delete usersignup`
```
MatousJobanek, line 82: couldn't you use:
```suggestion
Run `oc get usersignup -n toolchain-host-operator -o name | xargs oc delete usersignup`
```

MatousJobanek: or just:
```suggestion
Run `oc delete usersignups --all --all-namespaces`
```

you could also add a makefile target for that - it should be pretty easy - there is already one that contains the necessary commands: https://github.com/codeready-toolchain/toolchain-e2e/blob/master/make/clean.mk#L13-L14

rajivnathan: Thanks, adding a target for that is a great suggestion! 🥇  Added `clean-users` target in 3ee5c179b731f5c786807e4e1a5a919d78d863ed

rajivnathan:
@MatousJobanek thanks for the review, regarding merging the perf README with this one, we are actually using separate code for the setup which will also be in this setup folder (see PR https://github.com/codeready-toolchain/toolchain-e2e/pull/250). I mainly created this doc in a separate PR because I thought we could put up some initial instructions that other teams can start looking at before the setup CLI is ready. But I agree there is some potential for consolidation with the perf tests that we should think about. Would like to bring @xcoulon into that discussion as well. But I was thinking for now it may be good to keep the setup instructions for onboarding new users here so that we can point them to 1 place where they can follow instructions a bit easier since the main README has quite a bit more stuff in it.

MatousJobanek:
Yeah, you are right. I looked at this PR at first and then at the other one that contains the code. Let's keep the instruction separated - it makes more sense. :+1:

setup/README.adoc
```diff
@@ -0,0 +1,85 @@
...
+=== Prereqs
+
+1. Provision an OCP cluster on AWS with sufficient resources: 4xlarge master nodes and at least 3 2xlarge worker nodes.
+
+For example, if using the openshift-install tool to install OCP on AWS you can use an install-config.yaml file that has the following configuration:
+
+----
+apiVersion: v1
+baseDomain: devcluster.openshift.com
+compute:
```
pmacik, line 14: Isn't this `compute:` redundant? There is another on at line 27.

rajivnathan: You're right, sorry about that. I'll fix it.

rajivnathan: Done 8f838bac9640657be1d14ae9151fa13b8da1bb69 Thanks, nice catch!

alexeykazakov (approved):

PR #256, opened 2021-03-16
Add tests for pre deactivation user notification

sbryzak:
Related PR: https://github.com/codeready-toolchain/host-operator/pull/393

Fixes https://issues.redhat.com/browse/CRT-994


test/e2e/user_management_test.go
```diff
@@ -203,6 +212,39 @@ func (s *userManagementTestSuite) TestUserDeactivation() {
...
+		tierDeactivationDuration := time.Duration(baseTier.Spec.DeactivationTimeoutDays + 1) * time.Hour * 24
+		murMember1, err = s.hostAwait.UpdateMasterUserRecordStatus(murMember1.Name, func(mur *v1alpha1.MasterUserRecord) {
+			mur.Status.ProvisionedTime = &metav1.Time{Time: time.Now().Add(-tierDeactivationDuration)}
+		})
+		require.NoError(s.T(), err)
+		s.T().Logf("masteruserrecord '%s' provisioned time adjusted to %s", murMember1.Name,
+			murMember1.Status.ProvisionedTime.String())
+
+		// The user should be set to deactivating, but not deactivated
+		_, err = s.hostAwait.WaitForUserSignup(userSignupMember1.Name, wait.UntilUserSignupHasConditions(Deactivating()...))
```
alexeykazakov, line 245: After the wait let's verify the usersignup resources: `VerifyResourcesProvisionedForSignup(...)` to make sure the user is still active.
Also, shouldn't we check that the Deactivating Notificaton resource is created? It will be in a failing state but we at least could verify it's created, right?

test/e2e/base_user_test.go
```diff
@@ -100,6 +101,7 @@ func newBannedUser(host *wait.HostAwaitility, email string) *v1alpha1.BannedUser
 func (s *baseUserIntegrationTest) deactivateAndCheckUser(userSignup *v1alpha1.UserSignup, mur *v1alpha1.MasterUserRecord) {
 	userSignup, err := s.hostAwait.UpdateUserSignupSpec(userSignup.Name, func(us *v1alpha1.UserSignup) {
 		us.Spec.Deactivated = true
+		states.SetDeactivating(us, true)
```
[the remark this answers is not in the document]
sbryzak: It's "technically" not required, as the UserSignup controller will still go ahead with the deactivation even if the `deactivating` state isn't set.  I'm setting it here for consistency though, as the general deactivation workflow as per the deactivation controller is 1) set deactivating true -> 2) wait for notification -> 3) wait for deactivation time to arrive -> 4) set deactivation true.  Strictly speaking though you're right, we don't need it and I can remove it from here if you think it's confusing.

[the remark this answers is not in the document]
alexeykazakov: We should be able to deactivate a UserSignup by just setting Spec.Deactivated = true. So, let's remove this line. It will prove that it works as expected.

test/e2e/user_management_test.go
sbryzak: So this is not so easy as it seems.  In the Notification controller, if running within the e2e tests environment the notification is always treated as sent successfully:

```
	// if the environment is set to e2e do not attempt sending via mailgun
	if r.config.GetEnvironment() != "e2e-tests" {
		// Send the notification via the configured delivery service
		err = r.deliveryService.Send(notCtx, notification)
		if err != nil {
			return reconcile.Result{}, r.wrapErrorWithStatusUpdate(reqLogger, notification,
				r.setStatusNotificationDeliveryError, err, "failed to send notification")
		}
		reqLogger.Info("Notification has been sent")
	} else {
		reqLogger.Info("Notification has been skipped")
	}

	return reconcile.Result{
		Requeue:      true,
		RequeueAfter: r.config.GetDurationBeforeNotificationDeletion(),
	}, r.updateStatus(reqLogger, notification, r.setStatusNotificationSent)
```


What this means, is that the Notification resource gets deleted fasted than we have the check that it was created:

```
{"level":"info","ts":1619741085.5813034,"logger":"controller_notification","msg":"Reconciling Notification","Request.Namespace":"sbryzak-host-30090225","Request.Name":"usertostartdeactivating-provisioned-sh9tx"}
{"level":"info","ts":1619741085.5813544,"logger":"controller_notification","msg":"the Notification is sent so we can deal with its deletion","Request.Namespace":"sbryzak-host-30090225","Request.Name":"usertostartdeactivating-provisioned-sh9tx"}
{"level":"info","ts":1619741085.581392,"logger":"controller_notification","msg":"the Notification has been sent for a longer time than the 'durationBeforeNotificationDeletion', so it's ready to be deleted","Request.Namespace":"sbryzak-host-30090225","Request.Name":"usertostartdeactivating-provisioned-sh9tx","durationBeforeNotificationDeletion":"5s"}
{"level":"info","ts":1619741085.594358,"logger":"controller_notification","msg":"Reconciling Notification","Request.Namespace":"sbryzak-host-30090225","Request.Name":"usertostartdeactivating-provisioned-sh9tx"}
```

I don't have an easy solution for checking whether the Notification resource was ever created, any ideas for how we might do this?

alexeykazakov: Ok. Then let's skip the notification check part.

sbryzak: Actually, one idea did occur to me in the meantime.  What do you think of handling it this way:

https://github.com/codeready-toolchain/host-operator/pull/393/files#diff-7a61df0b58a303589124157586cb34986301fcbb8e75a4d3fe0f5efb1543bc15R127

alexeykazakov: I'm fine with this if it works :)

sbryzak: Ah geez, turns out there's other tests that actually wait for the notification to be deleted.  Seems like there's no easy fix after all, I'll revert this change and remove the notification check.

test/e2e/user_management_test.go
```diff
@@ -41,11 +41,17 @@ func (s *userManagementTestSuite) SetupSuite() {
 }
 
 func (s *userManagementTestSuite) TearDownTest() {
-	s.ctx.Cleanup()
+	// TODO uncomment
+	//s.ctx.Cleanup()
```
alexeykazakov, line 45: Reminder

test/e2e/user_management_test.go
```diff
@@ -45,7 +45,12 @@ func (s *userManagementTestSuite) TearDownTest() {
 }
 
 func (s *userManagementTestSuite) TestUserDeactivation() {
-	s.hostAwait.UpdateHostOperatorConfig(test.AutomaticApproval().Enabled())
+	s.hostAwait.UpdateHostOperatorConfig(
+		test.AutomaticApproval().Enabled(),
+		test.Deactivation().DeactivatingNotificationDays(-1))
```
MatousJobanek, line 50: Sorry for my dumb question - why the negative number? wouldn't that mean that the deactivating notification will be send 1 day after deactivation?

test/e2e/user_management_test.go
```diff
@@ -156,26 +161,30 @@ func (s *userManagementTestSuite) TestUserDeactivation() {
...
 		excludedMurMember1, err = s.hostAwait.UpdateMasterUserRecordStatus(excludedMurMember1.Name, func(mur *v1alpha1.MasterUserRecord) {
 			mur.Status.ProvisionedTime = &metav1.Time{Time: time.Now().Add(-tierDeactivationDuration)}
 		})
 		require.NoError(s.T(), err)
-		s.T().Logf("masteruserrecord '%s' provisioned time adjusted", excludedMurMember1.Name)
+		s.T().Logf("masteruserrecord '%s' provisioned time adjusted to %s", excludedMurMember1.Name, excludedMurMember1.Status.ProvisionedTime.String())
+
+		// The non-excluded user should be set to deactivated
+		_, err = s.hostAwait.WaitForUserSignup(userSignupMember1.Name, wait.UntilUserSignupHasConditions(Deactivated()...))
+		require.NoError(s.T(), err)
```
MatousJobanek, line 187: Correct me if I'm wrong, but the code on the following lines  189-194 seems to be doing the very same thing

test/e2e/user_management_test.go
```diff
@@ -156,26 +161,30 @@ func (s *userManagementTestSuite) TestUserDeactivation() {
...
 		require.NoError(t, err)
 
 		// We cannot wait days for testing deactivation so for the purposes of the e2e tests we use a hack to change the provisioned time
 		// to a time far enough in the past to trigger auto deactivation. Subtracting the given period from the current time and setting this as the provisioned
 		// time should test the behaviour of the deactivation controller reconciliation.
-		tierDeactivationDuration := time.Duration(baseTier.Spec.DeactivationTimeoutDays) * time.Hour * 24
+		tierDeactivationDuration := time.Duration(baseTier.Spec.DeactivationTimeoutDays + 1) * time.Hour * 24
 		murMember1, err = s.hostAwait.UpdateMasterUserRecordStatus(murMember1.Name, func(mur *v1alpha1.MasterUserRecord) {
 			mur.Status.ProvisionedTime = &metav1.Time{Time: time.Now().Add(-tierDeactivationDuration)}
 		})
```
MatousJobanek, line 174: Maybe stupid question, but shouldn't we also set the `state.deactivating = true` & and add the "Deactivating" condition to the status part in the `UserSignup` resource?

test/e2e/user_management_test.go
MatousJobanek: Ok, now I probably understand why you set the negative number - to skip the "wait" phase between deactivating & deactivation. I believe that better would be to modify the status part in the UserSignup "manually" as we do with `ProvisionedTime` in MUR.

MatousJobanek:
@sbryzak I guess that the e2e tests are failing because the `deactivateAndCheckUser` function sets the `Deactivated = true` so the UserSignup controller will delete the MUR, however, in your code you expect the status to be set.

MatousJobanek:
ah, now I see that you fixed it in https://github.com/codeready-toolchain/toolchain-e2e/pull/256/commits/c6262aeb3e237ff7b811fd0cbbfedd197a29d760 but wouldn't it be better to let UserSignup go through the lifecycle or to just set the Status manually? You know - to follow the real scenario...

test/e2e/user_management_test.go
sbryzak: It seems you're right, I've removed this duplicate check.

test/e2e/user_management_test.go
sbryzak: Alexey is correct in his summary.  And we do check the deactivating state here already, it's included in the `Deactivated()` condition set:

```
		userSignupMember1, err = s.hostAwait.WaitForUserSignup(userSignupMember1.Name,
			wait.UntilUserSignupHasConditions(Deactivated()...),
			wait.UntilUserSignupHasStateLabel(v1alpha1.UserSignupStateLabelValueDeactivated))
```

test/e2e/user_management_test.go
sbryzak: I'm not sure we can update `LastTransitionTime` the same way we can update `ProvisionedTime`, without having to modify the way that the `addOrUpdateStatusCondition()` function works in toolchain-common.

test/e2e/base_user_test.go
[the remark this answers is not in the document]
sbryzak: Done

test/e2e/registration_service_test.go
```diff
@@ -341,6 +343,7 @@ func (s *registrationServiceTestSuite) TestSignupOK() {
 
 		// Deactivate the usersignup
 		userSignup, err = s.hostAwait.UpdateUserSignupSpec(userSignup.Name, func(us *v1alpha1.UserSignup) {
+			states.SetDeactivating(us, true)
```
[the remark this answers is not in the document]
sbryzak: Done.

sbryzak:
> ah, now I see that you fixed it in c6262ae but wouldn't it be better to let UserSignup go through the lifecycle or to just set the Status manually? You know - to follow the real scenario...

@MatousJobanek @alexeykazakov literally just asked me to remove the setting of the "deactivating" state from this test :)  I think this is a good one to keep as is, since we already have tests for the standard user lifecycle.  This one is good to have for the distinct reason that it simulates us manually deactivating a user account, which we may be required to do sometimes.

alexeykazakov (approved):

test/e2e/user_management_test.go
MatousJobanek: You can update anything you want - you just need to call `Client.Status().Update()` to update the sub-resource. That would of course mean that you couldn't use the `addOrUpdateStatusCondition()` function - you would need to modify it manually.

test/e2e/user_management_test.go
MatousJobanek: > As far as I can see this test verifies that the signup with auto-deactivation enabled is auto-deactivated if it reached the timeout.

Yes, but right now the timeout is 3 days after notification was sent, which is something we don't verify here.

> state.deactivating is an internal state. It's implementation detail in this context IMO. We should not rely on this state to test auto-deactivation.

Yes and no, as I said before - with this change we start counting the timeout from the time when we sent the notification and not based on the deactivation timeout set in the config.
Ideally, we should also verify the use-case when we already reach the timeout set in configuration, but we haven't sent the notification yet - in that case, it should not deactivate/remove the user, but it should wait another 3 days. That's actually the reason why we did so many changes in the controllers. Unfortunately, such a test is not here nor in host-operator repo as unit-test (it would be good to add it).

Anyway, my main point was that to make the test working we need to set the `DeactivatingNotificationDays(-1)` - to negative number. TBH, this is really weird, and it's not really clear why we are doing that - the fact how we count when to deactivate the user (and not from which time) is exactly the implementation detail we should avoid here.

MatousJobanek:
@sbryzak do we verify anywhere that the `state=deactivating` is set? I found only checks for the conditions but not for the `UserSignup.Spec.State`

test/e2e/user_management_test.go
MatousJobanek: @sbryzak @alexeykazakov thoughts?

alexeykazakov: We should have tests for it I agree. But they should be additional tests. IMO the test which verifies the expired account should not relay on manual setting the deactivating state. Setting the state to "deactivated" should be enough. And this is what this test does as far as I can see. And changing the default configuration in e2e tests is IMO fine. We do it all over the place. If it's not clear why we change the configuration then we need to comment it in the code.

For proper testing the "deactivating" state we should add separate tests which do care about that intermediate state. If testing it as e2e is tricky then we should make sure that it's fully covered by Unit tests at least (it should be covered by Unit tests regardless).

MatousJobanek: > IMO the test which verifies the expired account should not rely on manual setting the deactivating state

I would rather let the controller set the condition and here just update the transition time than - in the same way as we do with the ProvisionedTime.

> Setting the state to "deactivated" should be enough. And this is what this test does as far as I can see. 

Not really - these tests don't verify manual deactivation but automatic, see for example this one: https://github.com/sbryzak/toolchain-e2e/blob/12370e3c64e78e3b0249abe134bd353c51ad4dc4/test/e2e/user_management_test.go#L157-L217
By mistake, I added the comment to the test with the name: `"tests for tiers with automatic deactivation disabled"` I should have added it to the one that I referenced `"tests for tiers with automatic deactivation enabled"` to make it more clear. Anyway, in both cases, we should rely on the step of setting the `deactivating` state and not on setting `DeactivatingNotificationDays` to a negative value to skip this step.

> And changing the default configuration in e2e tests is IMO fine.  We do it all over the place.

yes, but we set meaningful values, not negative ones to skip some state in the UserSignup lifecycle. 

> For proper testing the "deactivating" state we should add separate tests which do care about that intermediate state. 

This is exactly the place where it should be. The whole flow is not verified anywhere - it should be verified in these tests where we deal with the automatic approval.

alexeykazakov: Initially @sbryzak added the deacitivating state in all tests which needed a deactivated account. And I asked not to do that.
So, sorry I didn't get the context of this particular test into account when replying.
I agree that in this particular test where we are testing the automated deactivation (not manual) it's better to test the entire flow. But I would test it differently:
- Provision an account.
- Update the provisioning timestamp to something old enough
- Wait for the deactivating state/status
- Update the deactivating notification status timestamp to something old enough
- Wait for the account to be deactivated

So, we do not modify any default configuration, don't skip any steps of the flow and don't set any states manually. We would test exactly what we want: automated flow.

MatousJobanek: This flow is exactly what I had in my mind - sorry for not being clear :+1: 
@sbryzak could you please take care of it?

PR #257, opened 2021-03-16
refactor: use OpenShift template when provisioning users

xcoulon:
also, use a flag to specify the path to the YAML template


MatousJobanek (approved):

setup/resources/user-workloads.yaml
```diff
@@ -0,0 +1,157 @@
+kind: Template
```
MatousJobanek, line 1: The README contains a reference to the current sample resource: https://github.com/codeready-toolchain/toolchain-e2e/tree/master/setup/sample so you should probably replace it or change the README link.
You could also add information that the file should contain the OS template

setup/resources/create_resources.go
```diff
@@ -0,0 +1,57 @@
...
+	applycl := applycl.NewApplyClient(cl, s)
+	for _, obj := range objs {
+		if _, err := applycl.ApplyObject(obj.GetRuntimeObject()); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func getTemplateFromFile(s *runtime.Scheme, filename string) (*templatev1.Template, error) {
```
rajivnathan, line 48: I passed in the old user-workloads.yaml file https://github.com/codeready-toolchain/toolchain-e2e/blob/master/setup/sample/user-workloads.yaml and surprisingly it did not fail with an error. No resources were created in the workspace which is expected since there is no OpenShift template but I believe it should result in an error. I suppose it's failing somewhere in this function

setup/resources/user-workloads.yaml
```diff
@@ -0,0 +1,157 @@
...
+apiVersion: v1
+metadata:
+  name: nodejs-sample
+objects:
+  - apiVersion: apps/v1
+    kind: Deployment
+    metadata:
+      namespace: ${NAMESPACE}
```
rajivnathan, line 9: What happens if someone adds a new resource but forgets to set `namespace: ${NAMESPACE}`? Will the resource be created in an unexpected namespace?

setup/resources/create_resources.go
xcoulon: ok let me check that

setup/resources/user-workloads.yaml
xcoulon: I need to check that. I wanted to enforce the namespace on the template objects between the `Process` and `Apply` funcs, but I couldn't find an easy way to do it. But I guess I should give it another try, so we don't have to deal with this parameter at all...

setup/resources/create_resources.go
xcoulon: fixed in 86b510c: an error is returned if the decoding of the content YAML file did not return an OpenShift template

setup/resources/user-workloads.yaml
xcoulon: done in 00c8f2e and 86b510c

setup/resources/user-workloads.yaml
xcoulon: done in 00c8f2e: there is no template parameter anymore, yet the objects' namespace is enforced just before creation.

setup/README.adoc
```diff
@@ -44,17 +44,19 @@ publish: External
 [start=2]
 . Install your operator(s)
 
-. Create a user-workloads.yaml file. This file should contain resources that the setup tool will automatically create in user namespaces. Use the sample https://raw.githubusercontent.com/codeready-toolchain/toolchain-e2e/master/setup/sample/user-workloads.yaml[user-workloads.yaml] as a base and add custom resources that your operator uses.
+. Create an OpenShift template file which with all the resources that this setup tool will automatically create on behalf of the users in their `-stage` namespaces. You can use the sample https://raw.githubusercontent.com/codeready-toolchain/toolchain-e2e/master/setup/resources/user-workloads.yaml[user-workloads.yaml] as a base and add other resources that the users may want to create when using your operator.
```
rajivnathan, line 47:
```suggestion
. Create an OpenShift template file which has all the resources that this setup tool will automatically create on behalf of the users in their `-stage` namespaces. You can use the sample https://raw.githubusercontent.com/codeready-toolchain/toolchain-e2e/master/setup/resources/user-workloads.yaml[user-workloads.yaml] as a base and add other resources that the users may want to create when using your operator. 
```

rajivnathan (approved):
Nice, looks good! Thanks for adding tests too. 🥇  Made a minor wording suggestion, but not really necessary.

PR #263, opened 2021-03-22
Create more rolebindings and CSVs during setup

rajivnathan:
- Adds a prepare step that installs one of the available operators that works at the "All Namespaces" scope. I chose Kiali since it seems pretty lightweight and I couldn't get the OSD route-monitor operator working but we could use any operator that fills this need. Maybe once we onboard SBO we can use that since with the improvements it has a small footprint and will be part of the Sandbox experience which is better than using another operator which can be an external influence on the test results
- Adds 12 RoleBindings to bring the resource numbers w/3000 users in line with production
- Removes an unnecessary secret

### DIFF
--- a/setup/operators/install_operators.go
+++ b/setup/operators/install_operators.go
@@ -1,0 +1,47 @@
+package operators
+
+import (
+	"context"
+
+	"github.com/codeready-toolchain/toolchain-e2e/setup/resources"
+	"github.com/operator-framework/api/pkg/operators/v1alpha1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const AllNamespacesOperatorName = "kiali-ossm"
+const AllNamespacesOperatorPrefix = "kiali-operator"
+
+func EnsureAllNamespacesOperator(cl client.Client, csvNamespace string) error {
+	hasCSV, err := resources.HasCSVWithPrefix(cl, AllNamespacesOperatorPrefix, csvNamespace)
+	if err != nil && !k8serrors.IsNotFound(err) {
+		return err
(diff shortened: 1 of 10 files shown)


setup/operators/install_operators.go
```diff
@@ -0,0 +1,50 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	AllNamespacesOperatorName   = "kiali-ossm"
+	AllNamespacesOperatorPrefix = "kiali-operator"
+)
+
+func EnsureAllNamespacesOperator(cl client.Client, csvNamespace string) error {
```
alexeykazakov, line 19: Let's add a comment explaining why we install operators here.

alexeykazakov (approved):
I wonder if really care if the router operator works or not. I guess just installing it would do the trick of creating the needed amount of CSVs?

But anyway I'm fine with it sing kiali for now. But let's add some comments in the code to explain why we are doing it.

MatousJobanek (approved):
Cool :+1: 
BTW, how about adding a note to the README to make it transparent? https://github.com/codeready-toolchain/toolchain-e2e/tree/master/setup#dev-sandbox-setup-1 it's just a proposal - I don't insist on it

setup/operators/install_operators.go
rajivnathan: Sure, I had a comment in root.go but it's worth having one here too. Done 1bbdd9c38f46e6cad22e261315e7f2b837b72b90

rajivnathan:
@alexeykazakov Sorry, should have been more clear. I meant I couldn't get the route operator to create the CSVs so that's why I went with kiali.

rajivnathan:
@MatousJobanek Good point about making it more transparent. Added a note about it: ef8f694806203ddf2f4ab934911a4da3d0ac1a14

PR #265, opened 2021-03-23
updated e2e to check Member has apiEndpoint

ranakan19:
wrt fix: https://issues.redhat.com/browse/CRT-971
changes in host-operator in : https://github.com/codeready-toolchain/host-operator/pull/394

PR #267, opened 2021-03-25
Adding e2e tests for adding apiEndpoint

ranakan19:
Creating a new PR for https://github.com/codeready-toolchain/toolchain-e2e/pull/265, such that the branch name (apiEndpoint) is the same as the one on host-operator
wrt fix: https://issues.redhat.com/browse/CRT-971
changes in host-operator in : codeready-toolchain/host-operator#394


testsupport/wait/host.go
```diff
@@ -679,6 +679,18 @@ func UntilAllMembersHaveUsageSet() ToolchainStatusWaitCriterion {
...
+	return func(a *HostAwaitility, toolchainStatus *toolchainv1alpha1.ToolchainStatus) bool {
+		for _, member := range toolchainStatus.Status.Members {
+			// check Member field ApiEndpoint is assigned
+			if member.ApiEndpoint == "" {
+				return false
+			}
+		}
+		return true
+	}
+}
```
alexeykazakov, line 692: I approved this PR before but I think we can actually improve this test by verifying the actual api endpoint value instead of just checking it's not an empty string:

```suggestion
func UntilAllMembersHaveApiEndpoint(apiEndpoint string) ToolchainStatusWaitCriterion {
	return func(a *HostAwaitility, toolchainStatus *toolchainv1alpha1.ToolchainStatus) bool {
		for _, member := range toolchainStatus.Status.Members {
			// check Member field ApiEndpoint is assigned
			if member.ApiEndpoint == apiEndpoint {
				return false
			}
		}
		return true
	}
}
```

And in VerifyToolchainStatus():

```
	...
	memberCluster, found, err := a.GetToolchainCluster(cluster.Member, a.Namespace, nil)
	require.NoError(t, err)
	require.True(t, found)
	_, err := hostAwait.WaitForToolchainStatus(wait.UntilToolchainStatusHasConditions(ToolchainStatusReadyAndUnreadyNotificationNotCreated()...),
		wait.UntilAllMembersHaveUsageSet(),
		wait.UntilAllMembersHaveApiEndpoint(memberCluster.Spec.APIEndpoint))
	...
```

MatousJobanek: Since both member operators run in the same cluster, then using the same `memberCluster.Spec.APIEndpoint` should be fine. But it would be fine to add a comment why do we expect all members to have the same API endpoint.
In addition to that, you need to change also the comparison in the if statement:
```suggestion
func UntilAllMembersHaveApiEndpoint(apiEndpoint string) ToolchainStatusWaitCriterion {
	return func(a *HostAwaitility, toolchainStatus *toolchainv1alpha1.ToolchainStatus) bool {
		for _, member := range toolchainStatus.Status.Members {
			// check Member field ApiEndpoint is assigned
			if member.ApiEndpoint != apiEndpoint {
				return false
			}
		}
		return true
	}
}
```

ranakan19: @alexeykazakov but there can be multiple member clusters right? So all members wouldn’t have the same apiEndpoint.

alexeykazakov: @ranakan19 we do have multiple members but as Matous mentioned above they are physically the same cluster in our e2e tests. So api endpoints are the same. But I agree with Matous that we need to add a comment explaining that.

ranakan19: would this always be the case though? i didn't see matous's comment before, refreshed it twice to see it now :)

alexeykazakov: For the feasible future yes. It's pretty tricky to create multiple clusters in openshift-ci (if even possible). So we use different namespaces where we install master operators to imitate multiple clusters. But since they all run in the same physical cluster the api endpoint is the same.

ranakan19: done

testsupport/wait/host.go
```diff
@@ -679,6 +679,19 @@ func UntilAllMembersHaveUsageSet() ToolchainStatusWaitCriterion {
 	}
 }
 
+func UntilAllMembersHaveApiEndpoint(apiEndpoint string) ToolchainStatusWaitCriterion {
+	return func(a *HostAwaitility, toolchainStatus *toolchainv1alpha1.ToolchainStatus) bool {
+		//Since both member operators run in the same cluster, then using the same memberCluster.Spec.APIEndpoint should be fine.
```
alexeykazakov, line 684:
```suggestion
		//Since all member operators currently run in the same cluster in the e2e test environment, then using the same memberCluster.Spec.APIEndpoint for all the member clusters should be fine.
```

alexeykazakov (approved):
Looks good! Just one minor suggestion to improve the comment.

PR #271, opened 2021-03-29
feat: add posibility to deploy dev env without second member

MatousJobanek:
Because we don't need to deploy two member operators in one cluster when either setting up the cluster for performance tests or just for dev purposes, I've added make targets to be able to define how many members should be deployed.


alexeykazakov (approved):
Nice improvement!

PR #273, opened 2021-04-01
Split setup template

rajivnathan:
Modify the setup tool to accept multiple template files as params and update the README instructions to ask onboarding teams to create a separate template file for their own operators that will be used alongside the user-workloads.yaml template provided in this repo.


setup/README.adoc
```diff
@@ -71,12 +74,12 @@ Note that the template does not take any parameter, and all resources will be cr
 NAME               MURS   READY   LAST UPDATED
 toolchain-status   0      True    2021-03-24T22:39:36Z
 ```
-. Populate the cluster with users (by default the tool will provision 3000 users) and create resources based on the user-workloads.yaml file you created in the prereqs step above. +
-`go run setup/main.go --template=<path_to_yaml_template_file>`
+. Populate the cluster with users (by default the tool will provision 3000 users) and create resources by providing template files using the `--template` parameter. Typically, the provided user-workloads.yaml file is used along with the onboarding template you created in the prereq steps above.+
+`go run setup/main.go --template=setup/resources/user-workloads.yaml --template=<path_to_onboarding_template_from_prereq_step>`
```
alexeykazakov, line 78: Shouldn't we just always add `setup/resources/user-workloads.yaml`?
So, users would need to specify only additional templates and `setup/resources/user-workloads.yaml` is always used regardless.

WDYT? Do you have any case in mind when we would need not to use `setup/resources/user-workloads.yaml`? If yes then maybe we want to add another optional param to disable `setup/resources/user-workloads.yaml` usage?

rajivnathan: Good idea, done in https://github.com/codeready-toolchain/toolchain-e2e/pull/273/commits/6ce315edf88e9467bc3aa0eba98ce3195f8089e5 we will add it automatically and list the templates that will be used before they confirm they want to run the setup tool

alexeykazakov (approved):

alexeykazakov (approved):

MatousJobanek (approved):

PR #277, opened 2021-04-07
Update tests to use base tier as default

rajivnathan:
Related PR: https://github.com/codeready-toolchain/host-operator/pull/402

The above PR changes auto approval to use the base tier by default. This PR updates the e2e tests to reflect that as well and use the base tier instead of the basic tier in most tests since the basic tier will likely be removed in the future.


alexeykazakov (approved):
Nice!

MatousJobanek (approved):

PR #281, opened 2021-04-19
Add remove self-provisioner step

rajivnathan:
Add a step to recommend removing the self provisioner role.


alexeykazakov (approved):
Thanks!

PR #283, opened 2021-04-22
Onboarding setup tool improvements

rajivnathan:
Related Issue: https://issues.redhat.com/browse/CRT-1030

Summary of changes:
- each resource is created in a separate go routine with its own apply client
- move idler updates to its own job separate from resource creation and report progress separately
- added some timing information that is printed at the end of the setup
  - avg idler update time
  - avg time per user setup

Note: With these changes the setup is approx. 10% faster than without these changes but still not as fast as without OpenShift templates because the apply client does 2 network requests for each call to `ApplyObject` so people running the setup tool with increased network latency are still likely to experience longer run times for the tool.

The next step I plan to look into, is updating the tool to run the setup on the cluster itself instead of from the local machine to eliminate network latency issues.


alexeykazakov (approved):
Didn't review it very carefully but looks good overall! :)

PR #291, opened 2021-05-05
test: verify email address in MUR annotations

xcoulon:
E2E tests for https://github.com/codeready-toolchain/host-operator/pull/419


testsupport/user_setup.go
```diff
@@ -96,6 +96,9 @@ func CreateAndApproveSignup(t *testing.T, hostAwait *wait.HostAwaitility, userna
 	// First, wait for the MasterUserRecord to exist, no matter what status
 	mur, err := hostAwait.WaitForMasterUserRecord(userSignup.Status.CompliantUsername, wait.UntilMasterUserRecordHasConditions(Provisioned(), ProvisionedNotificationCRCreated()))
 	require.NoError(t, err)
+	// check that there's an annotation with the user's email address
+	assert.NotEmpty(t, userSignup.Annotations[toolchainv1alpha1.UserSignupUserEmailAnnotationKey])
+	assert.Equal(t, userSignup.Annotations[toolchainv1alpha1.UserSignupUserEmailAnnotationKey], mur.Annotations[toolchainv1alpha1.MasterUserRecordEmailAnnotationKey])
```
alexeykazakov, line 101: You don't really check the actual email here. I don't remember if we auto-generate the email in the user token in our tests but would be great to actually test the email value or at least that it's not empty.

And it would probably useful to have the annotation check (for both UserSignup and MUR) here too: https://github.com/codeready-toolchain/toolchain-e2e/blob/a8046eb7c461d7c1f44263b655830698016f7c86/testsupport/user_assertions.go#L22
At least that the UserAccount and MUR annotations match.

MatousJobanek: I think that this is already done on line 100:
```
assert.NotEmpty(t, userSignup.Annotations[toolchainv1alpha1.UserSignupUserEmailAnnotationKey])
```

MatousJobanek: but I agree that we could add a check here
https://github.com/codeready-toolchain/toolchain-e2e/blob/936c87f041fdb1aea2222f7ae35879aee8c40b06/testsupport/user_setup.go#L81-L83
to really verify the actual content of the email and not only that it's not empty.

xcoulon: done in e3d79e4 (hopefully)

MatousJobanek (approved):
Cool :+1:

PR #293, opened 2021-05-07
Migrate VerificationRequired property to states

sbryzak:
Tests for https://github.com/codeready-toolchain/host-operator/pull/422


test/e2e/usersignup_test.go
```diff
@@ -321,6 +322,9 @@ func (s *userSignupIntegrationTest) createUserSignupVerificationRequiredAndAsser
 	userSignup.Spec.Approved = true
 
 	// Set verification required
+	states.SetVerificationRequired(userSignup, true)
```
alexeykazakov, line 325: I guess we don't really need this e2e PR at all. If we run the corresponding host PR with master e2e then it would prove that the migration step (in host) works.

When we deploy the host and then drop the old API then we will need to update the e2e tests too.
WDYT?

PR #298, opened 2021-05-12
Removed Approved property from UserSignup, replaced with states

sbryzak:
Fixes https://issues.redhat.com/browse/CRT-1004

Related host-operator PR: https://github.com/codeready-toolchain/host-operator/pull/432/files


go.mod
```diff
@@ -44,6 +44,7 @@ require (
 )
 
 replace (
+	github.com/codeready-toolchain/api => github.com/sbryzak/api v0.0.0-20210511043252-f5014d872796
```
alexeykazakov, line 47: Just a reminder.

test/e2e/registration_service_test.go
```diff
@@ -486,7 +486,7 @@ func (s *registrationServiceTestSuite) TestPhoneVerification() {
 	require.False(s.T(), mpStatus["verificationRequired"].(bool))
 
 	// Now approve the usersignup.
-	userSignup.Spec.Approved = true
+	states.SetApproved(userSignup, true)
```
alexeykazakov, line 489: Maybe worth leaving at least one test which uses the old API to verify the migration code works?

go.mod
MatousJobanek: Shouldn't be this done later? I thought that this PR should verify the migration and also that UserSignup controller correctly uses the approved state (it's at least paired with that one). Removal should be done later in another PR when everything is migrated.
In addition, before removing it we should also update the registration service as well as sandbox-sre, shouldn't we?

test/e2e/registration_service_test.go
sbryzak: I think the unit tests are sufficient for verifying the migration, don't you?

go.mod
sbryzak: I've got the registration service PR ready to go, but I won't pull the trigger on that until this one is merge otherwise the e2e tests CI job will complain about too many branches with the same name.  Will discuss with @alexeykazakov tomorrow the best way to proceed with this set of changes for minimal disruption.

test/e2e/registration_service_test.go
alexeykazakov: Covering it by e2e would be even better if it's easy to do. So, if we simply leaving one test which uses the old API as is, then we might get a migration thing covered by our e2e very cheap, right?

test/e2e/base_user_test.go
```diff
@@ -139,6 +140,7 @@ func (s *baseUserIntegrationTest) reactivateAndCheckUser(userSignup *v1alpha1.Us
 	userSignup, err = s.hostAwait.UpdateUserSignupSpec(userSignup.Name, func(us *v1alpha1.UserSignup) {
 		states.SetDeactivating(us, false)
 		states.SetDeactivated(us, false)
+		states.SetApproved(us, true)
```
alexeykazakov, line 143: Why we need to explicitly set the states to Approved here but we didn't have to do it with the old API?

sbryzak: We require this now because the `SetDeactivated()` function sets the `approved` state to false.

sbryzak: Technically, we could just call `SetApproved()` and we get both `deactivating` and `deactivated` set to false for us for free.

alexeykazakov: But `states.SetDeactivated(us, false)` does not touch "approved", does it? https://github.com/codeready-toolchain/toolchain-common/blob/master/pkg/states/state_manager.go#L52
It deletes it in case of `states.SetDeactivated(us, true)`. But not for `states.SetDeactivated(us, false)`.

sbryzak: You're correct, but it's [this line](https://github.com/codeready-toolchain/toolchain-e2e/blob/master/test/e2e/base_user_test.go#L103) that's calling it, not the prior line.

alexeykazakov: I see... So, for manual re-activation/approval we need to restore this state. This should be fine.

go.mod
alexeykazakov: Yeah I guess it's OK to update the api dependency in e2e here if we don't need it in the code anymore. But what about keeping it for testing the migration: https://github.com/codeready-toolchain/toolchain-e2e/pull/298#discussion_r630726134 ?

test/e2e/registration_service_test.go
sbryzak: No problem, I've reverted to the original code here in order to test the migration.

go.mod
sbryzak: We are now testing the migration, so the api dependency cannot be updated as of yet.

alexeykazakov (approved):

test/e2e/registration_service_test.go
```diff
@@ -487,6 +487,8 @@ func (s *registrationServiceTestSuite) TestPhoneVerification() {
 
 	// Now approve the usersignup.
 	userSignup.Spec.Approved = true
+	//states.SetApproved(userSignup, true)
```
sbryzak, line 490: @alexeykazakov so it seems that this is possibly not a good place to test the migration.  From what I can gather from the currently failing e2e tests, a phone verification test is failing as the migration is not happening fast enough.  Perhaps it might be better to test the migration somewhere in usersignup_test instead?

alexeykazakov: I doubt the problem is with the speed of the migration. We use `WaitForMasterUserRecord()` to wait for the MUT get provisioned after updating the Approved in the UserSignup spec. The wait timeout should be more than enough.

The test looks correct. So, let's investigate what's happening to make sure there is no bugs in the runtime code.

test/e2e/base_user_test.go
MatousJobanek: @sbryzak 
I think that we should not call the `states.SetApproved(us, true)`  here, otherwise, you won't be able to verify reactivation with automatic approval. By setting `states=approved` you approve the UserSignup manually, which is not the purpose of this function!

> We require this now because the SetDeactivated() function sets the approved state to false.

I think that it only deletes the `approved` from the states: https://github.com/codeready-toolchain/toolchain-common/blob/master/pkg/states/state_manager.go#L49-L56 in addition, the `SetDeactivated` is called with `false` which means that it won't touch the `approved` state at all.

If I'm not mistaken, when we reactivate (without requiring phone verification), then the `states` should be empty.

test/e2e/registration_service_test.go
```diff
@@ -537,16 +548,32 @@ func (s *registrationServiceTestSuite) TestPhoneVerification() {
...
-	err = s.hostAwait.Client.Get(context.TODO(), types.NamespacedName{Namespace: s.hostAwait.Namespace, Name: userSignup.Name}, userSignup)
+	userSignup, err = s.hostAwait.WaitForUserSignup(userSignup.Name,
+		wait.UntilUserSignupMigrated())
 	require.NoError(s.T(), err)
 
 	// Now mark the original UserSignup as deactivated
 	states.SetDeactivated(userSignup, true)
 
 	// Update the UserSignup
+	log.Infof("### Updating UserSignup to set it deactivated: [%s]", userSignup.Name)
```
alexeykazakov, line 559:
```suggestion
```

test/e2e/registration_service_test.go
```diff
@@ -537,16 +548,32 @@ func (s *registrationServiceTestSuite) TestPhoneVerification() {
...
+
 	err = s.hostAwait.Client.Update(context.TODO(), userSignup)
 	require.NoError(s.T(), err)
 
+	// TODO diagnostic code, remove this
+	userSignup, err = s.hostAwait.WaitForUserSignup(userSignup.Name)
+	require.NoError(s.T(), err)
+	b, err := json.Marshal(userSignup)
+	require.NoError(s.T(), err)
+	log.Infof("### Reloaded UserSignup after setting deactivated: [%s]", string(b))
```
alexeykazakov, line 569:
```suggestion
```

alexeykazakov (approved):
Looks good but please remove the debugging logs.

test/e2e/registration_service_test.go
sbryzak: Removed

test/e2e/registration_service_test.go
sbryzak: Removed

PR #299, opened 2021-05-14
fix: wait for expected number of resources

MatousJobanek:
fix for flaky tests

### DIFF
--- a/testsupport/tiers/checks.go
+++ b/testsupport/tiers/checks.go
@@ -805,51 +805,70 @@ func count(resource v1.ResourceName) v1.ResourceName {
 
 func numberOfToolchainRoles(number int) namespaceObjectsCheck { // nolint: unparam
 	return func(t *testing.T, ns *v1.Namespace, memberAwait *wait.MemberAwaitility, userName string) {
-		roles := &rbacv1.RoleList{}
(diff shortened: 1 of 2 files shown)


testsupport/tiers/checks.go
```diff
@@ -805,51 +805,70 @@ func count(resource v1.ResourceName) v1.ResourceName {
 
 func numberOfToolchainRoles(number int) namespaceObjectsCheck { // nolint: unparam
 	return func(t *testing.T, ns *v1.Namespace, memberAwait *wait.MemberAwaitility, userName string) {
-		roles := &rbacv1.RoleList{}
-		err := memberAwait.Client.List(context.TODO(), roles, providerMatchingLabels, client.InNamespace(ns.Name))
+		err := memberAwait.WaitForExpectedNumberOfResources("Role", number, func() (int, error) {
+			roles := &rbacv1.RoleList{}
+			err := memberAwait.Client.List(context.TODO(), roles, providerMatchingLabels, client.InNamespace(ns.Name))
+			require.NoError(t, err)
+			assert.Len(t, roles.Items, number)
```
alexeykazakov, line 812: We don't need this assert (or any other asserts) in the list function, do we?

MatousJobanek: no, we don't - it's forgotten zombie from the original code :ghost: 
removed in 2d25d9b 
Thanks :pray:

xcoulon (approved):

PR #319, opened 2021-06-11
chore: verify reactivated users are not deleted

MatousJobanek:
paired with: https://github.com/codeready-toolchain/host-operator/pull/459

### DIFF
--- a/test/e2e/base_user_test.go
+++ b/test/e2e/base_user_test.go
@@ -130,7 +130,29 @@ func (s *baseUserIntegrationTest) deactivateAndCheckUser(userSignup *toolchainv1
 	require.True(s.T(), states.Deactivated(userSignup), "usersignup should be deactivated")
 }
 
-func (s *baseUserIntegrationTest) reactivateAndCheckUser(userSignup *toolchainv1alpha1.UserSignup, mur *toolchainv1alpha1.MasterUserRecord) {
+func (s *baseUserIntegrationTest) reactivateAndRequirePhoneVerification(userSignup *toolchainv1alpha1.UserSignup, mur *toolchainv1alpha1.MasterUserRecord) {
+	err := s.hostAwait.Client.Get(context.TODO(), types.NamespacedName{
+		Namespace: userSignup.Namespace,
+		Name:      userSignup.Name,
+	}, userSignup)
+	require.NoError(s.T(), err)
+
+	userSignup, err = s.hostAwait.UpdateUserSignupSpec(userSignup.Name, func(us *toolchainv1alpha1.UserSignup) {
+		states.SetDeactivating(us, false)
+		states.SetDeactivated(us, false)
+		states.SetVerificationRequired(us, true)
+	})
+	require.NoError(s.T(), err)
+	s.T().Logf("user signup '%s' reactivated, but requiring phone verification", userSignup.Name)
+
+	userSignup, err = s.hostAwait.WaitForUserSignup(userSignup.Name,
+		wait.UntilUserSignupHasConditions(ReactivatedAndVerificationRequired()...),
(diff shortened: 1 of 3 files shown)


test/e2e/user_management_test.go
```diff
@@ -138,14 +143,7 @@ func (s *userManagementTestSuite) TestUserDeactivation() {
...
 		// time should test the behaviour of the deactivation controller reconciliation.
-		manyManyDaysAgo := 999999999999999
-		durationDelta := time.Duration(manyManyDaysAgo) * time.Hour * 24
-		updatedProvisionedTime := &metav1.Time{Time: time.Now().Add(-durationDelta)}
-		murMember1, err = s.hostAwait.UpdateMasterUserRecordStatus(murMember1.Name, func(mur *toolchainv1alpha1.MasterUserRecord) {
-			mur.Status.ProvisionedTime = updatedProvisionedTime
-		})
-		require.NoError(s.T(), err)
-		s.T().Logf("masteruserrecord '%s' provisioned time adjusted", murMember1.Name)
+		murMember1, updatedProvisionedTime := s.setOldCreationTimestamp(murMember1.Name)
```
xcoulon, line 146: it would probably be more explicit or understandable when reading the code here if the `manyManyDaysAgo` value was an argument of the `setOldCreationTimestamp` method.
But it's not a blocker.

test/e2e/base_user_test.go
```diff
@@ -130,7 +130,29 @@ func (s *baseUserIntegrationTest) deactivateAndCheckUser(userSignup *toolchainv1
...
+		states.SetVerificationRequired(us, true)
+	})
+	require.NoError(s.T(), err)
+	s.T().Logf("user signup '%s' reactivated, but requiring phone verification", userSignup.Name)
+
+	userSignup, err = s.hostAwait.WaitForUserSignup(userSignup.Name,
+		wait.UntilUserSignupHasConditions(ReactivatedAndVerificationRequired()...),
+		wait.UntilUserSignupHasStateLabel(toolchainv1alpha1.UserSignupStateLabelValueNotReady))
+	require.NoError(s.T(), err)
+	require.False(s.T(), states.Deactivated(userSignup), "usersignup should not be deactivated")
```
xcoulon, line 152: can you check that the `userSignup` you get after calling `s.hostAwait.WaitForUserSignup()` is the same (ie, same UID) as the one passed as an arg of the reactivateAndRequirePhoneVerification?
(at this stage, we can expect that the `activation-counter` annotation was incremented, I believe)

test/e2e/user_management_test.go
MatousJobanek: It will always be called with the same value - I don't see any benefit for doing it - that's why I just copied it and called the function `setOldCreationTimestamp`

xcoulon: ok, fair enough. As I said, it's just a cosmetic change anyways

test/e2e/base_user_test.go
MatousJobanek: I'm sorry, but I don't follow what you are asking for

MatousJobanek: > can you check that the userSignup you get after calling s.hostAwait.WaitForUserSignup() is the same (ie, same UID) as the one passed as an arg of the reactivateAndRequirePhoneVerification?

ok, now I probably know what you meant, but it doesn't make any sense to me why I should check that. I used the name of the UserSignup, so why should it be different?

> (at this stage, we can expect that the activation-counter annotation was incremented, I believe)

not yet - this user was reactivated, but it still requires phone verification. In other words, the UserSignup hasn't been approved yet, so the number cannot be incremented.

MatousJobanek:
closing since there is no way to modify the creation timestamp of UserSignup

PR #320, opened 2021-06-14
Sync config to members

rajivnathan:
Verifies that the mechanism for syncing MemberOperatorConfigs from the ToolchainConfig is working.

Related PR: https://github.com/codeready-toolchain/host-operator/pull/461


test/e2e/e2e_test.go
```diff
@@ -22,7 +22,25 @@ func TestE2EFlow(t *testing.T) {
...
+	defaultMemberConfig := testconfig.NewMemberOperatorConfig(testconfig.MemberStatus().RefreshPeriod("5s"))
+	member2Config := testconfig.NewMemberOperatorConfig(testconfig.MemberStatus().RefreshPeriod("10s"))
+	hostAwait.UpdateToolchainConfig(testconfig.AutomaticApproval().Disabled(), testconfig.Members().Default(defaultMemberConfig.Spec), testconfig.Members().SpecificPerMemberCluster(member2Await.ClusterName, member2Config.Spec))
+
+	t.Run("verify MemberOperatorConfigs synced from ToolchainConfig to member clusters", func(t *testing.T) {
+		t.Run("verify ToolchainConfig has synced status", func(t *testing.T) {
+			VerifyToolchainConfig(t, hostAwait, wait.UntilToolchainConfigHasSyncedStatus(ToolchainConfigSyncComplete()))
+		})
+		// TODO uncomment VerifyMemberOperatorConfig once the serviceaccount issue is resolved:
+		// memberoperatorconfigs.toolchain.dev.openshift.com "config" is forbidden: User "system:serviceaccount:rsenthil-member-operator:e2e-service-account" cannot get resource "memberoperatorconfigs" in API group "toolchain.dev.openshift.com" in the namespace "rsenthil-member-operator"
```
rajivnathan, line 34: I'm not sure why the serviceaccount is missing permissions to get memberoperatorconfig resources. I'll continue digging around on Monday but if anyone can point me to where that's set it'd be much appreciated.

alexeykazakov: Because it's missing here: https://github.com/codeready-toolchain/toolchain-common/blob/master/scripts/add-cluster.sh#L60

alexeykazakov: Hm.. actually it should be in the member client role and it's there: https://github.com/codeready-toolchain/toolchain-common/blob/master/scripts/add-cluster.sh#L90
Not in the host as I said above.

MatousJobanek: The reason is that e2e SA gets the same permissions as the member-operator SA. So, if the SA of the member-operator deployment doesn't have permissions to get that resource, then this SA won't have it either. 

And since there is no controller in member-operator that would require that permission, then it's missing in the ClusterRole: https://github.com/codeready-toolchain/member-operator/blob/master/deploy/cluster_role.yaml
The end-goal is to have separate ClusterRole for the e2e SA with broader privileges easier to manage

testsupport/wait/host.go
```diff
@@ -769,6 +769,57 @@ func (a *HostAwaitility) GetToolchainConfig() *toolchainv1alpha1.ToolchainConfig
...
+}
+
+func hasToolchainConfigSyncedStatus(t *testing.T, expectedCondition toolchainv1alpha1.Condition, toolchainConfig *toolchainv1alpha1.ToolchainConfig) bool {
+	if test.ContainsCondition(toolchainConfig.Status.Conditions, expectedCondition) {
+		t.Logf("status conditions match in ToolchainConfig")
+		return true
+	}
+	t.Logf("waiting for status condition of ToolchainConfig. Actual: '%+v'; Expected: '%+v'", toolchainConfig.Status.Conditions, expectedCondition)
+	return false
+}
```
MatousJobanek, line 788: couldn't be the content of `hasToolchainConfigSyncedStatus` part of `UntilToolchainConfigHasSyncedStatus`? Or is there any reason for keeping it in separated function?

testsupport/wait/host.go
```diff
@@ -769,6 +769,57 @@ func (a *HostAwaitility) GetToolchainConfig() *toolchainv1alpha1.ToolchainConfig
...
+
+// WaitForToolchainStatus waits until the ToolchainStatus is available with the provided criteria, if any
+func (a *HostAwaitility) WaitForToolchainConfig(criteria ...ToolchainConfigWaitCriterion) (*toolchainv1alpha1.ToolchainConfig, error) {
+	// there should only be one toolchain status with the name toolchain-status
+	name := "config"
+	toolchainConfig := &toolchainv1alpha1.ToolchainConfig{}
+	err := wait.Poll(a.RetryInterval, 2*a.Timeout, func() (done bool, err error) {
+		toolchainConfig = &toolchainv1alpha1.ToolchainConfig{}
+		// retrieve the toolchainstatus from the host namespace
```
MatousJobanek, line 797: copy-paste
```suggestion
// WaitForToolchainConfig waits until the ToolchainConfig is available with the provided criteria, if any
func (a *HostAwaitility) WaitForToolchainConfig(criteria ...ToolchainConfigWaitCriterion) (*toolchainv1alpha1.ToolchainConfig, error) {
	// there should only be one ToolchainConfig with the name "config"
	name := "config"
	toolchainConfig := &toolchainv1alpha1.ToolchainConfig{}
	err := wait.Poll(a.RetryInterval, 2*a.Timeout, func() (done bool, err error) {
		toolchainConfig = &toolchainv1alpha1.ToolchainConfig{}
		// retrieve the ToolchainConfig from the host namespace
```

testsupport/wait/member.go
```diff
@@ -880,6 +880,69 @@ func (a *MemberAwaitility) WaitForMemberStatus(criteria ...MemberStatusWaitCrite
...
+	}
+}
+
+func hasMatchingMemberOperatorConfig(t *testing.T, expectedMemberOperatorConfig *toolchainv1alpha1.MemberOperatorConfig, memberConfig *toolchainv1alpha1.MemberOperatorConfig) bool {
+	if reflect.DeepEqual(expectedMemberOperatorConfig, memberConfig) {
+		t.Logf("the MemberOperatorConfig matches the expected configuration")
+	}
+	t.Logf("waiting for MemberOperatorConfig to be synced. Actual: '%+v'; Expected: '%+v'", memberConfig, expectedMemberOperatorConfig)
+	return false
+}
```
MatousJobanek, line 912: the same question - couldn't it be part of one function `UntilMemberConfigMatches`?

testsupport/wait/host.go
rajivnathan: 🤷‍♂️  I noticed it was like that for a couple of the other functions eg. https://github.com/codeready-toolchain/toolchain-e2e/blob/03450280870985ff393b5cf445d14088d6dba2e2/testsupport/wait/member.go#L817 So I thought it was done for some reason.

I'll just make it a single function: ef269aeb7f7686ab2cc1992474c672dea13271a9

testsupport/wait/member.go
rajivnathan: Made this a single func too: ef269aeb7f7686ab2cc1992474c672dea13271a9

testsupport/wait/host.go
MatousJobanek: `hasMemberStatusUsageSet` is done for a reason - it's used at two places - in `wait/host.go` & `wait/member.go` files

MatousJobanek (commented):
Looks good, just waiting until you uncomment the lines :-)

MatousJobanek (approved):
ah, let me approve it right away because I believe that you won't forget to uncomment them :wink:

testsupport/wait/host.go
rajivnathan: ah, ok 🙂

rajivnathan:
@MatousJobanek I don't think I'll be able to uncomment them in this PR because I need to update the member operator to add a memberoperatorconfig controller that has the kubebuilder annotations and I don't think the e2e pairing will allow me to do both the host operator changes and the member operator changes required. You can correct me if I'm wrong but if not I will leave them commented in this PR and I already have the member operator changes ready to go up afterwards and I'll create a separate e2e PR to simply uncomment the VerifyMemberOperatorConfig stuff.

MatousJobanek:
ah, I thought that you could create a very simple controller (that does nothing but watching the CRD) with that missing marker. Then open a PR in member-operator repo with these changes + regenerated role and we can merge the PR before this one. Then we will just retest it

rajivnathan:
I already tested it locally with my working member operator controller. I just don't have the member operator PR up yet because I'm still working on the unit tests. At this point I think creating a very simple controller would actually be more work so are you okay with merging it as is?

PR #321, opened 2021-06-14
test option to force metric synchronization from resources

xcoulon:
E2E Tests for https://github.com/codeready-toolchain/host-operator/pull/463

Verify that the config flag triggers the counter synchronization based on UserSignup resources


alexeykazakov (approved):
Looks OK but I'm a bit concerned about the tests which mess with the operator deployments.. Just hopping we are not creating any flaky tests here which could effect other tests.

xcoulon:
> Looks OK but I'm a bit concerned about the tests which mess with the operator deployments.. Just hopping we are not creating any flaky tests here which could effect other tests.

I see what you mean 😬 
FWIW, I've added a `cleanup` func to ensure that the host-operator replicas is back to 1 after the test in 488d2ee.

test/e2e/toolchaincluster_test.go
```diff
@@ -174,3 +176,86 @@ func caBundle(bundle string) clusterOption {
 		c.Spec.CABundle = bundle
 	}
 }
+
+func TestForceSynchronization(t *testing.T) {
```
rajivnathan, line 180:
```suggestion
func TestForceMetricsSynchronization(t *testing.T) {
```
Maybe we can change this func name to `TestForceMetricsSynchronization`? I'm not sure it would be immediately obvious that this test is about metrics since it's in toolchaincluster_test.go

test/e2e/toolchaincluster_test.go
```diff
@@ -174,3 +176,86 @@ func caBundle(bundle string) clusterOption {
...
+	err := hostAwait.Client.List(context.TODO(), userSignups, client.InNamespace(hostAwait.Namespace))
+	require.NoError(t, err)
+	for _, userSignup := range userSignups.Items {
+		err := hostAwait.Client.Delete(context.TODO(), &userSignup)
+		require.NoError(t, err)
+	}
+	// then wait until all UserSignups have been deleted
+	for _, userSignup := range userSignups.Items {
+		hostAwait.WaitUntilUserSignupDeleted(userSignup.Name)
+	}
```
rajivnathan, line 204: This deletion of all usersignups logic seems like it could be useful as a common function. WDYT?

testsupport/metrics.go
```diff
@@ -89,6 +89,13 @@ func (m *MetricsAssertionHelper) WaitForMetricDelta(family string, delta float64
 	m.await.AssertMetricReachesValue(string(family), adjustedValue, labels...)
 }
 
+// WaitForMetric waits for the metric value to reach the exact value (ie, without prior baseline value)
+func (m *MetricsAssertionHelper) WaitForMetric(family string, value float64, labels ...string) {
+	// The delta is relative to the starting value, eg. If there are 3 usersignups when a test is started and we are waiting
+	// for 2 more usersignups to be created (delta is +2) then the actual metric value (adjustedValue) we're waiting for is 5
```
rajivnathan, line 95:
```suggestion
```
I guess this comment no longer applies?

testsupport/wait/host.go
```diff
@@ -377,7 +377,23 @@ func (a *HostAwaitility) WaitUntilBannedUserDeleted(name string) error {
 	})
 }
 
-// WaitUntilMasterUserRecordDeleted waits until MUR with the given name is deleted (ie, not found)
+// WaitUntilUserSignupDeleted waits until the UserSignup with the given name is deleted (ie, not found)
+func (a *HostAwaitility) WaitUntilUserSignupDeleted(name string) error {
+	return wait.Poll(a.RetryInterval, a.Timeout, func() (done bool, err error) {
+		mur := &toolchainv1alpha1.UserSignup{}
+		if err := a.Client.Get(context.TODO(), types.NamespacedName{Namespace: a.Namespace, Name: name}, mur); err != nil {
```
rajivnathan, line 384:
```suggestion
		usersignup := &toolchainv1alpha1.UserSignup{}
		if err := a.Client.Get(context.TODO(), types.NamespacedName{Namespace: a.Namespace, Name: name}, usersignup); err != nil {
```

rajivnathan (approved):
Cool stuff! 👍  Just had some minor comments/suggestiosn

test/e2e/toolchaincluster_test.go
xcoulon: yes, that's a good idea, let me do that

testsupport/metrics.go
xcoulon: indeed ;)

testsupport/wait/host.go
xcoulon: oh! thanks, I missed that 😮

test/e2e/toolchaincluster_test.go
```diff
@@ -174,3 +176,86 @@ func caBundle(bundle string) clusterOption {
 		c.Spec.CABundle = bundle
 	}
 }
+
+func TestForceMetricsSynchronization(t *testing.T) {
```
MatousJobanek, line 180: May I know why is this test in toolchaincluster_test.go? it doesn't have anything in common with ToolchainCluster resources.

test/e2e/toolchaincluster_test.go
xcoulon: done in 85154de

test/e2e/toolchaincluster_test.go
```diff
@@ -174,3 +176,86 @@ func caBundle(bundle string) clusterOption {
...
+	defer ctx.Cleanup()
+	hostAwait.UpdateToolchainConfig(
+		testconfig.AutomaticApproval().Enabled(),
+		testconfig.Metrics().ForceSynchronization(false))
+	t.Cleanup(func() {
+		err := hostAwait.ScaleDeployment(hostAwait.Namespace, "host-operator", 1)
+		require.NoError(t, err)
+	})
+	// before creating a batch of users, let's remove all remainings of previous tests, so
+	// we know exactly what numbers we're dealing with
```
MatousJobanek, line 193: wouldn't it be easier to list existing UserSignups and store the number at the beginning of the test?

test/e2e/toolchaincluster_test.go
```diff
@@ -174,3 +176,86 @@ func caBundle(bundle string) clusterOption {
...
+	err := hostAwait.Client.List(context.TODO(), userSignups, client.InNamespace(hostAwait.Namespace))
+	require.NoError(t, err)
+	for _, userSignup := range userSignups.Items {
+		err := hostAwait.Client.Delete(context.TODO(), &userSignup)
+		require.NoError(t, err)
+	}
+	// then wait until all UserSignups have been deleted
+	for _, userSignup := range userSignups.Items {
+		hostAwait.WaitUntilUserSignupDeleted(userSignup.Name)
+	}
```
MatousJobanek: rather than cleaning it up as a part of this test setup it would be much better (and probably also more proper) to do it at the end of each of the tests.

test/e2e/toolchaincluster_test.go
```diff
@@ -174,3 +176,86 @@ func caBundle(bundle string) clusterOption {
...
+	// now we can create our new users ;)
+	CreateMultipleSignups(t, ctx, hostAwait, memberAwait, 5) // 5 external users
```
MatousJobanek, line 197: we already provision a batch of users in e2e_test.go: https://github.com/codeready-toolchain/toolchain-e2e/blob/master/test/e2e/e2e_test.go#L64 so couldn't we move the test?

test/e2e/toolchaincluster_test.go
```diff
@@ -174,3 +176,86 @@ func caBundle(bundle string) clusterOption {
...
+				"external": toolchainStatus.Status.Metrics[toolchainv1alpha1.MasterUserRecordsPerDomainMetricKey]["external"] + 1, // increase current value by 1
+			}
+			toolchainStatus.Status.Metrics[toolchainv1alpha1.UserSignupsPerActivationAndDomainMetricKey] = toolchainv1alpha1.Metric{
+				"1,external": toolchainStatus.Status.Metrics[toolchainv1alpha1.UserSignupsPerActivationAndDomainMetricKey]["1,external"] + 1, // increase current value by 1
+			}
+			err = hostAwait.Client.Status().Update(context.TODO(), toolchainStatus)
+			require.NoError(t, err)
+			// when restarting the pod
+			err = hostAwait.ScaleDeployment(hostAwait.Namespace, "host-operator", 1)
+			require.NoError(t, err)
```
MatousJobanek, line 231: I'm just curious - is it really safe to do that? Won't be the deployment scaled up by the operator controller in the meantime?

MatousJobanek (commented):
TBH, I'm not a fan of scaling the deployment down to zero & tampering the ToolchainStatus in the meantime. Wouldn't be more safe to:
1. provision users (2 users (1 external & 1 internal is more than enough)
2. let host-operator count them
3. modify manually the `toolchain.dev.openshift.com/verification-counter` annotations in the UserSignups (as a result, the ToolchianStatus won't have correct numbers)
4. set the `ForceSynchronization` flag
5. restart the deployment
6. verify that ToolchainStatus has new "activation" counts

test/e2e/toolchaincluster_test.go
xcoulon: no, the host-operator pod get deleted, that's all.

test/e2e/toolchaincluster_test.go
xcoulon: because once again I got mixed with toolchainstatus 😬

xcoulon:
> TBH, I'm not a fan of scaling the deployment down to zero & tampering the ToolchainStatus in the meantime. Wouldn't be more safe to:
> 
> 1. provision users (2 users (1 external & 1 internal is more than enough)
> 2. let host-operator count them
> 3. modify manually the `toolchain.dev.openshift.com/verification-counter` annotations in the UserSignups (as a result, the ToolchianStatus won't have correct numbers)
> 4. set the `ForceSynchronization` flag
> 5. restart the deployment
> 6. verify that ToolchainStatus has new "activation" counts

ok, I can do that

test/e2e/toolchainstatus_test.go
```diff
@@ -0,0 +1,94 @@
...
+		testconfig.Metrics().ForceSynchronization(false))
+
+	userSignups := CreateMultipleSignups(t, ctx, hostAwait, memberAwait, 2)
+
+	// delete the current toolchainstatus/toolchain-status resource and restart the host-operator pod,
+	// so we can start with accurate counters/metrics (and not get flaky because of previous tests)
+	err := hostAwait.DeleteToolchainStatus("toolchain-status")
+	require.NoError(t, err)
+	err = hostAwait.DeletePods(client.InNamespace(hostAwait.Namespace), client.MatchingLabels{"name": "host-operator"})
+	require.NoError(t, err)
```
MatousJobanek, line 32: May I know why? We shouldn't expect that the ToolchainStatus doesn't match the actual state/count of the resources. Otherwise, we wouldn't be able to verify the metrics in other tests, or am I missing something?

test/e2e/toolchainstatus_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+	metricsAssertion := InitMetricsAssertion(t, hostAwait, []string{memberAwait.ClusterName, member2Await.ClusterName})
+
+	t.Run("verify metrics are still correct after restarting pod", func(t *testing.T) {
+		// when
+		err := hostAwait.DeletePods(client.InNamespace(hostAwait.Namespace), client.MatchingLabels{"name": "host-operator"})
```
MatousJobanek, line 38: May I know what is the purpose of this deletion of the host-operator pod?

MatousJobanek (commented):
I have questions regarding the setup of the test and the first subtest. 
The following subtest `tampering activation-counter annotations` looks good though :+1:

test/e2e/toolchainstatus_test.go
xcoulon: that's right. I just wanted to make sure that we started the test with accurate metrics compared to existing resources, but indeed, they should already match. If it's a real concern, I can remove that part

test/e2e/toolchainstatus_test.go
xcoulon: that's a bit superfluous and unnecessary, indeed. I removed it in cff0582

test/e2e/toolchainstatus_test.go
MatousJobanek: Understood - thanks for explanation :+1:

MatousJobanek: btw, when the work on the `ToolchainConfig_controller` is finished, then we should drop the deletion of the pods. The `ToolchainConfig_controller` should watch the `ToolchainStatus` resource and recreate it. 
Could you please add a `TODO:` comment to this part?

@rajivnathan  do we have any Jira issue for this effort? - I mean watching the `ToochainStatus` CRD in `ToolchainConfig_controller`

MatousJobanek (approved):
Thanks for changing the tests :+1: :pray:

test/e2e/toolchainstatus_test.go
xcoulon: are we also going to trigger a full sync if the new `metrics.forceSynchronization` field is set to `true`?

PR #322, opened 2021-06-15
Ensure namespace deleted before userAccount 

ranakan19:
ensure cascade deletion


test/e2e/e2e_test.go
```diff
@@ -200,6 +202,90 @@ func TestE2EFlow(t *testing.T) {
...
+		johnMUR, err := hostAwait.WaitForMasterUserRecord(johnsmithName)
+		require.NoError(t, err)
+
+		fmt.Printf("\n >>>>>>>>>>>>>>>>>>>>>>>>>>>> MUR fetched>>>>>>>>>>>>>>>>")
+		// set deletion timestamp and add finalizer
+		deletionTS := metav1.Now()
+		johnMUR.DeletionTimestamp = &deletionTS
+		coputil.AddFinalizer(johnMUR, "finalizer.toolchain.dev.openshift.com")
+		if err := hostAwait.Client.Update(context.TODO(), johnMUR); err != nil {
+			fmt.Errorf("failed to set deletionTS and finalizer")
```
xcoulon, line 216: here you should do something like `require.NoError(t, err)` instead of just printing the error, because otherwise the code will continue to execute.

test/e2e/e2e_test.go
```diff
@@ -200,6 +202,90 @@ func TestE2EFlow(t *testing.T) {
...
+	t.Run("verify userAccount is not deleted if namespace is not deleted", func(t *testing.T) {
+		// given
+		johnMUR, err := hostAwait.WaitForMasterUserRecord(johnsmithName)
+		require.NoError(t, err)
+
+		fmt.Printf("\n >>>>>>>>>>>>>>>>>>>>>>>>>>>> MUR fetched>>>>>>>>>>>>>>>>")
+		// set deletion timestamp and add finalizer
+		deletionTS := metav1.Now()
+		johnMUR.DeletionTimestamp = &deletionTS
+		coputil.AddFinalizer(johnMUR, "finalizer.toolchain.dev.openshift.com")
```
xcoulon, line 214: why do you want to add a finalizer on the MUR?

test/e2e/e2e_test.go
```diff
@@ -200,6 +202,90 @@ func TestE2EFlow(t *testing.T) {
...
+		memberPod := corev1.Pod{
+			TypeMeta: metav1.TypeMeta{
+				Kind:       "Pod",
+				APIVersion: "v1",
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "test-useraccount-delete-1",
+				Namespace: "johnsmith-dev",
+				Finalizers: []string{
+					"kubernetes",
```
xcoulon, line 234: you should add another finalizer here. `kubernetes` is the default finalizer IIRC, but you should add your own to prevent the pod deletion and make sure that the namespace is stuck in a `terminating` phase until your finalizer on the pod is removed

test/e2e/e2e_test.go
ranakan19: its the entry condition to manageCleanup here https://github.com/codeready-toolchain/host-operator/blob/master/controllers/masteruserrecord/masteruserrecord_controller.go#L119 - it expects the MUR to have both a timestamp and finalizer before calling manageCleanup, which in-turn deletes userAccount.

test/e2e/e2e_test.go
ranakan19: the namespace is stuck in terminating state even with kubernetes finalizer

ranakan19:
closing this PR, since https://github.com/codeready-toolchain/toolchain-e2e/pull/336 now handles the e2e tests

PR #324, opened 2021-06-16
Verify memberoperatorconfigs

rajivnathan:
Verifies the memberoperatorconfigs for each member cluster.

Related PR: https://github.com/codeready-toolchain/member-operator/pull/265


alexeykazakov (commented):
Maybe I've missed that but do you have a test to verify that the update member config is picked up by the controller/cache?

rajivnathan:
@alexeykazakov no, not yet because the only thing it picks up at the moment is the memberstatus refresh period and I'm not sure that's worth testing specifically. But on the other hand it doesn't seem like we have an e2e test for it. WDYT, use the refresh period or add a test later when we have a few more choices. eg. Some che config

alexeykazakov:
I will leave it to you. Just don't forget to add a test later if you want to wait for it.

alexeykazakov (approved):

PR #325, opened 2021-06-18
feat: drop operator-sdk framework

MatousJobanek:
This PR drops the usage of operator-sdk from e2e tests. The main reason is that the latest operator-sdk doesn't provide the same tools & functionalities. Also, the SDK team discourages users from using operator-sdk and recommends using EnvTest for writing tests for operators. However, EnvTest is primarily meant for testing operators without a cluster - it's something more like a "real" alternative to FakeClient for integration tests. In addition, EnvTest doesn't provide any useful alternatives to operator-sdk functions we use right now.
In other words - we would need to replace the operator-sdk functions with our solution as part of the migration anyway. I just decided to do it in a separate PR.

I'm keeping operator-sdk in the Dockerfile.tools as well as in our READMEs for now - the reasons are:
* I'll remove it from Dockerfile.tools as the next step in separate PR in all repos together
* I'll keep it in READMEs because it will be needed as a setup part in the future - just the latest version. The setup part will be executed as a GH Actions workflow and not in openshift-ci - that's why we don't need it in the dockerfile.


alexeykazakov (approved):

make/perf.mk
```diff
@@ -22,5 +22,5 @@ perf-run:
 	oc get toolchaincluster -n $(MEMBER_NS)
 	oc get toolchaincluster -n ${MEMBER_NS_2}
 	-oc new-project $(TEST_NS) --display-name perf-tests 1>/dev/null
-	ARTIFACT_DIR=${ARTIFACT_DIR} USER_COUNT=3000 USER_BATCH_SIZE=100 MEMBER_NS=${MEMBER_NS} MEMBER_NS_2=${MEMBER_NS_2} HOST_NS=${HOST_NS} REGISTRATION_SERVICE_NS=${REGISTRATION_SERVICE_NS} operator-sdk test local ./test/perf --no-setup --operator-namespace $(TEST_NS) --verbose --go-test-flags "-timeout=120m -failfast" || \
+	ARTIFACT_DIR=${ARTIFACT_DIR} USER_COUNT=3000 USER_BATCH_SIZE=100 MEMBER_NS=${MEMBER_NS} MEMBER_NS_2=${MEMBER_NS_2} HOST_NS=${HOST_NS} REGISTRATION_SERVICE_NS=${REGISTRATION_SERVICE_NS} go test ./test/e2e -v -timeout=60m -failfast || \
```
xcoulon, line 25: shouldn't we run the tests in `./test/perf` ?

MatousJobanek: yeah, right :man_facepalming: thanks
```suggestion
	ARTIFACT_DIR=${ARTIFACT_DIR} USER_COUNT=3000 USER_BATCH_SIZE=100 MEMBER_NS=${MEMBER_NS} MEMBER_NS_2=${MEMBER_NS_2} HOST_NS=${HOST_NS} REGISTRATION_SERVICE_NS=${REGISTRATION_SERVICE_NS} go test ./test/perf -v -timeout=60m -failfast || \
```

PR #331, opened 2021-06-24
Autoscaler buffer priority should be higher than autoscaler threshold

alexeykazakov:
Tests for https://github.com/codeready-toolchain/member-operator/pull/271


rajivnathan (approved):
I don't really understand the reason for the change but I'm assuming you know what you're doing. 🙂  Is the threshold value something that is configured in the autoscaler?

alexeykazakov:
Yes. See https://issues.redhat.com/browse/OHSS-4812 for details :)

MatousJobanek (approved):

PR #336, opened 2021-06-30
E2E tests for ensuring namespace is deleted before userAccount.

ranakan19:
E2E tests foreground cascade deletion for related story: https://issues.redhat.com/browse/CRT-897

related PR: https://github.com/codeready-toolchain/member-operator/pull/262

opening this PR inplace of https://github.com/codeready-toolchain/toolchain-e2e/pull/322 for PR name matching


test/e2e/e2e_test.go
```diff
@@ -222,6 +224,92 @@ func TestE2EFlow(t *testing.T) {
...
+		assert.Error(t, err, "laracroft-dev namespace got deleted")
+
+		err = memberAwait.WaitUntilNSTemplateSetDeleted(laraUserName)
+		assert.Error(t, err, "NSTemplateSet got deleted")
+
+		err = memberAwait.WaitUntilUserAccountDeleted(laraUserName)
+		assert.Error(t, err, "userAccount got deleted")
+
+		err = hostAwait.WaitUntilMasterUserRecordDeleted(laraUserName)
+		require.Error(t, err)
```
MatousJobanek, line 281: In these calls, you wait for the whole timeout (which is 1 minute each) and check if the resource was deleted or not. You expect that all calls should return error - that the resource wasn't deleted during the 1 minute.
Since you do it 4 times, it probably waits 4 minutes - in that case you hit the threshold in MUR controller: https://github.com/codeready-toolchain/host-operator/blob/master/controllers/masteruserrecord/masteruserrecord_controller.go#L314-L316 which means that k8s starts scheduling and prolonging the periods between the reconciles. This is most likely the reason why it takes so long until the MUR is deleted when the UserSignup is successfully removed.

Instead of waiting 4 minutes, you should just wait for example 10 or max 20 seconds and check all resources alltogether. You can take a look at the other wait methods which check that a resource is not removed.

ranakan19: This makes sense, trying it now.

test/e2e/e2e_test.go
```diff
@@ -222,6 +224,103 @@ func TestE2EFlow(t *testing.T) {
...
+		memberPod := corev1.Pod{
+			TypeMeta: metav1.TypeMeta{
+				Kind:       "Pod",
+				APIVersion: "v1",
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      podName,
+				Namespace: userNamespace,
+				Finalizers: []string{
+					"kubernetes",
```
alexeykazakov, line 245: Shouldn't we use our own finalizer to make sure we have full control over it and it's not managed by something else?
For example:
```suggestion
					"toolchain-e2e-tests",
```

test/e2e/e2e_test.go
```diff
@@ -222,6 +224,103 @@ func TestE2EFlow(t *testing.T) {
...
+		require.NoError(t, err)
+
+		// now remove finalizer from pod and check all are deleted
+		err = memberAwait.Client.Get(context.TODO(), types.NamespacedName{Namespace: userNamespace, Name: podName}, pod)
+		require.NoError(t, err)
+		require.Equal(t, pod.Finalizers[0], "kubernetes")
+		pod.Finalizers = nil
+		err = memberAwait.Client.Update(context.TODO(), pod)
+		require.NoError(t, err)
+		require.Equal(t, len(pod.Finalizers), 0)
```
alexeykazakov, line 305: It's safer to update resources in a pool function to avoid possible conflicts if the resource (the pod in this case) has been changed between `Get()` and `Update()` but some controller in the cluster. Otherwise we could create another flaky test.
We what we use for modifying MUR for example: https://github.com/codeready-toolchain/toolchain-e2e/blob/8592e7778727ac79295a76c8ef126f278f333e92/testsupport/wait/host.go#L117
For example we can introduce something like:
```
func (a *MemberAwaitility) UpdatePod(namespace, podName string, modifyPod func(pod *corev1.Pod)) (*corev1.Pod, error) {
...
}
```
And use it:
```
err = memberAwait.UpdatePod(namespace, podName, func(pod *corev1.Pod) {
    pod.Finalizers[0] = nil
})
```

xcoulon (approved):

testsupport/wait/member.go
```diff
@@ -151,6 +164,35 @@ func (a *MemberAwaitility) WaitForUserAccount(name string, criteria ...UserAccou
...
+				a.T.Logf("waiting for useraccount '%s' in namespace '%s' to be deleted, but userAccount not found", name, a.Namespace)
+				return false, nil
+			}
+			return false, err
+		}
+
+		if obj.DeletionTimestamp == nil {
+			a.T.Logf("useraccount '%s' in namespace '%s' does not have deletion timestamp", name, a.Namespace)
+			return false, nil
+		}
```
rajivnathan, line 183: Couldn't this be another UserAccountWaitCriteria?
eg. 
```
func UntilUserAccountIsBeingDeleted(conditions ...toolchainv1alpha1.Condition) UserAccountWaitCriterion {
	return func(a *MemberAwaitility, ua *toolchainv1alpha1.UserAccount) bool {
		if obj.DeletionTimestamp == nil {
			a.T.Logf("useraccount '%s' in namespace '%s' does not have deletion timestamp", name, a.Namespace)
			return false
		}
                 return true
	}
}
```
Then we don't need the `WaitForUserAccountBeingDeletedWithCondition` func anymore?

It could then be used like this:
```
memberAwait.WaitForUserAccount(laraUserName, wait.UntilUserAccountIsBeingDeleted(), wait.UntilUserAccountHasCondition(TerminatingUserAccount()))
```

testsupport/wait/member.go
```diff
@@ -167,6 +209,19 @@ func UntilNSTemplateSetHasConditions(conditions ...toolchainv1alpha1.Condition)
 	}
 }
 
+// UntilNSTemplateSetHasCondition returns a `NSTemplateSetWaitCriterion` which checks that the given
+// NSTemlateSet contains the given condition
+func UntilNSTemplateSetHasCondition(condition toolchainv1alpha1.Condition) NSTemplateSetWaitCriterion {
```
rajivnathan, line 214: Same as for UntilUserAccountContainsCondition

testsupport/wait/member.go
```diff
@@ -203,6 +258,34 @@ func (a *MemberAwaitility) WaitForNSTmplSet(name string, criteria ...NSTemplateS
 	return nsTmplSet, err
 }
 
+// WaitForNSTmplSetIsBeingDeletedWithCondition wait until the NSTemplateSet with the given name has deletion timestamp and matches optional conditions
+func (a *MemberAwaitility) WaitForNSTmplSetIsBeingDeletedWithCondition(name string, criteria ...NSTemplateSetWaitCriterion) (*toolchainv1alpha1.NSTemplateSet, error) {
```
rajivnathan, line 262: Same as for WaitForUserAccountBeingDeletedWithCondition

test/e2e/e2e_test.go
```diff
@@ -237,6 +239,105 @@ func TestE2EFlow(t *testing.T) {
...
+		_, err = memberAwait.WithRetryOptions(wait.TimeoutOption(time.Second * 10)).WaitForNamespaceInTerminating(userNamespace)
+		require.NoError(t, err)
+
+		// Check all corresponding resources are blocked by the terminating ns and have expected conditions
+		NSTemplateSetCondition := toolchainv1alpha1.Condition{
+			Type:    toolchainv1alpha1.ConditionReady,
+			Status:  corev1.ConditionFalse,
+			Reason:  toolchainv1alpha1.NSTemplateSetTerminatingFailedReason,
+			Message: "user namespace laracroft-dev deletion was triggered but is not complete yet, something could be blocking ns deletion",
+		}
```
rajivnathan, line 301: Should we move this to a function in testsupport/conditions.go too? It can just accept a parameter for the namespace that we can set to `laracroft-dev`?

rajivnathan (approved):
Looks great overall! 🥇  Just left a few minor comments

testsupport/wait/member.go
```diff
@@ -127,6 +127,31 @@ func UntilUserAccountHasConditions(conditions ...toolchainv1alpha1.Condition) Us
...
+			a.T.Logf("status conditions found in UserAccount '%s`", ua.Name)
+			return true
+		}
+		a.T.Logf("waiting for status condition of UserSignup '%s'. Actual: '%+v'; Expected: '%+v'", ua.Name, ua.Status.Conditions, condition)
+		return false
+	}
+}
+
+// UntilUserAccountIsBeingDeleted returns a `UserAccountWaitCriterion` which checks that the given
+//// USerAccount contains the given condition has deletion timestamp set
```
rajivnathan, line 144:
```suggestion
// USerAccount has the deletion timestamp set
```

MatousJobanek (commented):
I saw that the added test is failing. No idea why the UserAccount is not in a termination state.
```
     member.go:138: waiting for status condition of UserSignup 'laracroft'. Actual: '[{Type:Ready Status:True LastTransitionTime:2021-07-28 02:46:16 +0000 UTC Reason:Provisioned Message: LastUpdatedTime:<nil>}]'; Expected: '{Type:Ready Status:False LastTransitionTime:0001-01-01 00:00:00 +0000 UTC Reason:Terminating Message:deleting user/identity LastUpdatedTime:<nil>}'
=== CONT  TestE2EFlow/verify_userAccount_is_not_deleted_if_namespace_is_not_deleted
    e2e_test.go:304: 
        	Error Trace:	e2e_test.go:304
        	Test:       	TestE2EFlow/verify_userAccount_is_not_deleted_if_namespace_is_not_deleted 
```

PR #341, opened 2021-07-09
chore: apply config files from makefile

MatousJobanek:
Since the scripts will be moved to CICD repo, then all config files should be applied from makefile

### DIFF
--- a/make/test.mk
+++ b/make/test.mk
@@ -167,7 +167,7 @@ create-member1:
 .PHONY: create-member2
 create-member2:
 ifeq ($(SECOND_MEMBER_MODE),true)
-	@echo "Deploying member operator to ${MEMBER_NS_2}..."
+	@echo "Deploying second member operator to ${MEMBER_NS_2}..."
 	$(MAKE) create-project PROJECT_NAME=${MEMBER_NS_2}
 	oc apply -f deploy/member2-operator/member-operator-config-map.yaml -n ${MEMBER_NS_2}
 endif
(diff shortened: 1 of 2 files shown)


make/test.mk
```diff
@@ -151,44 +157,37 @@ get-and-publish-host-operator:
...
-	if [[ ${SECOND_MEMBER_MODE} == true ]]; then curl -sSL https://raw.githubusercontent.com/codeready-toolchain/toolchain-common/master/scripts/add-cluster.sh | bash -s -- -t member -tn e2e -mn $(MEMBER_NS_2) -hn $(HOST_NS) -s -mm 2; fi
+.PHONY: create-member1
+create-member1:
+	@echo "Deploying member operator to $(MEMBER_NS)..."
+	$(MAKE) create-project PROJECT_NAME=${MEMBER_NS}
+
+.PHONY: create-member2
+create-member2:
+ifeq ($(SECOND_MEMBER_MODE),true)
+	@echo "Deploying member operator to ${MEMBER_NS_2}..."
```
rajivnathan, line 170:
```suggestion
	@echo "Deploying second member operator to ${MEMBER_NS_2}..."
```

make/test.mk
```diff
@@ -151,44 +157,37 @@ get-and-publish-host-operator:
...
-	-oc project $(HOST_NS)
-	-oc apply -f deploy/host-operator/secrets.yaml -n $(HOST_NS)
-	-oc apply -f deploy/host-operator/host-operator-config-map.yaml -n $(HOST_NS)
+	@echo "Deploying host operator to ${HOST_NS}..."
+	$(MAKE) create-project PROJECT_NAME=${HOST_NS}
 
 .PHONY: create-host-resources
 create-host-resources:
-	oc apply -f deploy/host-operator/nstemplatetier-base.yaml -n $(HOST_NS)
+	oc apply -f deploy/host-operator/ -n ${HOST_NS}
```
rajivnathan, line 185: Maybe we should create a folder within `deploy/host-operator` for each environment? eg. `deploy/host-operator/e2e-tests` and `deploy/host-operator/dev` and then chance this to `oc apply -f deploy/host-operator/${ENVIRONMENT} -n ${HOST_NS}` WDYT?

MatousJobanek: sure, but right now there is only one env. Support for multiple environments will come with your PR - that's why I left the original structure/format :-)

rajivnathan (approved):

PR #343, opened 2021-07-20
chore: reduce test verbosity

xcoulon:
down from 27.6k lines to 12.2k lines and 9.7MB to 1.3MB of logs
when running the whole suite of e2e tests.

Fixes https://issues.redhat.com/browse/CRT-1159


xcoulon:
@MatousJobanek @rajivnathan with the latest changes in https://github.com/codeready-toolchain/toolchain-e2e/pull/343/commits/4ab81e5a1301705b809816616ed0d30eb55b93bb, the error message in the console look like this, with a focus on the diff (see the red/green characters)
image

Also, to avoid getting disturbed with the `LastTransitionTime` and `LastUpdateTime` fields of the `Condition`, these are reset  before generating the diff message.

WDYT?

xcoulon:
@MatousJobanek @rajivnathan with the change in https://github.com/codeready-toolchain/toolchain-e2e/pull/343/commits/911d489358fd7508607146954956dc5c619bfbd2 that intentionally triggered a failure, here's how the test report looks on https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/codeready-toolchain_toolchain-e2e/343/pull-ci-codeready-toolchain-toolchain-e2e-master-e2e/1418557622075789312 

image

testsupport/wait/host.go
```diff
@@ -63,20 +64,27 @@ func (a *HostAwaitility) WaitForMasterUserRecord(name string, criteria ...Master
...
 	})
+	// log message if an error occurred
+	if err != nil {
+		a.T.Logf("failed to find MasterUserRecord '%s' with matching criteria: %v", name, err)
+		for _, c := range criteria {
+			if !c.Match(a, mur) {
+				a.T.Log(c.Diff(mur))
+			}
+		}
+	}
```
MatousJobanek, line 87: it looks like that it could be the same for all wait functions - couldn't we introduce some more abstract wait framework that would do such operations automatically? You know - to avoid code duplications

testsupport/wait/host.go
```diff
@@ -162,138 +168,158 @@ func (a *HostAwaitility) UpdateUserSignupSpec(userSignupName string, modifyUserS
...
-		a.T.Logf("waiting for status condition of MasterUserRecord '%s'. Actual: '%+v'; Expected: '%+v'", mur.Name, mur.Status.Conditions, conditions)
-		return false
+func UntilMasterUserRecordHasConditions(expected ...toolchainv1alpha1.Condition) MasterUserRecordWaitCriterion {
+	return MasterUserRecordWaitCriterion{
+		Match: func(a *HostAwaitility, actual *toolchainv1alpha1.MasterUserRecord) bool {
+			return test.ConditionsMatch(actual.Status.Conditions, expected...)
+		},
+		Diff: func(actual *toolchainv1alpha1.MasterUserRecord) string {
+			return fmt.Sprintf("expected conditions to match:\n%s", Diff(actual.Status.Conditions, expected))
+		},
```
MatousJobanek, line 211: why do they use different ways of printing the diff when they both verify conditions?

testsupport/wait/host.go
```diff
@@ -162,138 +168,158 @@ func (a *HostAwaitility) UpdateUserSignupSpec(userSignupName string, modifyUserS
...
 
 func WithMurName(name string) MasterUserRecordWaitCriterion {
-	return func(a *HostAwaitility, mur *toolchainv1alpha1.MasterUserRecord) bool {
-		return mur.Name == name
+	return MasterUserRecordWaitCriterion{
+		Match: func(a *HostAwaitility, actual *toolchainv1alpha1.MasterUserRecord) bool {
+			return actual.Name == name
+		},
+		Diff: func(actual *toolchainv1alpha1.MasterUserRecord) string {
+			return "" // unused
```
MatousJobanek, line 236: Why? We should print that there was no MUR with the given name found.

testsupport/wait/host.go
```diff
@@ -162,138 +168,158 @@ func (a *HostAwaitility) UpdateUserSignupSpec(userSignupName string, modifyUserS
...
-		}
-		a.T.Logf("waiting for status condition of UserSignup '%s' to contain the expected. Actual: '%+v'; Expected: '%+v'", ua.Name, ua.Status.Conditions, condition)
-		return false
+func ContainsCondition(expected toolchainv1alpha1.Condition) UserSignupWaitCriterion {
+	return UserSignupWaitCriterion{
+		Match: func(a *HostAwaitility, actual *toolchainv1alpha1.UserSignup) bool {
+			return test.ContainsCondition(actual.Status.Conditions, expected)
+		},
+		Diff: func(actual *toolchainv1alpha1.UserSignup) string {
+			return fmt.Sprintf("expected conditions to contain: %s.\n\tactual: %s", spew.Sdump(expected), spew.Sdump(actual.Status.Conditions))
```
MatousJobanek, line 289: The same question - why don't we use the diff?

MatousJobanek: This line makes me also thinking - wouldn't it be easier to marshal the objects (interfaces) into YAML format and print the expected and actual one? you could also use some tool for comparing the YAMLs.

MatousJobanek (commented):
TBH, I didn't go properly through the whole PR, I have a few questions.
The diff that you showed in your comment looks nice, but I think that using the YAML format could be better also for later troubleshooting/investigating, WDYT? IIRC, using YAML format was one of @sbryzak's requests.

make/test.mk
```diff
@@ -196,7 +196,8 @@ create-host-resources:
 ifeq ($(HOST_REPO_PATH),)
 	$(eval HOST_REPO_PATH = /tmp/codeready-toolchain/host-operator)
 endif
-	oc apply -f deploy/host-operator/${ENVIRONMENT}/ -n ${HOST_NS}
+	# ignore if these resources already exist (nstemplatetiers may have already been created by operator)
+	-oc create -f deploy/host-operator/${ENVIRONMENT}/ -n ${HOST_NS}
```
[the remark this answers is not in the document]
xcoulon: In some test executions, I had troubles with the NSTemplateTiers whose TierTemplate version was still `-000000` and I suspect that the `oc apply` command was executed after the operator already started, which means that the NSTemplateTier resources were reset again, and the e2e tests would fail. So I thought we may want to skip the `oc apply` (ie, ignore if `oc create` fails) when there's such a race condition

testsupport/wait/awaitility.go
```diff
@@ -367,15 +356,15 @@ func (a *Awaitility) AssertMetricReachesValue(family string, expectedValue float
 // WaitUntilMetricHasValueOrMore waits until the exposed metric with the given family
 // and label key-value pair has reached the expected value (or more)
 func (a *Awaitility) WaitUntilMetricHasValueOrMore(family string, expectedValue float64, labels ...string) error {
-	a.T.Logf("Waiting for metric '%s{%v}' to reach '%v' or more", family, labels, expectedValue)
+	a.T.Logf("waiting for metric '%s{%v}' to reach '%v' or more", family, labels, expectedValue)
 	err := wait.Poll(a.RetryInterval, a.Timeout, func() (done bool, err error) {
 		value, err := getMetricValue(a.RestConfig, a.MetricsURL, family, labels)
 		if err != nil {
-			a.T.Logf("Waiting for metric '%s{%v}' to reach '%v' or more but error occurred: %s", family, labels, expectedValue, err.Error())
+			a.T.Logf("waiting for metric '%s{%v}' to reach '%v' or more but error occurred: %s", family, labels, expectedValue, err.Error())
```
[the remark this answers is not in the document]
xcoulon: simply because I missed that

testsupport/wait/diff.go
```diff
@@ -0,0 +1,93 @@
...
+		}
+		return result
+	case toolchainv1alpha1.UserAccountStatusEmbedded:
+		return toolchainv1alpha1.UserAccountStatusEmbedded{
+			Cluster:   obj.Cluster,
+			SyncIndex: obj.SyncIndex,
+			UserAccountStatus: toolchainv1alpha1.UserAccountStatus{
+				Conditions: ResetTimeFields(obj.UserAccountStatus.Conditions).([]toolchainv1alpha1.Condition),
+			},
+		}
```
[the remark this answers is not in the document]
xcoulon: only those containing conditions, so we don't have diffs on timestamps, which would be just extra noise

testsupport/wait/host.go
xcoulon: the `UntilMasterUserRecordHasCondition` expects the resources conditions to contain the given `expected` condition (among others), whereas `UntilMasterUserRecordHasConditions` expects a full match. But I'm not sure if this answers your question

xcoulon:
> TBH, I didn't go properly through the whole PR, I have a few questions.
> The diff that you showed in your comment looks nice, but I think that using the YAML format could be better also for later troubleshooting/investigating, WDYT? IIRC, using YAML format was one of @sbryzak's requests.


I'm personally happy with the output as it appears above, i.e., showing Golang structs. There are useful metadata such as string length or slice sizes that can help in spotting the differences, IMO

testsupport/wait/host.go
xcoulon: the `Diff` method isn't called, but the caller (test) will report an error. I think that covers our needs.

make/test.mk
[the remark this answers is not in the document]
MatousJobanek: in which cases did it happen? when you executed either `make test-e2e` or `make dev-deploy-e2e`? I haven't seen such an error, that's why I'm asking...

testsupport/wait/host.go
MatousJobanek: TBH, it's very confusing when it should use/print the diff and when the full dump.

testsupport/wait/host.go
MatousJobanek: I thought that printing the diff or the error log is a responsibility of the criteria that wasn't met, not the responsibility of the caller. This is at least what I see in the other cases.

testsupport/wait/diff.go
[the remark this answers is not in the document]
MatousJobanek: ok, but we use conditions probably in all our CRs, right? So is it really too bad to see the diffs of the timestamps? I mean - is it worth it?

just an idea - if we transform the object into a YAML format and then into a generic map of `map[interface{}]interface{}` couldn't we then transitively reset these timestamps in a generic way?

MatousJobanek:
> I'm personally happy with the output as it appears above, i.e., showing Golang structs. There are useful metadata such as string length or slice sizes that can help in spotting the differences, IMO

Ok, I see. For me, the additional data is more or less just a noise :man_shrugging: But that's just my point of view.
To be clear, I don't insist on using YAML format. I just thought that it could be fine since we use it in `oc` command and at other places too. But if the rest of the team is fine with this format, I'll be fine too ;-)

testsupport/wait/diff.go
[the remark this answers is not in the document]
xcoulon: I think it's simpler to keep it this way for now, and adapt the code when other resources with conditions are compared. Or we could use the `reflect` package to make things more generic, but at the cost of maintainability.
We could also just ignore this part, but then we'll have extra noise each time there's a diff, and I think we should do something to avoid it.

testsupport/wait/host.go
xcoulon: it's implementation details ;) The idea is that the test will call the `Diff` func and to display the message when one or more condiitions did not match, and provide the most accurate message to help understand what did not match

testsupport/wait/host.go
xcoulon: right, my answer wasn't accurate. I'll add a message and display it

testsupport/wait/host.go
xcoulon: the `criteria` type may differ, so I'm not sure we can factorize much, but I will take a look

testsupport/wait/host.go
xcoulon: `spew.Sdump` already marshalls the objects. I think that in the case of finding a conditions in a set of conditions, dumping the objects is enough. I would suggest we improve the output of `Diff` in the future if it's not clear enough as it is proposed here.

testsupport/wait/host.go
xcoulon: done in 1641853

testsupport/wait/host.go
xcoulon: done in a941d0c

testsupport/wait/awaitility.go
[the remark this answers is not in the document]
xcoulon: addressed in accf2a8 (some funcs were unused so I removed them)

make/test.mk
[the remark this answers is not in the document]
xcoulon: it happened when running `IMAGE_BUILDER=docker make test-e2e` locally

testsupport/wait/diff.go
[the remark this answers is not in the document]
MatousJobanek: > We could also just ignore this part, but then we'll have extra noise each time there's a diff, and I think we should do something to avoid it.

If there is no generic way of avoiding it, then - personally - I would accept the extra noise so we don't have to have such rewrites of the entire CR objects. I'm just afraid of the complexity because we would need to do it for all CRs and maintain it as soon as we add/remove/modify any parameter.

testsupport/wait/host.go
MatousJobanek: I see. I know that it's an implementation detail but we will be the ones who will implement it - it's not really clear to me in which cases it should print the dump and in which the diff.

MatousJobanek: I mean - wouldn't it be easier to print the actual content vs the expected one? That logic could be fully generic and it could provide functions like:
```go
PrintExpectedVsActual(expected, actual interface{})
PrintActualThatDoesNotContain(actual, shouldContain interface{})
```
all the objects can be marshaled either as a YAML or as a  JSON, right? If we use correct indentation, then such contents are very easy to compare.

MatousJobanek:
I have one more thought regarding the diffs & printing the whole object contents. In your implementation, the logic prints just the diff of either the conditions or spec or some other sub-part of the CR. However, when the test fails, it's sometimes very useful to know the whole context, which means not only the expected vs actual part of the CR, but also the rest of the CR:
* what was set in the spec
* in which states were the other conditions 
* which labels the CR contains
* the `managedFields` part to know which entity created/updated the CR for the last time
* etc...

So that being said - when the test fails, it would be nice to have something like:
```
Timeout was reached while waiting for the condition being available in UserSignup '12345'


ExpectedCondition:
----------------------------------------------
lastTransitionTime: ""
reason: Deactivated
status: "True"
type: Complete
----------------------------------------------

Actual UserSignup:
--------------------------------------------------------------------------------------------
apiVersion: toolchain.dev.openshift.com/v1alpha1
kind: UserSignup
metadata:
  annotations:
    toolchain.dev.openshift.com/activation-counter: "1"
    toolchain.dev.openshift.com/user-email: mjobanek@redhat.com
    toolchain.dev.openshift.com/verification-counter: "0"
  creationTimestamp: "2021-05-25T10:01:34Z"
  generation: 3
  labels:
    toolchain.dev.openshift.com/email-hash: 4517c516e47a37b2745ed5b12fdecaac
    toolchain.dev.openshift.com/state: deactivated
  managedFields:
    fieldsType: FieldsV1
...
bla bla bla
...
spec:
  familyName: Jobanek
  givenName: Matous
  states:
  - deactivated
  userid: "6583713"
  username: mjobanek@redhat.com
status:
  compliantUsername: mjobanek-2
  conditions:
  - lastTransitionTime: "2021-06-24T10:01:46Z"
    reason: NotificationCRCreated
    status: "True"
    type: Approved
...
bla bla bla
...
  - lastTransitionTime: "2021-06-24T10:01:46Z"
    reason: Deactivated
    status: "True"
    type: Complete
--------------------------------------------------------------------------------------------
```

sbryzak (commented):
PR looks good, however the proof will be in actually performing a test run.  I'll check this out today, I'm looking forward to seeing the new logs :)

sbryzak:
> PR looks good, however the proof will be in actually performing a test run. I'll check this out today, I'm looking forward to seeing the new logs :)

So it turns out that this PR needs to be merged with the changes from master before it will work... I did take a quick look at doing this but it seems that it's not a trivial merge, and likely requires Xavier's intimate knowledge of his changes to carry out successfully.

testsupport/wait/awaitility.go
```diff
@@ -147,6 +146,7 @@ func (a *Awaitility) WaitForToolchainClusterWithCondition(clusterType cluster.Ty
 // WaitForNamedToolchainClusterWithCondition waits until there is a ToolchainCluster with the given name
 // and with the given ClusterCondition (if it the condition is nil, then it skips this check)
 func (a *Awaitility) WaitForNamedToolchainClusterWithCondition(name string, condition *toolchainv1alpha1.ToolchainClusterCondition) (toolchainv1alpha1.ToolchainCluster, error) {
+	a.T.Logf("waiting for ToolchainCluster '%s' in namespace '%s'", name, a.Namespace)
```
ranakan19, line 149: would it be better to have the logger message indicating that it is waiting for the ToolchainCluster conditions?
```suggestion
	a.T.Logf("waiting for ToolchainCluster '%s' in namespace '%s' to have condition '%v'", name, a.Namespace, condition)
```

testsupport/wait/awaitility.go
```diff
@@ -537,7 +511,7 @@ func (a *Awaitility) Cleanup(objects ...client.Object) {
 						if deleted, err := a.verifyMurDeleted(isUserSignup, userSignup, false); !deleted || err != nil {
 							return false, err
 						}
-						a.T.Logf("%s: %s is deleted", kind, metaAccess.GetName())
+						fmt.Println("")
```
ranakan19, line 514:
```suggestion
						
```

testsupport/wait/diff.go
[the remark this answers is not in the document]
rajivnathan: I think from a maintenance perspective, it's easier to live with the extra noise than to have to keep ensuring we update this function for each type that has a timestamp. If it bothers us enough then we should find a more generic solution that doesn't increase the complexity/maintainability.

testsupport/wait/member.go
```diff
@@ -72,158 +73,197 @@ func (a *MemberAwaitility) WithRetryOptions(options ...RetryOption) *MemberAwait
...
 func UntilUserAccountHasConditions(conditions ...toolchainv1alpha1.Condition) UserAccountWaitCriterion {
-	return func(a *MemberAwaitility, ua *toolchainv1alpha1.UserAccount) bool {
-		if test.ConditionsMatch(ua.Status.Conditions, conditions...) {
-			a.T.Logf("status conditions match in UserAccount '%s`", ua.Name)
-			return true
-		}
-		a.T.Logf("waiting for status condition of UserSignup '%s'. Actual: '%+v'; Expected: '%+v'", ua.Name, ua.Status.Conditions, conditions)
-		return false
+	return UserAccountWaitCriterion{
+		Match: func(actual *toolchainv1alpha1.UserAccount) bool {
```
rajivnathan, line 136: Here too?

testsupport/wait/host.go
rajivnathan: Wouldn't the diff help when it's not immediately obvious? eg. If only the reason is different

rajivnathan (commented):
I took a quick looks and I really like the reduced output and the changes in general! The assertions that have Diffs are awesome and are definitely an improvement 👍  I think in some cases the Diffs are not being used but may be beneficial to make failures more obvious or just reduce the time it takes for someone to identify the differences.

testsupport/wait/diff.go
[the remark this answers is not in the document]
ranakan19: i agree with Matous, it would be an additional step for maintenance of CRS (not sure about the complexity though). 
I couldn't follow how reflect package adds to the cost of maintainability to compare the two options, could you please break it down for me? ( I tried looked at the reflect package, but maybe i'm missing out the obvious)

testsupport/wait/host.go
```diff
@@ -162,138 +162,200 @@ func (a *HostAwaitility) UpdateUserSignupSpec(userSignupName string, modifyUserS
 }
 
 // MasterUserRecordWaitCriterion checks if a MasterUserRecord meets the given condition
```
ranakan19, line 164: the comment probably needs to be changed since MasterUserRecordWaitCriterion is not a function now?

testsupport/wait/member.go
```diff
@@ -72,158 +73,197 @@ func (a *MemberAwaitility) WithRetryOptions(options ...RetryOption) *MemberAwait
...
-		expectedSpec.NSTemplateSet = toolchainv1alpha1.NSTemplateSetSpec{}
-		return reflect.DeepEqual(userAccount.Spec, *expectedSpec)
+	return UserAccountWaitCriterion{
+		Match: func(actual *toolchainv1alpha1.UserAccount) bool {
+			userAccount := actual.DeepCopy()
+			userAccount.Spec.NSTemplateSet = toolchainv1alpha1.NSTemplateSetSpec{}
+			expectedSpec := expected.DeepCopy()
+			expectedSpec.NSTemplateSet = toolchainv1alpha1.NSTemplateSetSpec{}
+			return reflect.DeepEqual(userAccount.Spec, *expectedSpec)
+		},
```
ranakan19, line 112: adding to rajiv's finding, missing a diff here too?

ranakan19:
The changes look great, down by over 15k lines is simply awesome!

testsupport/wait/awaitility.go
xcoulon: that's a good point, thanks!

testsupport/wait/diff.go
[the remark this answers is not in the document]
xcoulon: @ranakan19 with the `reflect` package, we could look at the structure of the given object and look for all fields of type `metav1.Time`. But we also need to look at all nested structs, too (can be `[]interface{}`, `toolchainv1alpha1.UserAccountStatus`, etc.) and that's where the code gets more complicated, I believe.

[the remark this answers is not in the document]
xcoulon: ok, so I'll remove this `ResetTimeFields` for now, and we'll see as time goes by if the diffs on the timestamp causes confusion or not.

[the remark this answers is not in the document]
xcoulon: done in 5e252dd

testsupport/wait/host.go
xcoulon: good point! Addressed in e18eb4b for all similar structs (among other changes to improve readability of code)

testsupport/wait/host.go
xcoulon: > Wouldn't the diff help when it's not immediately obvious? eg. If only the reason is different

that's the purpose of the `Diff` func, indeed. 
I suggest with mark this conversation as "resolved" for now

xcoulon: ah, I see what you meant now! 🤦‍♂️ 
Let me add a diff to compare the structs and spot the differences more easily ;)

xcoulon: done in 9085b05

testsupport/wait/member.go
xcoulon: done in 9085b05

testsupport/wait/member.go
```diff
@@ -72,158 +73,197 @@ func (a *MemberAwaitility) WithRetryOptions(options ...RetryOption) *MemberAwait
...
-			mur.Spec.UserAccounts[0].Spec.UserAccountSpecBase,
-			ua.Spec.UserID,
-			mur.Spec.UserID,
-			ua.Spec.Disabled,
-			mur.Spec.Disabled)
-		return ua.Spec.UserID == mur.Spec.UserID &&
-			ua.Spec.Disabled == mur.Spec.Disabled &&
-			reflect.DeepEqual(ua.Spec.UserAccountSpecBase, mur.Spec.UserAccounts[0].Spec.UserAccountSpecBase)
+	return UserAccountWaitCriterion{
+		Match: func(actual *toolchainv1alpha1.UserAccount) bool {
```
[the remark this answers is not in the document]
xcoulon: fixed in 95d3382

testsupport/wait/member.go
xcoulon: done in 3b87339

testsupport/wait/member.go
```diff
@@ -1140,18 +1259,15 @@ func (a *MemberAwaitility) waitForAutoscalingBufferDeployment() {
 }
 
 // WaitForExpectedNumberOfResources waits until the number of resources matches the expected count
-func (a *MemberAwaitility) WaitForExpectedNumberOfResources(kind string, expected int, list func() (int, error)) error {
-	err := wait.Poll(a.RetryInterval, a.Timeout, func() (done bool, err error) {
+func (a *MemberAwaitility) WaitForExpectedNumberOfResources(expected int, list func() (int, error)) error {
+	return wait.Poll(a.RetryInterval, a.Timeout, func() (done bool, err error) {
```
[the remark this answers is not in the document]
xcoulon: done in 4b61008

xcoulon:
@MatousJobanek @sbryzak @ranakan19 @rajivnathan Hopefully I addressed all comments. Could you please give this PR another look?

rajivnathan (approved):
Great work! 🥇

xcoulon:
> I have one more thought regarding the diffs & printing the whole object contents. In your implementation, the logic prints just the diff of either the conditions or spec or some other sub-part of the CR. However, when the test fails, it's sometimes very useful to know the whole context, which means not only the expected vs actual part of the CR, but also the rest of the CR:
> 
> * what was set in the spec
> * in which states were the other conditions
> * which labels the CR contains
> * the `managedFields` part to know which entity created/updated the CR for the last time
> * etc...
> 
> So that being said - when the test fails, it would be nice to have something like:
> 
> ...

with 82a3eec we can have the following output:

```
member.go:1076: waiting for MemberStatus 'toolchain-member-status' to match criteria
    member.go:1098: failed to find MemberStatus with matching criteria:
        ----
        actual:
        (*v1alpha1.MemberStatus)({
         TypeMeta: (v1.TypeMeta) &TypeMeta{Kind:,APIVersion:,},
         ObjectMeta: (v1.ObjectMeta) &ObjectMeta{Name:toolchain-member-status,GenerateName:,Namespace:toolchain-member-operator,SelfLink:,UID:5bb93529-653d-4599-a495-68f851ba70f8,ResourceVersion:145470,Generation:1,CreationTimestamp:2021-08-30 12:21:06 +0200 CEST,DeletionTimestamp:<nil>,DeletionGracePeriodSeconds:nil,Labels:map[string]string{},Annotations:map[string]string{toolchain.dev.openshift.com/last-applied-configuration: {"metadata":{"name":"toolchain-member-status","namespace":"toolchain-member-operator","creationTimestamp":null},"spec":{},"status":{"resourceUsage":{}}},},OwnerReferences:[]OwnerReference{},Finalizers:[],ClusterName:,ManagedFields:[]ManagedFieldsEntry{ManagedFieldsEntry{Manager:member-operator,Operation:Update,APIVersion:toolchain.dev.openshift.com/v1alpha1,Time:2021-08-30 12:21:56 +0200 CEST,FieldsType:FieldsV1,FieldsV1:{"f:metadata":{"f:annotations":{".":{},"f:toolchain.dev.openshift.com/last-applied-configuration":{}}},"f:spec":{},"f:status":{".":{},"f:che":{".":{},"f:conditions":{".":{},"k:{\"type\":\"Ready\"}":{".":{},"f:lastTransitionTime":{},"f:lastUpdatedTime":{},"f:reason":{},"f:status":{},"f:type":{}}}},"f:conditions":{".":{},"k:{\"type\":\"Ready\"}":{".":{},"f:lastTransitionTime":{},"f:lastUpdatedTime":{},"f:reason":{},"f:status":{},"f:type":{}}},"f:host":{".":{},"f:conditions":{".":{},"k:{\"type\":\"Ready\"}":{".":{},"f:lastTransitionTime":{},"f:lastUpdatedTime":{},"f:reason":{},"f:status":{},"f:type":{}}}},"f:memberOperator":{".":{},"f:buildTimestamp":{},"f:conditions":{".":{},"k:{\"type\":\"Ready\"}":{".":{},"f:lastTransitionTime":{},"f:lastUpdatedTime":{},"f:reason":{},"f:status":{},"f:type":{}}},"f:deploymentName":{},"f:revision":{},"f:version":{}},"f:resourceUsage":{".":{},"f:memoryUsagePerNodeRole":{".":{},"f:master":{},"f:worker":{}}},"f:routes":{".":{},"f:conditions":{".":{},"k:{\"type\":\"Ready\"}":{".":{},"f:lastTransitionTime":{},"f:lastUpdatedTime":{},"f:reason":{},"f:status":{},"f:type":{}}},"f:consoleURL":{}}}},},},},
         Spec: (v1alpha1.MemberStatusSpec) {
         },
         Status: (v1alpha1.MemberStatusStatus) {
          Che: (*v1alpha1.CheStatus)({
           Conditions: ([]v1alpha1.Condition) (len=1) {
            (v1alpha1.Condition) {
             Type: (v1alpha1.ConditionType) (len=5) "Ready",
             Status: (v1.ConditionStatus) (len=4) "True",
             LastTransitionTime: (v1.Time) 2021-08-30 12:40:27 +0200 CEST,
             Reason: (string) (len=25) "CheUserDeletionNotEnabled",
             Message: (string) "",
             LastUpdatedTime: (*v1.Time)(2021-08-30 12:40:27 +0200 CEST)
            }
           }
          }),
          MemberOperator: (*v1alpha1.MemberOperatorStatus)({
           Version: (string) (len=5) "0.0.1",
           Revision: (string) (len=7) "unknown",
           BuildTimestamp: (string) (len=7) "unknown",
           DeploymentName: (string) (len=34) "member-operator-controller-manager",
           Conditions: ([]v1alpha1.Condition) (len=1) {
            (v1alpha1.Condition) {
             Type: (v1alpha1.ConditionType) (len=5) "Ready",
             Status: (v1.ConditionStatus) (len=4) "True",
             LastTransitionTime: (v1.Time) 2021-08-30 12:40:27 +0200 CEST,
             Reason: (string) (len=15) "DeploymentReady",
             Message: (string) "",
             LastUpdatedTime: (*v1.Time)(2021-08-30 12:40:27 +0200 CEST)
            }
           }
          }),
          HostConnection: (*v1alpha1.ToolchainClusterStatus)(<nil>),
          Host: (*v1alpha1.HostStatus)({
           Conditions: ([]v1alpha1.Condition) (len=1) {
            (v1alpha1.Condition) {
             Type: (v1alpha1.ConditionType) (len=5) "Ready",
             Status: (v1.ConditionStatus) (len=4) "True",
             LastTransitionTime: (v1.Time) 2021-08-30 12:40:27 +0200 CEST,
             Reason: (string) (len=19) "HostConnectionReady",
             Message: (string) "",
             LastUpdatedTime: (*v1.Time)(2021-08-30 12:40:27 +0200 CEST)
            }
           }
          }),
          Conditions: ([]v1alpha1.Condition) (len=1) {
           (v1alpha1.Condition) {
            Type: (v1alpha1.ConditionType) (len=5) "Ready",
            Status: (v1.ConditionStatus) (len=4) "True",
            LastTransitionTime: (v1.Time) 2021-08-30 12:40:27 +0200 CEST,
            Reason: (string) (len=18) "AllComponentsReady",
            Message: (string) "",
            LastUpdatedTime: (*v1.Time)(2021-08-30 12:40:27 +0200 CEST)
           }
          },
          ResourceUsage: (v1alpha1.ResourceUsage) {
           MemoryUsagePerNodeRole: (map[string]int) (len=2) {
            (string) (len=6) "master": (int) 49,
            (string) (len=6) "worker": (int) 27
           }
          },
          Routes: (*v1alpha1.Routes)({
           ConsoleURL: (string) (len=82) "https://console-openshift-console.apps.xcoulon-20210830a.devcluster.openshift.com/",
           CheDashboardURL: (string) "",
           Conditions: ([]v1alpha1.Condition) (len=1) {
            (v1alpha1.Condition) {
             Type: (v1alpha1.ConditionType) (len=5) "Ready",
             Status: (v1.ConditionStatus) (len=4) "True",
             LastTransitionTime: (v1.Time) 2021-08-30 12:40:27 +0200 CEST,
             Reason: (string) (len=15) "RoutesAvailable",
             Message: (string) "",
             LastUpdatedTime: (*v1.Time)(2021-08-30 12:40:27 +0200 CEST)
            }
           }
          })
         }
        })
        
        ----
        diffs:
        expected MemberStatus route for Console to be 'http://example.com' with condition '(v1alpha1.Condition) {
         Type: (v1alpha1.ConditionType) (len=5) "Ready",
         Status: (v1.ConditionStatus) (len=4) "True",
         LastTransitionTime: (v1.Time) 0001-01-01 00:00:00 +0000 UTC,
         Reason: (string) (len=15) "RoutesAvailable",
         Message: (string) "",
         LastUpdatedTime: (*v1.Time)(<nil>)
        }
        ', but it was: (*v1alpha1.Routes)({
         ConsoleURL: (string) (len=82) "https://console-openshift-console.apps.xcoulon-20210830a.devcluster.openshift.com/",
         CheDashboardURL: (string) "",
         Conditions: ([]v1alpha1.Condition) (len=1) {
          (v1alpha1.Condition) {
           Type: (v1alpha1.ConditionType) (len=5) "Ready",
           Status: (v1.ConditionStatus) (len=4) "True",
           LastTransitionTime: (v1.Time) 2021-08-30 12:40:27 +0200 CEST,
           Reason: (string) (len=15) "RoutesAvailable",
           Message: (string) "",
           LastUpdatedTime: (*v1.Time)(2021-08-30 12:40:27 +0200 CEST)
          }
         }
        })
```
which is close to your suggestion, even though the dump is the Go struct and not pure YAML (but it's still readable IMO)

MatousJobanek:
Ok, but is it really so hard to print out the pure YAML? The go struct is pretty hard to digest and process. 
For example, the `ObjectMeta` is put into a single very long line.

xcoulon:
with 00659fe, here's a example when the MemberStatus console URL does not match the expectation:

```
member.go:1085: waiting for MemberStatus 'toolchain-member-status' to match criteria
    member.go:1107: failed to find MemberStatus with matching criteria:
        ----
        actual:
        metadata:
          annotations:
            toolchain.dev.openshift.com/last-applied-configuration: '{"metadata":{"name":"toolchain-member-status","namespace":"toolchain-member-operator","creationTimestamp":null},"spec":{},"status":{"resourceUsage":{}}}'
          creationTimestamp: "2021-08-30T10:21:06Z"
          generation: 1
          managedFields:
            fieldsType: FieldsV1
            fieldsV1:
              f:metadata:
                f:annotations:
                  .: {}
                  f:toolchain.dev.openshift.com/last-applied-configuration: {}
              f:spec: {}
              f:status:
                .: {}
                f:che:
                  .: {}
                  f:conditions:
                    .: {}
                    k:{"type":"Ready"}:
                      .: {}
                      f:lastTransitionTime: {}
                      f:lastUpdatedTime: {}
                      f:reason: {}
                      f:status: {}
                      f:type: {}
                f:conditions:
                  .: {}
                  k:{"type":"Ready"}:
                    .: {}
                    f:lastTransitionTime: {}
                    f:lastUpdatedTime: {}
                    f:reason: {}
                    f:status: {}
                    f:type: {}
                f:host:
                  .: {}
                  f:conditions:
                    .: {}
                    k:{"type":"Ready"}:
                      .: {}
                      f:lastTransitionTime: {}
                      f:lastUpdatedTime: {}
                      f:reason: {}
                      f:status: {}
                      f:type: {}
                f:memberOperator:
                  .: {}
                  f:buildTimestamp: {}
                  f:conditions:
                    .: {}
                    k:{"type":"Ready"}:
                      .: {}
                      f:lastTransitionTime: {}
                      f:lastUpdatedTime: {}
                      f:reason: {}
                      f:status: {}
                      f:type: {}
                  f:deploymentName: {}
                  f:revision: {}
                  f:version: {}
                f:resourceUsage:
                  .: {}
                  f:memoryUsagePerNodeRole:
                    .: {}
                    f:master: {}
                    f:worker: {}
                f:routes:
                  .: {}
                  f:conditions:
                    .: {}
                    k:{"type":"Ready"}:
                      .: {}
                      f:lastTransitionTime: {}
                      f:lastUpdatedTime: {}
                      f:reason: {}
                      f:status: {}
                      f:type: {}
                  f:consoleURL: {}
            manager: member-operator
            operation: Update
            time: "2021-08-30T10:21:56Z"
          name: toolchain-member-status
          namespace: toolchain-member-operator
          resourceVersion: "252389"
          uid: 5bb93529-653d-4599-a495-68f851ba70f8
        spec: {}
        status:
          che:
            conditions:
            - lastTransitionTime: "2021-08-30T12:39:56Z"
              lastUpdatedTime: "2021-08-30T12:39:56Z"
              reason: CheUserDeletionNotEnabled
              status: "True"
              type: Ready
          conditions:
          - lastTransitionTime: "2021-08-30T12:39:56Z"
            lastUpdatedTime: "2021-08-30T12:39:56Z"
            reason: AllComponentsReady
            status: "True"
            type: Ready
          host:
            conditions:
            - lastTransitionTime: "2021-08-30T12:39:56Z"
              lastUpdatedTime: "2021-08-30T12:39:56Z"
              reason: HostConnectionReady
              status: "True"
              type: Ready
          memberOperator:
            buildTimestamp: unknown
            conditions:
            - lastTransitionTime: "2021-08-30T12:39:56Z"
              lastUpdatedTime: "2021-08-30T12:39:56Z"
              reason: DeploymentReady
              status: "True"
              type: Ready
            deploymentName: member-operator-controller-manager
            revision: unknown
            version: 0.0.1
          resourceUsage:
            memoryUsagePerNodeRole:
              master: 52
              worker: 27
          routes:
            conditions:
            - lastTransitionTime: "2021-08-30T12:39:56Z"
              lastUpdatedTime: "2021-08-30T12:39:56Z"
              reason: RoutesAvailable
              status: "True"
              type: Ready
            consoleURL: https://console-openshift-console.apps.xcoulon-20210830a.devcluster.openshift.com/
        
        ----
        diffs:
        expected MemberStatus route for Console to be 'http://example.com' with condition
        lastTransitionTime: null
        reason: RoutesAvailable
        status: "True"
        type: Ready
        
        but it was: 
        conditions:
        - lastTransitionTime: "2021-08-30T12:39:56Z"
          lastUpdatedTime: "2021-08-30T12:39:56Z"
          reason: RoutesAvailable
          status: "True"
          type: Ready
        consoleURL: https://console-openshift-console.apps.xcoulon-20210830a.devcluster.openshift.com/
```

MatousJobanek:
awesome, this looks much better :star_struck:

MatousJobanek:
@xcoulon I noticed that before you re-ran the e2e test it failed with:
```
 === CONT  TestUserManagement
    host.go:497: waiting until MasterUserRecord 'usertodeactivate2' in namespace 'toolchain-host-30133531' is deleted
    host.go:806: waiting for notifications to match criteria for user 'usertodeactivate2'
    host.go:825: failed to find notifications with matching criteria:
        actual:
        null
        
        ----
        diffs:
        
=== CONT  TestUserManagement/TestUserDeactivation
    base_user_test.go:68: 
        	Error Trace:	base_user_test.go:68
        	            				user_management_test.go:75
        	Test:       	TestUserManagement/TestUserDeactivation 
```
so there are two questions:
1. Why did it fail?
2. Shouldn't the message be improved?

xcoulon:
> so there are two questions:
> 
> 1. Why did it fail?
> 2. Shouldn't the message be improved?

I'll take a look

xcoulon:
> > so there are two questions:
> > 
> > 1. Why did it fail?
> > 2. Shouldn't the message be improved?
> 
> I'll take a look

I've push a change to improve the error message: 9745d45
but the test passed on my temp cluster, so I assume it's another flaky test?

xcoulon:
@MatousJobanek I updated the PR in a74bb64 to list the relevant resource in the current namespace (by "current", I mean the host-operator or the member-operator, depending on the assertion that failed).

Also, I introduced a small change in the "delete useraccount and expect recreation" test to make it fail, so we can have an example of output in case of test failure.

PR #352, opened 2021-07-29
Lookup subscription using prefix

rajivnathan:
It looks like the subscriptions now have an identifier appended so the setup tool needs to update to look for a prefix instead of an exact match


xcoulon (approved):

rajivnathan:
this is a change to the setup tool only so no need to wait for the e2e test to complete

ranakan19 (commented):
LGTM

PR #361, opened 2021-08-06
Update toolchainconfig for dev and e2e environments

rajivnathan:
Related PRs: 
- https://github.com/codeready-toolchain/registration-service/pull/205
- https://github.com/codeready-toolchain/toolchain-common/pull/204

Related Issue: https://issues.redhat.com/browse/CRT-1122

Summary of changes:
- Updates the toolchainconfig resource created for the dev and e2e environments
- Minor updates with updated toolchain-common dependency


deploy/host-operator/dev/toolchainconfig.yaml
```diff
@@ -6,6 +6,10 @@ spec:
   host:
     automaticApproval:
       enabled: true
+    deactivation:
+      deactivationDomainsExcluded: '@redhat.com'
+    registrationService:
+      environment: 'e2e-tests'
```
alexeykazakov, line 12: Should be "dev"

rajivnathan: Thanks for catching that. Fixed: e77ad7b91dee73517ee703a8773f849d51aee1f4

alexeykazakov (approved):

deploy/host-operator/e2e-tests/toolchainconfig.yaml
```diff
@@ -3,11 +3,34 @@ kind: ToolchainConfig
...
+        excludedEmailDomains: 'redhat.com,acme.com'
+        secret:
+          ref: 'host-operator-secret'
+          twilioAccountSID: 'twilio.account.sid'
+          twilioAuthToken: 'twilio.auth.token'
+          twilioFromNumber: 'twilio.from_number'
+    tiers:
+      durationBeforeChangeTierRequestDeletion: '5s'
+    toolchainStatus:
+      toolchainStatusRefreshTime: '1s'
```
MatousJobanek, line 28: I see that you don't specify the number of replicas of the reg-service as we do in `e2e.yaml` & `dev.yaml` config files. Wouldn't it be better to drop it from the file as well to make the config consistent? In that case you would also need to modify the e2e test: https://github.com/codeready-toolchain/toolchain-e2e/blob/master/testsupport/init.go#L53

deploy/host-operator/dev/toolchainconfig.yaml
```diff
@@ -6,6 +6,10 @@ spec:
   host:
     automaticApproval:
       enabled: true
+    deactivation:
+      deactivationDomainsExcluded: '@redhat.com'
+    registrationService:
+      environment: 'dev'
```
MatousJobanek, line 12: I guess that you should specify also the `analytics` section:
```
      analytics:
        segmentWriteKey: 'test segment write key'
        woopraDomain: 'test woopra domain'
      environment: 'dev'
```
as well as the secret 
```
        secret:
          ref: 'host-operator-secret'
          twilioAccountSID: 'twilio.account.sid'
          twilioAuthToken: 'twilio.auth.token'
          twilioFromNumber: 'twilio.from_number'
```

deploy/host-operator/e2e-tests/toolchainconfig.yaml
rajivnathan: Makes sense, I've added replicas in https://github.com/codeready-toolchain/api/pull/253 and I'll update this to include replicas.

deploy/host-operator/dev/toolchainconfig.yaml
rajivnathan: I believe the analytics config are only needed in the e2e environment? But I suppose the secret could be useful so I can add that.

alexeykazakov: we use analytics in prod

rajivnathan: Added verification secret in bd095f6afb661d77d0eb790ae56cd9555fa3ff62

rajivnathan: @alexeykazakov yeah, I meant for the toolchainconfig.yaml in the toolchain-e2e repo which is only for dev and e2e environments. The ToolchainConfig for prod in sandbox-sre have the analytics set correctly so we should be good.

deploy/host-operator/e2e-tests/toolchainconfig.yaml
rajivnathan: Actually, I'll add it in a separate PR since it won't be used now anyway. It'll be used in the host operator config migration PR.

rajivnathan: Added in https://github.com/codeready-toolchain/toolchain-e2e/pull/368

PR #364, opened 2021-08-09
Quick fix for a couple flaky tests

rajivnathan:
A couple of the tests don't specify a target cluster when creating a new UserSignup so the UserAccount and other member cluster resources could be on member1 or member2 but the tests only perform assertions on a particular member cluster which leads to test flakiness.

This PR sets the target cluster to the cluster used for the assertions but we can probably make this a little more intuitive and reduce/eliminate future test flakiness related to this type of problem.

### DIFF
--- a/testsupport/wait/member.go
+++ b/testsupport/wait/member.go
@@ -120,10 +120,10 @@ func UntilUserAccountMatchesMur(hostAwaitility *HostAwaitility) UserAccountWaitC
 func UntilUserAccountHasConditions(conditions ...toolchainv1alpha1.Condition) UserAccountWaitCriterion {
 	return func(a *MemberAwaitility, ua *toolchainv1alpha1.UserAccount) bool {
 		if test.ConditionsMatch(ua.Status.Conditions, conditions...) {
-			a.T.Logf("status conditions match in UserAccount '%s`", ua.Name)
+			a.T.Logf("status conditions match in UserAccount '%s' in namespace '%s'", ua.Name, ua.Namespace)
 			return true
 		}
-		a.T.Logf("waiting for status condition of UserSignup '%s'. Actual: '%+v'; Expected: '%+v'", ua.Name, ua.Status.Conditions, conditions)
+		a.T.Logf("waiting for status condition of UserSignup '%s' in namespace '%s'. Actual: '%+v'; Expected: '%+v'", ua.Name, ua.Namespace, ua.Status.Conditions, conditions)
 		return false
 	}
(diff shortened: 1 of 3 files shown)


ranakan19 (approved):
LGTM

ranakan19:
Maybe we can add the namespace name in the log message to make it clear? i.e https://github.com/codeready-toolchain/toolchain-e2e/blob/master/testsupport/wait/member.go#L240 here add the namespace it is looking in as well? I can do that if you think it makes sense

MatousJobanek (approved):
:+1:

MatousJobanek:
> Maybe we can add the namespace name in the log message to make it clear? i.e https://github.com/codeready-toolchain/toolchain-e2e/blob/master/testsupport/wait/member.go#L240 here add the namespace it is looking in as well? I can do that if you think it makes sense

yes, it would make sense - and do the same also for UserAccount

rajivnathan:
Added namespace in the log messages for nstemplateset and useraccount

testsupport/wait/member.go
```diff
@@ -120,10 +120,10 @@ func UntilUserAccountMatchesMur(hostAwaitility *HostAwaitility) UserAccountWaitC
 func UntilUserAccountHasConditions(conditions ...toolchainv1alpha1.Condition) UserAccountWaitCriterion {
 	return func(a *MemberAwaitility, ua *toolchainv1alpha1.UserAccount) bool {
 		if test.ConditionsMatch(ua.Status.Conditions, conditions...) {
-			a.T.Logf("status conditions match in UserAccount '%s`", ua.Name)
+			a.T.Logf("status conditions match in UserAccount '%s' in namespace '%s'", ua.Name, ua.Namespace)
 			return true
 		}
-		a.T.Logf("waiting for status condition of UserSignup '%s'. Actual: '%+v'; Expected: '%+v'", ua.Name, ua.Status.Conditions, conditions)
+		a.T.Logf("waiting for status condition of UserSignup '%s' in namespace '%s'. Actual: '%+v'; Expected: '%+v'", ua.Name, ua.Namespace, ua.Status.Conditions, conditions)
```
ranakan19, line 126:
```suggestion
		a.T.Logf("waiting for status condition of UserAccount '%s' in namespace '%s'. Actual: '%+v'; Expected: '%+v'", ua.Name, ua.Namespace, ua.Status.Conditions, conditions)
```

alexeykazakov (approved):
Good catch!

PR #370, opened 2021-08-17
Refactors tests to use SignupRequest API instead of CreateAndApproveSignup

sbryzak:
Fixes https://issues.redhat.com/projects/CRT/issues/CRT-1161


alexeykazakov (approved):
Looks good!

test/e2e/nstemplatetier_test.go
```diff
@@ -30,11 +30,16 @@ const (
...
 	// given
-	hostAwait, memberAwait, _ := WaitForDeployments(t)
+	hostAwait, memberAwait, memberAwait2 := WaitForDeployments(t)
 
 	// Create and approve "testingtiers" signups
 	testingTiersName := "testingtiers"
-	testingtiers := CreateAndApproveSignup(t, hostAwait, testingTiersName, memberAwait.ClusterName)
+	testingtiers, _ := NewSignupRequest(t, hostAwait, memberAwait, memberAwait2).
+		Username(testingTiersName).
+		ManuallyApprove().
```
MatousJobanek, line 39: originally, the target cluster was set, so we should set it here too
```suggestion
		ManuallyApprove().
		TargetCluster(memberAwait).
```

test/e2e/user_management_test.go
```diff
@@ -495,16 +495,17 @@ func (s *userManagementTestSuite) TestUserDisabled() {
 	s.hostAwait.UpdateToolchainConfig(testconfig.AutomaticApproval().Enabled(false))
 
 	// Create UserSignup
-	userSignup := CreateAndApproveSignup(s.T(), s.hostAwait, "janedoe", s.memberAwait.ClusterName)
+	userSignup, mur := s.newSignupRequest().
+		Username("janedoe").
+		EnsureMUR().
+		ManuallyApprove().
```
MatousJobanek, line 501:
```suggestion
		ManuallyApprove().
		TargetCluster(memberAwait).
```

test/metrics/metrics_test.go
```diff
@@ -33,7 +33,16 @@ func TestMetricsWhenUsersDeactivated(t *testing.T) {
...
 		username := fmt.Sprintf("user-%04d", i)
-		usersignups[username] = CreateAndApproveSignup(t, hostAwait, username, member2Await.ClusterName, WithEmail(username+"@redhat.com")) // testing on internal accounts
+
+		// Create UserSignup
+		usersignups[username], _ = NewSignupRequest(t, hostAwait, memberAwait, member2Await).
+			Username(username).
+			Email(username + "@redhat.com").
+			ManuallyApprove().
+			TargetCluster(member2Await).
+			RequireConditions(ConditionSet(Default(), ApprovedByAdmin())...).
```
MatousJobanek, line 43: May I know why don't you call `EnsureMUR()`? In this particular case, it could have a negative impact on the tests and could cause test flakiness
```suggestion
			RequireConditions(ConditionSet(Default(), ApprovedByAdmin())...).
			EnsureMUR().
```

MatousJobanek (changes requested):
I guess that you can remove the `CreateAndApproveSignup` function as well as the other functions that are not used anymore https://github.com/codeready-toolchain/toolchain-e2e/blob/master/testsupport/user_setup.go#L50-L132

BTW, there is still this `CreateMultipleSignups` function https://github.com/codeready-toolchain/toolchain-e2e/blob/master/testsupport/user_setup.go#L25-L48 which still doesn't use the SignupRequest API yet - shouldn't that be changed too?

In addition to what I have written - could you please check the cases when you don't call `EnsureMUR()`? Why did you skip that?

test/metrics/metrics_test.go
sbryzak: I still have a number of local changes that I haven't pushed yet due to continued test failures, pretty sure I've added `EnsureMUR()` here already.

test/e2e/nstemplatetier_test.go
```diff
@@ -30,11 +30,16 @@ const (
...
+		RequireConditions(ConditionSet(Default(), ApprovedByAdmin())...).
+		Execute().
+		Resources()
```
MatousJobanek, line 42: you still don't set the target cluster here now wait for the MUR to be fully provisioned - are you sure that it's fine?

MatousJobanek (commented):
Thanks a lot for addressing my comments. :pray: 
I believe that there is still one case when you should wait for the MUR to be fully provisioned

test/e2e/nstemplatetier_test.go
sbryzak: You're right, I should be setting the target cluster here and also calling EnsureMUR.  Fixed.

PR #372, opened 2021-08-19
fix: expect the tier label is correctly set

MatousJobanek:
When verifying the cluster-scoped resources it also checks if the resource has the corresponding tier label set.

paired with: https://github.com/codeready-toolchain/member-operator/pull/293


rajivnathan (approved):

alexeykazakov (approved):

testsupport/tiers/checks.go
```diff
@@ -614,13 +614,13 @@ func clusterResourceQuotaBuildConfig() clusterObjectsCheckCreator {
 
 func clusterResourceQuotaSecrets() clusterObjectsCheckCreator {
 	return func(tierName string) clusterObjectsCheck {
```
xcoulon, line 616: if I'm not mistaken, `tierName` becomes unused but for some reason, the linter did not catch that 😕

MatousJobanek: yeah, this is probably pretty hard for the linter to check that. 
At that time when I started working on it, the var was used to differentiate between `team` tier and the other tiers - `team` had different resource quotas. So I was thinking about keeping it and mainly renaming as it refers to the original tier, not the altered one.
But I'll most likely drop it to avoid any confusions.

MatousJobanek: dropped bef1d40

PR #373, opened 2021-08-20
chore: use default health check period in dev

MatousJobanek:


alexeykazakov (approved):

xcoulon (approved):

rajivnathan:
@alexeykazakov Since this is mainly cosmetic and only impacts the dev environment right now maybe we should just keep the config as is and ensure the fix for https://issues.redhat.com/browse/CRT-1217 resolves the issue?

alexeykazakov:
@rajivnathan I'm fine with both options (merge it or keep it as is if it helps you to verify the fix for 1217). So I will leave it to you :) Feel free to merge or not to merge :)

MatousJobanek:
@rajivnathan so do you want to keep it as it is to verify the fix for CRT-1217? I'm fine with both options too :wink:

rajivnathan:
@MatousJobanek I was thinking about making the timeout and period update automatically as the configuration updates but it looks like it would require more changes than I thought and would add more unnecessary complexity so I think it's better to just change the default in dev (merge this PR) and not change anything else. It's not like we'll be changing the health check period or timeout often anyway. So I'll merge this PR and close https://issues.redhat.com/browse/CRT-1217 with the same explanation.

PR #376, opened 2021-08-24
Adjust idler setup

rajivnathan:
Summary of changes

- The default number of total users is changed to 2000 (previously 3000)
- Updated default template to only create deployments with 0 replicas so you can expect there will be no pods running by default, only pods created via custom templates should be expected
- The Idler timeout is configurable via a flag. `--idler-timeout` or `-i` So you can expect pods and replicas will still be killed/scaled down in 15 seconds by default but the flag can be used to customize it
- The templates will be split into two categories:
  a. Default - the default template provided by the setup tool (setup/resources/user-workloads.yaml)
  b. Custom - custom templates provided by the user via the `--template` flag
- Removed the `--active` parameter. It is replaced by `--default` which specifies the number of users to have the default template applied and `--custom` which specifies the number of users to have the custom templates applied

Example command:
`go run setup/main.go --users 25 --default 10 --custom 5 --template=<path to template> --idler-timeout=12h`


xcoulon (approved):
"new" default will be 12h instead of 15s, but I guess it's better for other operator teams (ie, aligns with the default behaviour), and we can always specify a lower value our own usage

setup/cmd/root.go
```diff
@@ -70,6 +71,7 @@ func Execute() {
 	cmd.Flags().BoolVar(&skipCSVGen, "skip-csvgen", false, "if an all-namespaces operator should be installed to generate a CSV resource in each namespace")
 	cmd.Flags().BoolVar(&skipDefaultTemplate, "skip-default-template", false, "skip the setup/resources/user-workloads.yaml template file when creating resources")
 	cmd.Flags().IntVar(&operatorsLimit, "operators-limit", len(operators.Templates), "can be specified to limit the number of additional operators to install (by default all operators are installed to simulate cluster load in production)")
+	cmd.Flags().StringVarP(&idlerTimeout, "idler-timeout", "i", "12h", "overrides the default idler timeout")
```
alexeykazakov, line 74: Let's maybe keep it 15s by default and instruct the serverless folks how to use this param to set it to 12h?

alexeykazakov (approved):

setup/cmd/root.go
rajivnathan: Done https://github.com/codeready-toolchain/toolchain-e2e/pull/376/commits/21e5abd0c1f02f91f6961c32808a39bb48ef2f12

alexeykazakov (approved):

PR #377, opened 2021-08-26
feat: add makefile goal and GH Action to run unit tests

xcoulon:
we have unit tests on utility funcs declared in the `testsupport`
package, and we should run these tests when a PR is submitted
    
Fixes https://issues.redhat.com/browse/CRT-1192
    
### DIFF
--- a/make/test.mk
+++ b/make/test.mk
@@ -222,3 +222,14 @@ display-eval:
 	@echo 'export REGISTRATION_SERVICE_NS=$$HOST_NS'
 	@echo '# Run this command to configure your shell:'
 	@echo '# eval $$(make display-eval)'
+
+
+###########################################################
+#
+# Unit Tests (to verify the assertions and other utilities
+# in the `testsupport` package)
+#
+###########################################################
+.PHONY: test-unit
+test-unit:
+	@go test github.com/codeready-toolchain/toolchain-e2e/testsupport/... -failfast
--- a/make/test.mk
+++ b/make/test.mk
@@ -231,7 +231,16 @@ display-eval:
 #
 ###########################################################
 
+# Output directory for coverage information
+COV_DIR = $(OUT_DIR)/coverage
+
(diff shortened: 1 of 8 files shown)


make/test.mk
```diff
@@ -222,3 +222,25 @@ display-eval:
...
+.PHONY: test
+## Run the unit tests in the 'testsupport/...' packages
+test:
+	@go test github.com/codeready-toolchain/toolchain-e2e/testsupport/... -failfast
+
+.PHONY: test-with-coverage
+## Run the unit tests in the 'testsupport/...' packages (with coverage)
+test-with-coverage:
+	@-mkdir -p $(COV_DIR)
+	@go test -coverprofile=$(COV_DIR)/coverage.txt -covermode=atomic github.com/codeready-toolchain/toolchain-e2e/testsupport/...
```
MatousJobanek, line 246: Just out of curiosity - do we really need the code coverage in e2e-tests repo? I believe that we don't care in this case much, right?

xcoulon: > I believe that we don't care in this case much, right?

why?

MatousJobanek: because the code coverage will be very low - there is only one test file for now and there is no reason for covering the other functions/test helpers. It's like writing a test for a test - that could never end :-)

xcoulon: then feel free to remove it, if you really think it's useless.

MatousJobanek: It was just my opinion and more or less also a question of what you think about that.
I'm also afraid of publishing the code coverage report to the toolchain-e2e PRs as they relate only to unit tests inside toolchain-e2e repo and not to the e2e tests as someone could expect - it could lead to confusion why do we have so little coverage by e2e tests. But it's again only my feeling  :wink:

alexeykazakov: +1 @MatousJobanek

PR #378, opened 2021-09-01
test: add label on OLM namespace

xcoulon:
this is only needed if the cluster in use was configured
with the `sandbox-cli setup` command. In this case, the
`name=openshift-openshift-operator-lifecycle-manager` label
is needed to successfully deploy the operators.

### DIFF
--- a/make/resources/default-network-policies.yaml
+++ b/make/resources/default-network-policies.yaml
@@ -1,0 +1,66 @@
+apiVersion: template.openshift.io/v1
+kind: Template
+metadata:
+  name: default-network-policies
+objects:
+- apiVersion: networking.k8s.io/v1
+  kind: NetworkPolicy
+  metadata:
+    labels:
+      provider: sandbox-sre
+    name: allow-from-same-namespace
+    namespace: ${NAMESPACE}
+  spec:
+    ingress:
+    - from:
+      - podSelector: {}
+    podSelector: null
+- apiVersion: networking.k8s.io/v1
+  kind: NetworkPolicy
+  metadata:
+    labels:
+      provider: sandbox-sre
+    name: allow-from-ingress-namespace
+    namespace: ${NAMESPACE}
+  spec:
+    ingress:
+    - from:
+      - namespaceSelector:
+          matchLabels:
+            network.openshift.io/policy-group: ingress
+    podSelector: null
+- kind: NetworkPolicy
+  apiVersion: networking.k8s.io/v1
+  metadata:
+    labels:
+      provider: sandbox-sre
+    name: allow-from-openshift-customer-monitoring-namespace
+    namespace: ${NAMESPACE}
+  spec:
+    podSelector: {}
+    ingress:
+    - from:
+      - namespaceSelector:
+          matchLabels:
+            name: openshift-customer-monitoring
+    policyTypes:
+    - Ingress
+- kind: NetworkPolicy
+  apiVersion: networking.k8s.io/v1
+  metadata:
+    labels:
+      provider: sandbox-sre
+    name: allow-from-openshift-operator-lifecycle-manager-namespace
+    namespace: ${NAMESPACE}
+  spec:
+    podSelector: {}
+    ingress:
+    - from:
+      - namespaceSelector:
+          matchLabels:
+            name: openshift-operator-lifecycle-manager
+    policyTypes:
+    - Ingress
+parameters:
+- name: NAMESPACE
+  required: true
--- a/make/resources/default-network-policies.yaml
+++ b/make/resources/default-network-policies.yaml
@@ -3,64 +3,16 @@ kind: Template
 metadata:
   name: default-network-policies
 objects:
-- apiVersion: networking.k8s.io/v1
-  kind: NetworkPolicy
-  metadata:
-    labels:
-      provider: sandbox-sre
-    name: allow-from-same-namespace
-    namespace: ${NAMESPACE}
-  spec:
-    ingress:
-    - from:
-      - podSelector: {}
-    podSelector: null
-- apiVersion: networking.k8s.io/v1
-  kind: NetworkPolicy
-  metadata:
-    labels:
-      provider: sandbox-sre
-    name: allow-from-ingress-namespace
-    namespace: ${NAMESPACE}
-  spec:
-    ingress:
-    - from:
-      - namespaceSelector:
-          matchLabels:
-            network.openshift.io/policy-group: ingress
-    podSelector: null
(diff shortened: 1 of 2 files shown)


MatousJobanek (approved):

make/resources/default-network-policies.yaml
```diff
@@ -0,0 +1,66 @@
...
+    namespace: ${NAMESPACE}
+  spec:
+    podSelector: {}
+    ingress:
+    - from:
+      - namespaceSelector:
+          matchLabels:
+            name: openshift-customer-monitoring
+    policyTypes:
+    - Ingress
```
MatousJobanek, line 47: IIRC `openshift-customer-monitoring` namespace is not present in OCP

make/test.mk
```diff
@@ -35,7 +35,18 @@ test-e2e: deploy-e2e e2e-run
 
 .PHONY: deploy-e2e
 deploy-e2e: INSTALL_OPERATOR=true
-deploy-e2e: build clean-e2e-files deploy-host deploy-members e2e-service-account setup-toolchainclusters
+deploy-e2e: build clean-e2e-files label-olm-ns deploy-host deploy-members e2e-service-account setup-toolchainclusters
+
+label-olm-ns: create-host-project create-member1 create-member2
```
MatousJobanek, line 40: these makefile targets are after that called also inside of the deploy steps. Technically, it shouldn't be a problem, but I'm not 100% sure

MatousJobanek (commented):
wouldn't be easier to just enable all ingress traffic: https://kubernetes.io/docs/concepts/services-networking/network-policies/#default-allow-all-ingress-traffic
```yaml
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: allow-all-ingress
spec:
  podSelector: {}
  ingress:
  - {}
  policyTypes:
  - Ingress
```
so we don't have to deal with updating the file?

make/test.mk
xcoulon: these makefile targets will print in error saying that the project(s) already exist, but that won't cause failure

xcoulon: indeed, there will be an error because the project already exists, but it's ignored and the script goes on:
```
Error from server (AlreadyExists): project.project.openshift.io "toolchain-host-operator" already exists
make[1]: [create-project] Error 1 (ignored)
```

xcoulon:
> wouldn't be easier to just enable all ingress traffic: https://kubernetes.io/docs/concepts/services-networking/network-policies/#default-allow-all-ingress-traffic
> 
> ```yaml
> apiVersion: networking.k8s.io/v1
> kind: NetworkPolicy
> metadata:
>   name: allow-all-ingress
> spec:
>   podSelector: {}
>   ingress:
>   - {}
>   policyTypes:
>   - Ingress
> ```
> 
> so we don't have to deal with updating the file?



> wouldn't be easier to just enable all ingress traffic: https://kubernetes.io/docs/concepts/services-networking/network-policies/#default-allow-all-ingress-traffic
> 
> ```yaml
> apiVersion: networking.k8s.io/v1
> kind: NetworkPolicy
> metadata:
>   name: allow-all-ingress
> spec:
>   podSelector: {}
>   ingress:
>   - {}
>   policyTypes:
>   - Ingress
> ```
> 
> so we don't have to deal with updating the file?

good idea
done in 67308df

make/test.mk
```diff
@@ -35,7 +35,18 @@ test-e2e: deploy-e2e e2e-run
...
+# adds a label on the oc label ns/openshift-operator-lifecycle-manager name=openshift-operator-lifecycle-manager
+# so that deployment also works when network policies were configured with `sandbox-cli`
+	@-oc label --overwrite=true ns/openshift-operator-lifecycle-manager name=openshift-operator-lifecycle-manager
+	@echo "adding network policies in $(HOST_NS) namespace"
+	@-oc process -p NAMESPACE=$(HOST_NS) -f ${PWD}/make/resources/default-network-policies.yaml | oc apply -f -
+	@echo "adding network policies in $(MEMBER_NS) namespace"
+	@-oc process -p NAMESPACE=$(MEMBER_NS) -f ${PWD}/make/resources/default-network-policies.yaml | oc apply -f -
+	@echo "adding network policies in $(MEMBER_NS_2) namespace"
+	@-oc process -p NAMESPACE=$(MEMBER_NS_2) -f ${PWD}/make/resources/default-network-policies.yaml | oc apply -f -
```
MatousJobanek, line 49: There is one thing I just realized - in dev environment, we don't create the second member namespace and we even don't set the `MEMBER_NS_2` var. So I'm afraid that it would fail in this case

MatousJobanek: btw, we already have `create-project` target - why not moving the creation of the NetworkPolicies there?

xcoulon: good point. I wanted to be sure that the network policies would exist before the deployment (otherwise this latter may fail), but setting the policies in the `create-project` target looks way simpler and straightforward, indeed! thanks!

xcoulon: > There is one thing I just realized - in dev environment, we don't create the second member namespace and we even don't set the `MEMBER_NS_2` var. So I'm afraid that it would fail in this case

yes, let me handle this case, too

xcoulon: done in 70bceb2

make/test.mk
```diff
@@ -212,9 +217,12 @@ endif
...
 .PHONY: create-project
 create-project:
-	-oc new-project ${PROJECT_NAME} 1>/dev/null
-	-oc project ${PROJECT_NAME}
-	-oc label ns ${PROJECT_NAME} toolchain.dev.openshift.com/provider=codeready-toolchain
+	@-oc new-project ${PROJECT_NAME} 1>/dev/null
+	@-oc project ${PROJECT_NAME}
+	@-oc label ns --overwrite=true ${PROJECT_NAME} toolchain.dev.openshift.com/provider=codeready-toolchain
+	@echo "adding network policies in $(HOST_NS) namespace"
+	@-oc process -p NAMESPACE=$(HOST_NS) -f ${PWD}/make/resources/default-network-policies.yaml | oc apply -f -
```
MatousJobanek, line 224: this should use `PROJECT_NAME` var:
```suggestion
	@echo "adding network policies in $(PROJECT_NAME) namespace"
	@-oc process -p NAMESPACE=$(PROJECT_NAME) -f ${PWD}/make/resources/default-network-policies.yaml | oc apply -f -
```

xcoulon: of course! 🤦‍♂️

MatousJobanek (approved):
:+1:

PR #381, opened 2021-09-10
chore: verify view & exec perms for crtadmins

MatousJobanek:
verifies that the expected Role & RoleBindings resources are created

paired with: https://github.com/codeready-toolchain/host-operator/pull/510

### DIFF
--- a/testsupport/tiers/checks.go
+++ b/testsupport/tiers/checks.go
@@ -347,7 +347,7 @@ func rbacEditRole() namespaceObjectsCheck {
 				{
 					APIGroups: []string{""},
 					Resources: []string{"pods/exec"},
-					Verbs:     []string{"get", "list", "watch", "create", "update", "delete"},
+					Verbs:     []string{"get", "list", "watch", "create", "delete", "update"},
 				},
 			},
 		}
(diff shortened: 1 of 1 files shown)


testsupport/tiers/checks.go
```diff
@@ -300,7 +284,55 @@ func rbacEditRoleBinding() namespaceObjectsCheck {
 	}
 }
 
+func crtadminViewRoleBinding() namespaceObjectsCheck {
+	return func(t *testing.T, ns *v1.Namespace, memberAwait *wait.MemberAwaitility, userName string) {
```
xcoulon, line 288: not a blocker here, but rather a general thought: I wonder if it's better to use Go API or external YAML file to describe the expectation. WDYT?

xcoulon (approved):

testsupport/tiers/checks.go
MatousJobanek: what do you mean by Go API?

I'm fine with using it as it is, but also open to any discussion/proposals for improvement - feel free to add it to the next sync call agenda.

MatousJobanek: as for the external file - I guess that it's better to use just go types & functions - it makes the objects/resources configurable for other tiers that just contain different numbers/values

PR #387, opened 2021-09-24
Fix malformed yaml template for network policies

amisevsk:
While using `make dev-deploy-e2e`, I noticed the error message
```
Already on project "toolchain-host-operator" on server "https://api.dwo-2021-09-24.devcluster.openshift.com:6443".
namespace/toolchain-host-operator labeled
adding network policies in toolchain-host-operator namespace
error: failed to read input object (not a Template?): unable to decode "/home/amisevsk/git/toolchain-e2e/make/resources/default-network-policies.yaml": v1.Template.Objects: []runtime.RawExtension: decode slice: expect [ or n, but found {, error found in #10 byte of ...|objects":{"apiVersio|..., bigger context ...|a":{"name":"default-network-policies"},"objects":{"apiVersion":"networking.k8s.io/v1","kind":"Networ|...
error: no objects passed to apply
make[2]: [make/test.mk:224: create-project] Error 1 (ignored)
make[2]: Leaving directory '/home/amisevsk/git/toolchain-e2e'
```

This PR fixes the malformed yaml (`objects` is a list, not an object)

### DIFF
--- a/make/resources/default-network-policies.yaml
+++ b/make/resources/default-network-policies.yaml
@@ -3,16 +3,16 @@ kind: Template
 metadata:
   name: default-network-policies
 objects:
-  apiVersion: networking.k8s.io/v1
-  kind: NetworkPolicy
-  metadata:
-    namespace: ${NAMESPACE}
-    name: allow-all-ingress
-  spec:
-    podSelector: {}
-    ingress:
-    - {}
-    policyTypes:
-    - Ingress
+  - apiVersion: networking.k8s.io/v1
+    kind: NetworkPolicy
+    metadata:
+      namespace: ${NAMESPACE}
+      name: allow-all-ingress
+    spec:
+      podSelector: {}
+      ingress:
+      - {}
+      policyTypes:
+      - Ingress
 parameters:
 - name: NAMESPACE


alexeykazakov (approved):
Good catch. @rajivnathan do we actually use this template?

rajivnathan:
Good catch indeed. Yes, it does look like we use it. It was added in https://github.com/codeready-toolchain/toolchain-e2e/pull/378 but as for whether we need it I'll defer to @MatousJobanek and @xcoulon

MatousJobanek:
@amisevsk thanks a lot for this fix - good catch.
@alexeykazakov This template is not super needed for all cases in OCP, but is pretty useful for app-sre OSD clusters to enable communication from OLM namespace. Since our e2e tests are executed in OCP, then we didn't notice that it wasn't actually applied. 
There is only one case when the NetworkPolicy is needed in OCP - when you combine usage of toolchain-e2e with sandbox-sre. @xcoulon does that from time to time, so that's why he added it there.

xcoulon:
oh good catch indeed, @amisevsk !
IIRC, I added this network policy as a template after manually applying it (ie, not as template), and I did not spot the error afterwards 🤦‍♂️

amisevsk:
Happy to help! :tada:

PR #388, opened 2021-09-28
Reduce risk of flakiness when dealing with member clusters

rajivnathan:
### Related Issue
https://issues.redhat.com/browse/CRT-1208

### Reason for the changes
This PR tries to reduce the probability of introducing flaky tests related to usersignup provisioning and verification.

#### Current Problems
1. A usersignup can be created without specifying a target member cluster so it can end up targeted to any member cluster (this is valid) but verification functions in e2e tests only accept a specific member cluster to verify against so it can be easy to forget to specify a target cluster when creating the usersignup and then specify a specific member cluster to verify against which leads to flakiness.
2. The second member cluster in the e2e environment is mainly for testing multi-member specific functionality but is returned by the WaitForDeployments func by default increasing the likelihood that someone uses it accidentally

See https://github.com/codeready-toolchain/toolchain-e2e/pull/364 for more details.

### Summary of changes

- Wrap the host awaitility and member awaitilities in a single object so it can be passed to verification functions like VerifyResourcesProvisionedForSignup 
```
type Awaitilities struct {
	t                  *testing.T
	hostAwaitility     *wait.HostAwaitility
	memberAwaitilities []*wait.MemberAwaitility
}
```

- Updated the VerifyResourcesProvisionedForSignup func to automatically determine which member cluster to use based on the provided usersignup. This avoids the test writer having to think about which member cluster to select when calling VerifyResourcesProvisionedForSignup

- Defined a func to retrieve a member cluster that accepts a selector or returns the first member cluster if no selector is provided. So if someone wants to test using the second member cluster, it requires a specific action to get it.
```
type memberSelector func(*wait.MemberAwaitility) bool

func (a Awaitilities) SecondMember(m *wait.MemberAwaitility) bool {
	return m.ClusterName == a.memberAwaitilities[1].ClusterName
}

func (a Awaitilities) Member(selectors ...memberSelector) *wait.MemberAwaitility {
	require.NotEmpty(a.t, a.memberAwaitilities, "there are no initialized member awaitilities")
	for _, selector := range selectors {
		for _, m := range a.memberAwaitilities {
			if selector(m) {
				return m
			}
		}
	}
	return a.memberAwaitilities[0]
}
```

### DIFF
--- a/test/e2e/base_user_test.go
+++ b/test/e2e/base_user_test.go
@@ -19,15 +19,15 @@ import (
 
 type baseUserIntegrationTest struct {
 	suite.Suite
-	awaitilities wait.Awaitilities
+	wait.Awaitilities
 }
 
 func (s *baseUserIntegrationTest) newSignupRequest() SignupRequest {
-	return NewSignupRequest(s.T(), s.awaitilities)
+	return NewSignupRequest(s.T(), s.Awaitilities)
 }
 
 func (s *baseUserIntegrationTest) createAndCheckBannedUser(email string) *toolchainv1alpha1.BannedUser {
-	hostAwait := s.awaitilities.Host()
+	hostAwait := s.Host()
 	// Create the BannedUser
 	bannedUser := newBannedUser(hostAwait, email)
 	err := hostAwait.CreateWithCleanup(context.TODO(), bannedUser)
@@ -53,7 +53,7 @@ func newBannedUser(host *wait.HostAwaitility, email string) *toolchainv1alpha1.B
 }
 
 func (s *baseUserIntegrationTest) deactivateAndCheckUser(userSignup *toolchainv1alpha1.UserSignup, mur *toolchainv1alpha1.MasterUserRecord) {
-	hostAwait := s.awaitilities.Host()
+	hostAwait := s.Host()
 	userSignup, err := hostAwait.UpdateUserSignupSpec(userSignup.Name, func(us *toolchainv1alpha1.UserSignup) {
 		states.SetDeactivated(us, true)
 	})
@@ -87,7 +87,7 @@ func (s *baseUserIntegrationTest) deactivateAndCheckUser(userSignup *toolchainv1
 }
 
 func (s *baseUserIntegrationTest) reactivateAndCheckUser(userSignup *toolchainv1alpha1.UserSignup, mur *toolchainv1alpha1.MasterUserRecord) {
-	hostAwait := s.awaitilities.Host()
+	hostAwait := s.Host()
 	err := hostAwait.Client.Get(context.TODO(), types.NamespacedName{
 		Namespace: userSignup.Namespace,
 		Name:      userSignup.Name,
--- a/test/e2e/base_user_test.go
+++ b/test/e2e/base_user_test.go
@@ -27,7 +27,7 @@ func (s *baseUserIntegrationTest) newSignupRequest() SignupRequest {
 }
 
 func (s *baseUserIntegrationTest) createAndCheckBannedUser(email string) *toolchainv1alpha1.BannedUser {
-	hostAwait := s.awaitilities.Host(s.T())
+	hostAwait := s.awaitilities.Host()
 	// Create the BannedUser
 	bannedUser := newBannedUser(hostAwait, email)
 	err := hostAwait.CreateWithCleanup(context.TODO(), bannedUser)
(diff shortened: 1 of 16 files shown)


testsupport/init.go
```diff
@@ -23,12 +23,18 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+type Awaitilities struct {
```
xcoulon, line 26: just curious: can't we have this struct in the same `wait` package as `HostAwaitility` and `MemberAwaitility`?

test/e2e/base_user_test.go
```diff
@@ -19,19 +19,18 @@ import (
 
 type baseUserIntegrationTest struct {
 	suite.Suite
-	hostAwait    *wait.HostAwaitility
-	memberAwait  *wait.MemberAwaitility
-	member2Await *wait.MemberAwaitility
+	awaitilities Awaitilities
```
xcoulon, line 22: I believe the code could be even simpler if we used something like this:
```
type baseUserIntegrationTest struct {
	suite.Suite
	Awaitilities
}
```
and then instead of `hostAwait := s.awaitilities.Host()` we would have `hostAwait := s.Host()`.

And one step further, if we rename the new `Awaitilities.Host()` method to `Awaitilities.HostAwait()`, then the old `s.hostAwait` becomes `s.HostAwait()` and we could get rid of the extra `hostAwait := s.awaitilities.Host()` statements and simply do 

```
s.HostAwait().CreateWithCleanup(context.TODO(), bannedUser)
```

for example

test/e2e/e2e_test.go
```diff
@@ -23,7 +23,10 @@ import (
 func TestE2EFlow(t *testing.T) {
 	// given
 	// full flow from usersignup with approval down to namespaces creation
-	hostAwait, memberAwait, memberAwait2 := WaitForDeployments(t)
+	awaitilities := WaitForDeployments(t)
+	hostAwait := awaitilities.Host()
+	memberAwait := awaitilities.Member()
```
xcoulon, line 28: I believe that for the clarity/readability, it would be simpler if we have `Member1()` and `Member2()` methods, aside of the `Member(...)` one

testsupport/init.go
rajivnathan: That's a good point, it's probably a better place for it. Moved: https://github.com/codeready-toolchain/toolchain-e2e/pull/388/commits/fff83ef4317afc41269587072474e3509ee1ff9d

test/e2e/e2e_test.go
rajivnathan: I thought about that but I wanted to make it more difficult to get the 2nd member just so it makes it harder to mistakenly use it. 
>The second member cluster in the e2e environment is mainly for testing multi-member specific functionality but is returned by the WaitForDeployments func by default increasing the likelihood that someone uses it accidentally

But if you think it's too confusing we can change it

test/e2e/base_user_test.go
rajivnathan: It's much better indeed, thanks for the suggestion! 👍  Done in c7fd9ef45d20ff26b25476830325e59722253eac

test/e2e/e2e_test.go
xcoulon: see code a few lines below:
```
signups := CreateMultipleSignups(t, awaitilities, awaitilities.Member(), 5)
```
I believe it would be more explicit if it was:
```
signups := CreateMultipleSignups(t, awaitilities, awaitilities.Member1(), 5)
```
so we now we're creating the 5 users on `Member1`

rajivnathan: Ah, yes I agree for Member1 that would be good. My comment was more for Member2. I can make the change for Member1 but what do you think about my comments regarding Member2?

xcoulon: so maybe having something explicit like `awaitilities.Member(Member1)` and `awaitilities.Member(Member2)` would work? My concern is that ``awaitilities.Member()` without any arg looks a bit strange to read.

rajivnathan: Ok, fair enough. I went with your first suggestion of using `Member1()` and `Member2()` since it's less verbose than `awaitilities.Member(Member1)`

Done in  https://github.com/codeready-toolchain/toolchain-e2e/pull/388/commits/cbb3fd66aeea0906274f65af34ac03179ff178ad

xcoulon:
cool, thanks for taking my comments into account @rajivnathan!

xcoulon (approved):

MatousJobanek (approved):
Nice :star_struck: very simple change and it brings a much safer test suite :+1: 
Thanks for doing that :1st_place_medal:

PR #389, opened 2021-09-30
e2e test for awaiting recreation of roles and rolebindings

ranakan19:
tests for functionality added in https://github.com/codeready-toolchain/member-operator/pull/294 which is fix wrt :https://issues.redhat.com/browse/CRT-1202


test/e2e/e2e_test.go
```diff
@@ -347,6 +347,63 @@ func TestE2EFlow(t *testing.T) {
...
+			ManuallyApprove().
+			EnsureMUR().
+			TargetCluster(memberAwait).
+			RequireConditions(ConditionSet(Default(), ApprovedByAdmin())...).
+			Execute().Resources()
+		devNs := corev1.Namespace{}
+		err := memberAwait.Client.Get(context.TODO(), types.NamespacedName{Name: "wonderwoman-dev"}, &devNs)
+		require.NoError(t, err)
+
+		stageNs := corev1.Namespace{}
```
alexeykazakov, line 363: You are not testing the stage namespace. Let's parameterize the actual tests so they test the deleted role/rolebindgins in the provided namespace and run them against both namespaces.

alexeykazakov (commented):
Looks good! Just one comment to address.

test/e2e/e2e_test.go
ranakan19: done in 74a0068

testsupport/user_assertions.go
```diff
@@ -119,3 +122,35 @@ func GetMurTargetMember(t *testing.T, awaitilities wait.Awaitilities, mur *toolc
 	require.FailNowf(t, "Unable to find a target member cluster", "MasterUserRecord: %+v", mur)
 	return nil
 }
+
+func DeletedRoleAndAwaitRecreation(t *testing.T, memberAwait *wait.MemberAwaitility, ns corev1.Namespace, role string) {
+	userRole, err := memberAwait.WaitForRole(&ns, role)
+	require.NoError(t, err)
+	require.NotEmpty(t, userRole)
+	require.Contains(t, userRole.Labels, "toolchain.dev.openshift.com/owner")
```
MatousJobanek, line 130: I guess that this check could be moved to "namespaceObjectChecks" in https://github.com/codeready-toolchain/toolchain-e2e/blob/master/testsupport/tiers/checks.go
https://github.com/codeready-toolchain/toolchain-e2e/blob/master/testsupport/tiers/checks.go#L358-L466

ranakan19: sure, thank you for pointing that out. I wasn't super sure where to move it, so not my best decision to move it to user_assertions

ranakan19: actually on second thought, since the function deletes it and waits for recreation, Not sure if it is a check?

MatousJobanek: I meant just the single line:
```
 	require.Contains(t, userRole.Labels, "toolchain.dev.openshift.com/owner") 
```
because we expect that all the created objects should contain the label, right?

ranakan19: oh sure, i misunderstood.

ranakan19: @MatousJobanek done in 3723235

ranakan19:
/retest
(infra issues)

MatousJobanek (approved):
Thanks for moving the label check :+1: 
Good work with this e2e test-case :1st_place_medal:

PR #393, opened 2021-10-14
Test promotion resets deactivating state

rajivnathan:
Related Issue: https://issues.redhat.com/browse/CRT-1222

Related PRs:
https://github.com/codeready-toolchain/host-operator/pull/524

Verifies that the deactivating state is reset after promoting a user to a new tier.

### DIFF
--- a/test/e2e/user_management_test.go
+++ b/test/e2e/user_management_test.go
@@ -123,7 +123,7 @@ func (s *userManagementTestSuite) TestUserDeactivation() {
 		require.NoError(t, err)
 
 		// Move the user to the new tier without deactivation enabled
-		murSyncIndex := MoveUserToTier(t, hostAwait, userSignupMember1.Spec.Username, *baseDeactivationDisabledTier).Spec.UserAccounts[0].SyncIndex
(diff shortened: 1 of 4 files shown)


test/e2e/user_management_test.go
```diff
@@ -363,6 +363,32 @@ func (s *userManagementTestSuite) TestUserDeactivation() {
 			})
 		})
 	})
+
+	s.T().Run("test reset deactivating state when promoting user", func(t *testing.T) {
```
alexeykazakov, line 367: Shouldn't we move this tests to the other promoting tests rather than keep it in the user deactivation test?

MatousJobanek (approved):
Agree with Alexey that it would make more sense to move it to the promote section of `nstemplatetier_test.go` file

test/e2e/user_management_test.go
rajivnathan: Makes sense. Done in 8820c0ae7b8ae2b27ebcd6664371945e5ca01307

alexeykazakov (approved):
Thanks!

PR #395, opened 2021-10-18
Returning user

rajivnathan:
Related PR: https://github.com/codeready-toolchain/host-operator/pull/527

### DIFF
--- a/test/e2e/usersignup_test.go
+++ b/test/e2e/usersignup_test.go
@@ -466,6 +466,50 @@ func (s *userSignupIntegrationTest) TestTransformUsername() {
 	}
 }
 
+func (s *userSignupIntegrationTest) TestReturningUserProvisionedToLastCluster() {
+	hostAwait := s.Host()
+	memberAwait := s.Member1()
+	memberAwait2 := s.Member2()
+	s.T().Run("test returning users provisioned to same cluster as last time", func(t *testing.T) {
+		// given
+		var memberLimits []testconfig.PerMemberClusterOptionInt
+		toolchainStatus, err := hostAwait.WaitForToolchainStatus(wait.UntilToolchainStatusHasConditions(ToolchainStatusReadyAndUnreadyNotificationNotCreated()...))
+		require.NoError(t, err)
+		for _, m := range toolchainStatus.Status.Members {
+			if memberAwait.ClusterName == m.ClusterName {
+				memberLimits = append(memberLimits, testconfig.PerMemberCluster(memberAwait.ClusterName, m.UserAccountCount+1))
+			} else if memberAwait2.ClusterName == m.ClusterName {
+				memberLimits = append(memberLimits, testconfig.PerMemberCluster(memberAwait2.ClusterName, m.UserAccountCount+1))
+			}
+		}
+		require.Len(s.T(), memberLimits, 2)
+
+		hostAwait.UpdateToolchainConfig(testconfig.AutomaticApproval().Enabled(true).MaxNumberOfUsers(0, memberLimits...))
+
+		// when
+		signup1, _ := s.newSignupRequest().
+			Username("multimember-1").
+			Email("multi1@redhat.com").
+			EnsureMUR().
+			LastCluster(memberAwait).
+			RequireConditions(ConditionSet(Default(), ApprovedAutomatically())...).
+			Execute().Resources()
+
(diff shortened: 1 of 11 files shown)


testsupport/signup_request.go
```diff
@@ -204,7 +211,19 @@ func (r *signupRequest) Execute() SignupRequest {
 			}
 		}
 
-		userSignup, err = hostAwait.UpdateUserSignupSpec(userSignup.Name, doUpdate)
+		userSignup, err = hostAwait.UpdateUserSignup(userSignup.Name, doUpdate)
+		require.NoError(r.t, err)
+	}
+
+	if r.lastCluster != nil {
```
alexeykazakov, line 218: I'm a bit confused... Why do we care about the last cluster in the request? The last cluster is fully internal thing which sets an internal preference. It should not be translated to the usersignup.spec.target cluster which is set by "clients" and a must (not a preference).

test/e2e/usersignup_test.go
```diff
@@ -466,6 +466,85 @@ func (s *userSignupIntegrationTest) TestTransformUsername() {
...
+			Execute().Resources()
+
+		// then
+		// Confirm the MUR was created and target cluster was set
+		VerifyResourcesProvisionedForSignup(s.T(), s.Awaitilities, userSignup, "base")
+
+		s.T().Run("user activated -> deactivated -> reactivated", func(t *testing.T) {
+			// given
+			mur, err := hostAwait.WaitForMasterUserRecord(userSignup.Status.CompliantUsername, wait.UntilMasterUserRecordHasConditions(Provisioned(), ProvisionedNotificationCRCreated()))
+			require.NoError(t, err)
```
MatousJobanek, line 493: the `s.newSignupRequest()....Resources()` returns MUR as the second value - you can use that one instead of getting it here again.

test/e2e/usersignup_test.go
```diff
@@ -466,6 +466,85 @@ func (s *userSignupIntegrationTest) TestTransformUsername() {
...
+			Username("returninguser").
+			Email("returninguser@redhat.com").
+			EnsureMUR().
+			RequireConditions(ConditionSet(Default(), ApprovedAutomatically())...).
+			Execute().Resources()
+
+		// then
+		// Confirm the MUR was created and target cluster was set
+		VerifyResourcesProvisionedForSignup(s.T(), s.Awaitilities, userSignup, "base")
+
```
MatousJobanek, line 489: You don't need to explicitly verify the resources if you use: `.EnsureMUR()`
```suggestion
```

test/e2e/usersignup_test.go
```diff
@@ -466,6 +466,85 @@ func (s *userSignupIntegrationTest) TestTransformUsername() {
...
+			Email("multi2@redhat.com").
+			EnsureMUR().
+			LastCluster(memberAwait2).
+			RequireConditions(ConditionSet(Default(), ApprovedAutomatically())...).
+			Execute().Resources()
+
+		// then
+		// Confirm the MUR was created and target cluster was set
+		VerifyResourcesProvisionedForSignup(s.T(), s.Awaitilities, signup1, "base")
+		VerifyResourcesProvisionedForSignup(s.T(), s.Awaitilities, signup2, "base")
```
MatousJobanek, line 544: TBH, this test is a bit confusing for me. 
1. I don't think that this is a real scenario - if there is a usersignup created from the RegistrationService, then it should not contain the "last-cluster annotation".
2. you don't verify that the user was provisioned to the same cluster as it was before
3. I guess that the functionality of provisioning users to the same cluster is covered by the first test-case enough
4. I would rather extend the first test-case by adding another subtests that limits the capacity of the cluster the user was provisioned before and let it provision to the other one.

alexeykazakov: > I would rather extend the first test-case by adding another subtests that limits the capacity of the cluster the user was provisioned before and let it provision to the other one.

But let's be careful with changing cluster settings if it may affect other tests (don't want to create flaky tests). I'm fine with testing the happy path only in e2e if it's tricky to do the full testing. Our unit tests cover everything so it should be fine.
But if it's all straightforward then sure, let's do it in e2e too.

alexeykazakov (approved):
Yes! It looks pretty clean now. Thanks!

MatousJobanek (approved):
Cool :+1:

PR #397, opened 2021-10-19
refactor test: use ConfigMap as blocker resource

xcoulon:

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -327,14 +327,14 @@ func TestE2EFlow(t *testing.T) {
 			require.NoError(t, err)
 
 			// then check all are deleted
+			err = memberAwait.WaitUntilNamespaceDeleted(laraUserName, "dev")
+			assert.NoError(t, err, "laracroft-dev namespace is not deleted")
+
 			err = memberAwait.WaitUntilNSTemplateSetDeleted(laraUserName)
 			assert.NoError(t, err, "NSTemplateSet is not deleted")
 
 			err = memberAwait.WaitUntilUserAccountDeleted(laraUserName)
 			require.NoError(t, err)
-
-			err = memberAwait.WaitUntilNamespaceDeleted(laraUserName, "dev")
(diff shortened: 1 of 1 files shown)


test/e2e/e2e_test.go
```diff
@@ -324,26 +319,29 @@ func TestE2EFlow(t *testing.T) {
...
+			assert.NoError(t, err, "NSTemplateSet is not deleted")
 
-		err = memberAwait.WaitUntilNSTemplateSetDeleted(laraUserName)
-		assert.NoError(t, err, "NSTemplateSet is not deleted")
+			err = memberAwait.WaitUntilUserAccountDeleted(laraUserName)
+			require.NoError(t, err)
 
-		err = memberAwait.WaitUntilUserAccountDeleted(laraUserName)
-		require.NoError(t, err)
+			err = memberAwait.WaitUntilNamespaceDeleted(laraUserName, "dev")
```
ranakan19, line 336: Is there a reason for checking Namespace is deleted **after** checking for deletion of NSTemplateSet and UserAccount? If the Namespace is not deleted, other resources in hierarchy shouldn't be deleted is the expectation that i was working with.

xcoulon: well, I thought it would be more logical to wait for resources within the namespace to be deleted _before_ checking the namespace itself. But I don't think it makes much differences in the end, it's really a matter of aesthetic here. The point is, when all resources are deleted, the order of the assertions does not matter

MatousJobanek: Kanika wasn't talking about resources within the namespaces, but NSTempalteSet & UserAccount CRs.
I agree with her that a more logical flow would be:
1. wait until Namespaces are deleted
2. wait until NSTemplateSet is deleted
3. wait until UserAccount is deleted

xcoulon: @MatousJobanek yes, it's me who got confused! NSTemplateSet and UserAccounts are not resources of the user namespaces 🤦‍♂️ 

I restored the order of the assertions in dd3aedc

PR #406, opened 2021-11-05
Appstudio tier

alexeykazakov:
Tests for codeready-toolchain/host-operator#534

### DIFF
--- a/testsupport/tiers/checks.go
+++ b/testsupport/tiers/checks.go
@@ -299,7 +299,7 @@ func (a *appstudioTierChecks) GetNamespaceObjectChecks(nsType string) []namespac
 		numberOfToolchainServiceAccounts(1),
 	)
 
-	checks = append(checks, commonNetworkPolicyChecks()...)
+	checks = append(checks, append(commonNetworkPolicyChecks(), networkPolicyAllowFromCRW(), numberOfNetworkPolicies(6))...)
 	return checks
 }
 
--- a/testsupport/tiers/checks.go
+++ b/testsupport/tiers/checks.go
@@ -24,6 +24,7 @@ import (
 const (
 	// tier names
 	advanced                 = "advanced"
+	appstudio                = "appstudio"
 	base                     = "base"
 	basedeactivationdisabled = "basedeactivationdisabled"
 	baseextended             = "baseextended"
@@ -67,6 +68,9 @@ func NewChecks(tier string) (TierChecks, error) {
 
 	case advanced:
 		return &advancedTierChecks{baseTierChecks{tierName: advanced}}, nil
+
+	case appstudio:
+		return &appstudioTierChecks{baseTierChecks{tierName: appstudio}}, nil
 
 	case test:
 		return &testTierChecks{tierName: test}, nil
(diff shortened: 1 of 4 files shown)


testsupport/tiers/checks.go
```diff
@@ -271,6 +275,57 @@ func (a *testTierChecks) GetClusterObjectChecks() []clusterObjectsCheck {
 	return []clusterObjectsCheck{}
 }
 
+type appstudioTierChecks struct {
+	baseTierChecks
+}
```
MatousJobanek, line 280: The `appstudio` tier is not based on the `base` tier, so we should not reuse it
```suggestion
type appstudioTierChecks struct {
	tierName string
}
```

testsupport/tiers/checks.go
```diff
@@ -68,6 +69,9 @@ func NewChecks(tier string) (TierChecks, error) {
 	case advanced:
 		return &advancedTierChecks{baseTierChecks{tierName: advanced}}, nil
 
+	case appstudio:
+		return &appstudioTierChecks{baseTierChecks{tierName: appstudio}}, nil
```
MatousJobanek, line 73:
```suggestion
		return &appstudioTierChecks{tierName: appstudio}, nil
```

testsupport/tiers/checks.go
```diff
@@ -327,6 +382,20 @@ func rbacEditRoleBinding() namespaceObjectsCheck {
...
 }
 
+func userViewRoleBinding() namespaceObjectsCheck {
+	return func(t *testing.T, ns *v1.Namespace, memberAwait *wait.MemberAwaitility, userName string) {
+		rb, err := memberAwait.WaitForRoleBinding(ns, "user-view")
+		require.NoError(t, err)
+		assert.Len(t, rb.Subjects, 1)
+		assert.Equal(t, "User", rb.Subjects[0].Kind)
+		assert.Equal(t, userName, rb.Subjects[0].Name)
+		assert.Equal(t, "edit", rb.RoleRef.Name)
```
MatousJobanek, line 392: This is supposed to be `view` and not `edit`, right?
```suggestion
		assert.Equal(t, "view", rb.RoleRef.Name)
```

testsupport/tiers/checks.go
```diff
@@ -271,6 +275,57 @@ func (a *testTierChecks) GetClusterObjectChecks() []clusterObjectsCheck {
...
+type appstudioTierChecks struct {
+	baseTierChecks
+}
+
+func (a *appstudioTierChecks) GetTierObjectChecks() []tierObjectCheck {
+	return []tierObjectCheck{nsTemplateTier(a.tierName, 30)}
+}
+
+func (a *appstudioTierChecks) GetNamespaceObjectChecks(nsType string) []namespaceObjectsCheck {
+	checks := append(commonChecks,
```
MatousJobanek, line 287: you cannot use the `commonChecks` as they expect `user-edit` RoleBinding which is not part of this tier

testsupport/tiers/checks.go
```diff
@@ -271,6 +275,57 @@ func (a *testTierChecks) GetClusterObjectChecks() []clusterObjectsCheck {
...
+		userSaReadRoleBinding(),
+		execPodsRole(),
+		toolchainSaReadRole(),
+		appstudioServiceAccount(),
+		numberOfToolchainRoles(2),
+		numberOfToolchainRoleBindings(5),
+		numberOfToolchainServiceAccounts(1),
+	)
+
+	checks = append(checks, commonNetworkPolicyChecks()...)
```
MatousJobanek, line 302: `commonNetworkPolicyChecks` doesn't contain the networkPolicy for CRW. I should also check the number of present networkPolicies
```suggestion
	checks = append(checks, append(commonNetworkPolicyChecks(), networkPolicyAllowFromCRW(), numberOfNetworkPolicies(6))...)
```

xcoulon (approved):

PR #409, opened 2021-11-17
use 't' instead of 's.T()' in substests

xcoulon:


MatousJobanek (approved):
Thanks :+1:

PR #410, opened 2021-11-17
fix: don't reuse connections

MatousJobanek:
IIRC http.Transport reuses connections by default in go, so it could have happened that it reused some old connection that was closed by the server. In that case, the client could get the EOF error. 
by setting `req.Close = true` we should prevent from reusing the connection.


xcoulon (approved):

xcoulon:
@MatousJobanek awesome, I didn't know about that!

PR #416, opened 2021-12-01
feat: drop codecov to make the PR reviews easier

MatousJobanek:
There is no point in having the code-coverage for e2e tests and the CodeCov "warning" comments in the PR reviews are really annoying and make the review harder.


xcoulon (approved):
oh well, I thought it could be useful, but I understand it makes reviews harder :/

PR #420, opened 2021-12-09
test: ChangeTierRequest on Spaces

xcoulon:
E2E tests for https://github.com/codeready-toolchain/host-operator/pull/555


go.mod
```diff
@@ -35,4 +35,6 @@ require (
 	sigs.k8s.io/controller-runtime v0.8.3
 )
 
+replace github.com/codeready-toolchain/api => github.com/xcoulon/api v0.0.0-20211214154014-3e4fc7ad9c34
```
rajivnathan, line 38: Friendly reminder

testsupport/wait/host.go
```diff
@@ -1400,13 +1400,24 @@ func (a *HostAwaitility) printSpaceWaitCriterionDiffs(actual *toolchainv1alpha1.
...
-	a.listAndPrint("UserSignups", a.Namespace, &toolchainv1alpha1.UserSignupList{})
-	a.listAndPrint("MasterUserRecords", a.Namespace, &toolchainv1alpha1.MasterUserRecordList{})
-
+	// also include Spaces resources in the host namespace, to help troubleshooting
+	a.listAndPrint("Spaces", a.Namespace, &toolchainv1alpha1.SpaceList{})
 	a.T.Log(buf.String())
 }
 
+// UntilSpaceHasConditions returns a `SpaceWaitCriterion` which checks that the given
+// Space has exactly all the given status conditions
```
rajivnathan, line 1409:
```suggestion
// UntilSpaceHasTier returns a `SpaceWaitCriterion` which checks that the given
// Space has the expected tier name set in its Spec
```

test/e2e/space_test.go
```diff
@@ -128,29 +125,118 @@ func TestSpace(t *testing.T) {
...
+						// then
+						require.NoError(t, err)
+					})
+				})
 			})
 		})
 	})
 }
 
+func TestUpdateSpace(t *testing.T) {
```
rajivnathan, line 165:
```suggestion
func TestUpdateSpaceTier(t *testing.T) {
```
Would it be better to rename this func to be specifically about updating the tier? Also, maybe we should put this test in the nstemplatetier_test.go file?

test/e2e/space_test.go
```diff
@@ -128,29 +125,118 @@ func TestSpace(t *testing.T) {
...
+
+			// when
+			err = hostAwait.Client.Delete(context.TODO(), space)
+
+			// then
+			require.NoError(t, err)
+			err = hostAwait.WaitUntilSpaceDeleted(space.Name)
+			require.NoError(t, err)
+		})
+	})
```
rajivnathan, line 237:
```suggestion
```
I guess testing deletion isn't really required as part of the update tier test? Unless you think it provides some specific value to the test

test/e2e/space_test.go
```diff
@@ -128,29 +125,118 @@ func TestSpace(t *testing.T) {
...
+		// given
+		ctr := NewChangeTierRequest(hostAwait.Namespace, space.Name, "advanced")
+
+		// when
+		err = hostAwait.Client.Create(context.TODO(), ctr)
+
+		// then
+		require.NoError(t, err)
+		err = hostAwait.WaitUntilChangeTierRequestDeleted(ctr.Name)
+		space, err = hostAwait.WaitForSpace(space.Name, wait.UntilSpaceHasTier("advanced"), wait.UntilSpaceHasConditions(Provisioned()))
```
rajivnathan, line 215: Should we also verify that the tier hash label is updated to match the hash of the advanced tier?

test/e2e/space_test.go
xcoulon: In an upcoming PR this test will also contain a series of subtests for when the Space is migrated to another member cluster, so I'd rather keep it named as it is now

test/e2e/space_test.go
xcoulon: not really, indeed, but I wanted to clean up the resources at the end of the test. But I'm fine with removing this subtest for the sake of clarity/understanding, and as you said, because it's out of purpose

test/e2e/space_test.go
xcoulon: good point, since it just revealed in bug in the controller: `space` had both `toolchain.dev.openshift.com/advanced-tier-hash` and `toolchain.dev.openshift.com/base-tier-hash` labels!

test/e2e/space_test.go
```diff
@@ -128,29 +125,106 @@ func TestSpace(t *testing.T) {
...
+				err = hostAwait.Client.Delete(context.TODO(), s)
 
-				// then
+				// then it should fail while the member cluster is unknown (ie, unreachable)
 				require.NoError(t, err)
-				_, err = hostAwait.WaitForSpace(space.Name, wait.UntilSpaceHasConditions(TerminatingFailed("Cannot delete NSTemplateSet: unknown target member cluster: 'unknown'")))
+				s, err = hostAwait.WaitForSpace(s.Name, wait.UntilSpaceHasConditions(TerminatingFailed("cannot delete NSTemplateSet: unknown target member cluster: 'unknown'")))
 				require.NoError(t, err)
+
+				t.Run("update target cluster to unblock deletion", func(t *testing.T) {
```
MatousJobanek, line 144: just a side note - I'm not sure if we need to verify all these error-cases also in e2e tests - they are well covered by unit tests. And here it makes the e2e test execution even longer.
To be clear, I'm fine with keeping them there - it brings another level of quality. I just wanted to point out the consequence of it, so it's up to you :wink:

test/e2e/space_test.go
```diff
@@ -128,29 +125,106 @@ func TestSpace(t *testing.T) {
...
+			Name:      "oddity-" + strconv.Itoa(rand.Int()),
+		},
+		Spec: toolchainv1alpha1.SpaceSpec{
+			TargetCluster: memberAwait.ClusterName,
+			TierName:      "base",
+		},
+	}
+
+	// when
+	err := hostAwait.Client.Create(context.TODO(), space)
```
MatousJobanek, line 188: I would use 
```suggestion
	err := hostAwait.CreateWithCleanup(context.TODO(), space)
```
so it's cleaned at the end of the test.

test/e2e/space_test.go
```diff
@@ -128,29 +125,106 @@ func TestSpace(t *testing.T) {
...
+	t.Run("update tier", func(t *testing.T) {
+		// given
+		ctr := NewChangeTierRequest(hostAwait.Namespace, space.Name, "advanced")
+
+		// when
+		err = hostAwait.Client.Create(context.TODO(), ctr)
+
+		// then
+		require.NoError(t, err)
+		err = hostAwait.WaitUntilChangeTierRequestDeleted(ctr.Name)
```
MatousJobanek, line 214: We don't have to wait for the deletion of the ChangeTierRequest here - it can take a few seconds. I would move this at the very end of this test and here do this:
```suggestion
		_, err := hostAwait.WaitForChangeTierRequest(ctr.Name, toBeComplete)
		require.NoError(t, err)
```

test/e2e/space_test.go
```diff
@@ -128,29 +125,106 @@ func TestSpace(t *testing.T) {
...
+		space, err = hostAwait.WaitForSpace(space.Name, wait.UntilSpaceHasTier("advanced"), wait.UntilSpaceHasConditions(Provisioned()))
+		require.NoError(t, err)
+		nsTmplSet, err := memberAwait.WaitForNSTmplSet(space.Name, wait.UntilNSTemplateSetHasConditions(Provisioned()))
+		require.NoError(t, err)
+		tierChecks, err := tiers.NewChecks("advanced")
+		require.NoError(t, err)
+		tiers.VerifyGivenNsTemplateSet(t, memberAwait, nsTmplSet, tierChecks, tierChecks, tierChecks.GetExpectedTemplateRefs(hostAwait))
+		space, err = hostAwait.WaitForSpace(space.Name, wait.UntilSpaceHasConditions(Provisioned()))
+		require.NoError(t, err)
```
MatousJobanek, line 223: We already waited for the Space to be provisioned that at line 215, so why do it here again?

test/e2e/space_test.go
xcoulon: I don't think we can cover all error cases in the e2e tests, but for these particular cases, I would keep them, unless they cause more troubles than they bring benefits.

test/e2e/space_test.go
xcoulon: good idea, thanks!

test/e2e/space_test.go
xcoulon: that's right and it was removed in the mean time.

test/e2e/space_test.go
MatousJobanek: I didn't want to cover all error cases in e2e tests ever.
Ok, if they don't cause any troubles and if we are fine with making the execution of the e2e tests longer, then I'm fine with keeping them there :+1:

MatousJobanek (approved):

PR #426, opened 2021-12-16
test: stringify object(s) to YAML without managedFields

xcoulon:
When a test fails, resources are logged but now without the `metadata.managedFields` part, which is verbose and not relevant in the context of our E2E tests.


alexeykazakov (approved):
Nice!

xcoulon:
> Nice!

thanks! 😊

MatousJobanek:
I used the `metadata.managedFields` from time to time - it's useful when you need to know which client did what kind of operation and when on the CR. It's not used every time, but in some rare cases knowing who updated what is really useful.

xcoulon:
> I used the `metadata.managedFields` from time to time - it's useful when you need to know which client did what kind of operation and when on the CR. It's not used every time, but in some rare cases knowing who updated what is really useful.

that's interesting, I never used `managedFields` but now I vaguely recall you mentioned it. Could you elaborate/give an example of when this field is helpful?

MatousJobanek:
I cannot point you to the specific use-case when I used it in debugging e2e tests. But from the top of my head, I remember that we used it for example when we were investigating the bug in UI that caused the creation of thousands of replicasets: https://bugzilla.redhat.com/show_bug.cgi?id=1925180

PR #427, opened 2021-12-20
feat: add appstudio-e2e-deploy-latest target

MatousJobanek:
introduce: 
* appstudio-e2e environment config YAML files
* appstudio-e2e-deploy-latest target that deploys the latest toolchain operators in appstudio-e2e environment
* deletion of registration-service pods after the `ToolchainConfig` CR is created - this is for the case when we would run for example `appstudio-dev-deploy-latest` & `appstudio-e2e-deploy-latest` after each other, without cleaning the namespaces.

refactoring:
* move all appstudio related targets to `appstudio.mk`
* a few changes in the `clean.mk` to make it more flexible and reusable 

Fixes:
* the way we get the registration page URL - the `$(eval ROUTE = $(shell ` is evaluated at the very beginning when the target was called, so if it wasn't available at that time, then the echo would print only `Access the Landing Page here: https://`
* add deletion of the Spaces to the `clean-users` target

### DIFF
--- a/deploy/host-operator/appstudio-e2e/secrets.yaml
+++ b/deploy/host-operator/appstudio-e2e/secrets.yaml
@@ -1,0 +1,9 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: host-operator-secret
+type: Opaque
+data:
+  twilio.account.sid: QUM4MWE4MTcwMTQ5NDU2MmQ3OWFjZmRkNzlkNjhkZTg3Yw==
+  twilio.auth.token: NjZlMzllNGUxOTZiMjFlOTM1YzU0OTA4M2RhMGU3NzI=
+  twilio.from_number: KzE1MDA1NTUwMDA2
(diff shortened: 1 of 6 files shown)


xcoulon (approved):

deploy/host-operator/appstudio-e2e/secrets.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: host-operator-secret
+type: Opaque
+data:
+  twilio.account.sid: QUM4MWE4MTcwMTQ5NDU2MmQ3OWFjZmRkNzlkNjhkZTg3Yw==
+  twilio.auth.token: NjZlMzllNGUxOTZiMjFlOTM1YzU0OTA4M2RhMGU3NzI=
+  twilio.from_number: KzE1MDA1NTUwMDA2
```
alexeykazakov, line 9: This is minor but can we use some values which make it clear that those are dummy ones? The current values look real :)

alexeykazakov (approved):

deploy/host-operator/appstudio-e2e/secrets.yaml
MatousJobanek: Yeah, I'll do that for all secrets in this repo - but in the next PR and in the next year :-)

PR #429, opened 2021-12-21
e2e tests for websockets proxy

alexeykazakov:


test/e2e/proxy_test.go
```diff
@@ -188,3 +212,127 @@ func createPreexistingUserAndIdentity(t *testing.T, user proxyUsers) (*userv1.Us
...
+	user       proxyUser
+	connection *websocket.Conn
+	proxyHost  string
+
+	mu          sync.RWMutex
+	receivedCMs map[string]*corev1.ConfigMap
+}
+
+// start creates a new WebSocket connection. The method returns a function which is to be used to close the connection when done.
+func (w *wsWatcher) Start() func() {
```
xcoulon, line 241: this method doesn't need to be exported

test/e2e/proxy_test.go
```diff
@@ -188,3 +212,127 @@ func createPreexistingUserAndIdentity(t *testing.T, user proxyUsers) (*userv1.Us
...
+		err = json.Unmarshal(msg, &message)
+		require.NoError(w.t, err)
+		copyCM := message.ConfigMap
+		w.mu.Lock()
+		w.receivedCMs[copyCM.Name] = &copyCM
+		w.mu.Unlock()
+	}
+}
+
+func (w *wsWatcher) WaitForConfigMap(retryInterval, timeout time.Duration, expectedCMName string) (*corev1.ConfigMap, error) {
```
xcoulon, line 329: this method doesn't need to be exported

test/e2e/proxy_test.go
```diff
@@ -88,13 +98,27 @@ func TestProxyFlow(t *testing.T) {
...
+				defer closeConnection()
 
 				// when
 				proxyCl := hostAwait.CreateAPIProxyClient(user.token)
 				err := proxyCl.Create(context.TODO(), expectedCM)
 				require.NoError(t, err)
 
 				// then
+				// wait for the websocket watcher which uses the proxy to receive the config map
+				found, err := w.WaitForConfigMap(
```
xcoulon, line 113: what about testing that multiple resources can be received with the same websocket connection?

xcoulon (approved):

MatousJobanek (approved):
Nice :+1:

test/e2e/proxy_test.go
alexeykazakov: That's right but I decided to export `Start()` and `WaitForConfigMap()` as the only "public" API of the watcher struct. Only those two methods are supposed to be used by clients. Other methods only to be used by the watcher itself.
So if you don't mind I would keep it as is in attempt to make it more clear ("external" vs. internal API).

test/e2e/proxy_test.go
alexeykazakov: See above.

alexeykazakov:
@xcoulon @MatousJobanek thank you for the review!

PR #443, opened 2022-01-17
feat: don't build when use master

MatousJobanek:
* a few small changes in the makefiles to use the new functionality provided by https://github.com/codeready-toolchain/toolchain-cicd/pull/44
* don't create e2e ToolchainCluster CR in dev mode


make/run-cicd-script.mk
```diff
@@ -1,5 +1,5 @@
 
-OWNER_AND_BRANCH_LOCATION=codeready-toolchain/toolchain-cicd/master
+OWNER_AND_BRANCH_LOCATION=MatousJobanek/toolchain-cicd/support-deploy-latest
```
rajivnathan, line 2: I guess this needs to be reverted?

rajivnathan (commented):
Looks good, just 1 question

make/run-cicd-script.mk
MatousJobanek: yes, sure - this is to make the e2e tests pass and also to verify the change :-)

alexeykazakov (approved):

make/run-cicd-script.mk
MatousJobanek: updated a05a451

xcoulon (approved):

PR #446, opened 2022-01-17
refactor: rename awaitilities in init.go to avoid confusion

MatousJobanek:
When any of the awaitility vars are used in the same package, then the current names are pretty confusing as the IDE suggests using these "global" variables. So I renamed them to avoid confusion.


xcoulon (approved):
maybe we could rename those vars to `initHostAwait`, `initMemberAwait` and `initMemberAwait2` since they are used for initialization/setup, but it's not a big deal

MatousJobanek:
@xcoulon good point - renamed in c4682b2

xcoulon:
> @xcoulon good point - renamed in [c4682b2](https://github.com/codeready-toolchain/toolchain-e2e/commit/c4682b25e4eb4a2f795b742b1ceda12821901a83)

great, thanks for taking my comment into account @MatousJobanek 🙌

PR #447, opened 2022-01-18
refactor: extract some functions to make it available outside of the e2e package

MatousJobanek:
This PR extracts some Space-related functions and makes them a bit more generic so they can be easily used in the migration tests

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -27,7 +27,6 @@ require (
 	github.com/spf13/cobra v1.1.3
 	github.com/spf13/viper v1.7.1
 	github.com/stretchr/testify v1.7.0
-	honnef.co/go/tools v0.0.1-2020.1.6 // indirect
 	k8s.io/api v0.20.2
 	k8s.io/apimachinery v0.20.2
 	k8s.io/client-go v0.20.2
@@ -35,5 +34,7 @@ require (
 	k8s.io/metrics v0.20.2
 	sigs.k8s.io/controller-runtime v0.8.3
 )
+
+replace github.com/codeready-toolchain/toolchain-common => github.com/MatousJobanek/toolchain-common v0.0.0-20220118175315-c78f396a70df
 
 go 1.16
--- a/go.mod
+++ b/go.mod
@@ -2,7 +2,7 @@ module github.com/codeready-toolchain/toolchain-e2e
 
 require (
 	github.com/codeready-toolchain/api v0.0.0-20220110101730-04d30e90a574
-	github.com/codeready-toolchain/toolchain-common v0.0.0-20220110102259-392256a24155
+	github.com/codeready-toolchain/toolchain-common v0.0.0-20220118183120-7037f272026f
 	github.com/davecgh/go-spew v1.1.1
 	github.com/fatih/color v1.10.0
 	github.com/ghodss/yaml v1.0.0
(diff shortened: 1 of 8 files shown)


testsupport/space.go
```diff
@@ -0,0 +1,95 @@
...
+		},
+	}
+	if targetCluster != nil {
+		space.Spec.TargetCluster = targetCluster.ClusterName
+	}
+	return space
+}
+
+// GenerateName appends generated UUID to the given string
+func GenerateName(prefix string) string {
```
rajivnathan, line 36: Should this func be moved to a more common location since it's not specific to spaces?

testsupport/space.go
```diff
@@ -0,0 +1,95 @@
...
+	return fmt.Sprintf("%s-%s", prefix, uuid.Must(uuid.NewV4()).String())
+}
+
+// VerifyResourcesProvisionedForSpaceWithTier verifies that the Space for the given name is provisioned with all needed labels and conditions.
+// It also checks the NSTemplateSet and all related namespace & cluster scoped resources
```
rajivnathan, line 41: This function is pretty close to `VerifyResourcesProvisionedForSpaceWithTiers`, it wasn't immediately obvious to me so maybe we should change this comment to
```suggestion
// Same as VerifyResourcesProvisionedForSpaceWithTiers but reuses the provided tier name for the namespace and cluster resources names
```

testsupport/tiers/templaterefs.go
```diff
@@ -32,3 +37,33 @@ func clusterResourcesRevision(tier toolchainv1alpha1.NSTemplateTier) *string {
...
+	m, err := json.Marshal(templateRefs{Refs: refs})
+	if err != nil {
+		return "", err
+	}
+	md5hash := md5.New()
+	// Ignore the error, as this implementation cannot return one
+	_, _ = md5hash.Write(m)
+	hash := hex.EncodeToString(md5hash.Sum(nil))
+	return hash, nil
+}
```
rajivnathan, line 60: Looks like we have the same function here: https://github.com/codeready-toolchain/toolchain-common/blob/master/pkg/test/masteruserrecord/masteruserrecord.go#L125

Maybe we should just move that one to another package in toolchain-common and reference it?

rajivnathan: Actually, maybe this whole file can be moved to toolchain-common

rajivnathan (commented):
This simplifies things a lot! 👍

testsupport/space.go
MatousJobanek: I wasn't sure where to put it, so I left it there. But you are right - moved 5d18035

testsupport/tiers/templaterefs.go
MatousJobanek: good point - I wasn't aware of that. Move in https://github.com/codeready-toolchain/toolchain-common/pull/230

and I moved only these new functions - the other ones that are in this file are e2e specific, so I would keep them here

go.mod
```diff
@@ -36,4 +35,6 @@ require (
 	sigs.k8s.io/controller-runtime v0.8.3
 )
 
+replace github.com/codeready-toolchain/toolchain-common => github.com/MatousJobanek/toolchain-common v0.0.0-20220118175315-c78f396a70df
```
alexeykazakov, line 38: A reminder.

rajivnathan (approved):
Thanks! 🙌

alexeykazakov (approved):
Nice cleanup.

go.mod
MatousJobanek: updated dbf2e1c

PR #449, opened 2022-01-19
feat: introduce operator update tests

MatousJobanek:
This PR introduces migration infra & tests. It covers that the operator updates don't break anything and don't cause deletion of the existing accounts.
I called them "migration tests", but we could also call them "update tests" - I can change it if needed.

There are 3 significant changes:
1. a few modifications/refactoring in `test.mk` so:
    - by default the `test-e2e` target will:
      * setup the projects
      * deploy the latest version of our operators
      * run "before migration" tests which will provision UserSignups and Spaces
      * reinstall the operators to use the one from PR
      * run "after migration"tests which will check that nothing is broken nor deleted
      * run the standard set of tests
    - there is a new target `test-e2e-without-migration` which will skip the migration part. That means that it won't install the latest versions of operators nor run the migration tests.
    
2. new `test/migration/setup/setup_migration_test.go`
    - creates a few UserSignups & Spaces and puts the into various states
    - doesn't clean them so they can be used in the next "after migration" step
    - runs everything in go routines in parallel - to make it faster
    
3. new `test/migration/verify/verify_migration_test.go`
    - gets all UserSignups & Spaces in the previous step
    - verifies that they exist in the expected state
    - does a few operations on the resources to verify that everthing works as expected
    - all these verifications runs in go routines in parallel (to make it faster)
    - at the end it also executes the same set of setup functions as it was executed in `test/migration/setup/setup_migration_test.go` to verify that the setup logic is compatible also with the new version of the operator and e2e tests

### DIFF
--- a/make/test.mk
+++ b/make/test.mk
@@ -43,16 +43,16 @@ test-e2e-without-migration: prepare-e2e deploy-e2e e2e-run
 	@echo "To clean the cluster run 'make clean-e2e-resources'"
 
 .PHONY: verify-migration-and-deploy-e2e
-verify-migration-and-deploy-e2e: prepare-projects e2e-deploy-latest e2e-service-account setup-toolchainclusters e2e-migration-first-run get-publish-and-install-operators e2e-migration-second-run
+verify-migration-and-deploy-e2e: prepare-projects e2e-deploy-latest e2e-service-account setup-toolchainclusters e2e-migration-setup get-publish-and-install-operators e2e-migration-verify
 
-.PHONY: e2e-migration-first-run
-e2e-migration-first-run:
+.PHONY: e2e-migration-setup
+e2e-migration-setup:
 	@echo "Setting up the environment before testing the operator migration..."
 	$(MAKE) execute-tests MEMBER_NS=${MEMBER_NS} MEMBER_NS_2=${MEMBER_NS_2} HOST_NS=${HOST_NS} REGISTRATION_SERVICE_NS=${REGISTRATION_SERVICE_NS} TESTS_TO_EXECUTE="./test/migration/setup"
 	@echo "Environment successfully setup."
 
-.PHONY: e2e-migration-second-run
-e2e-migration-second-run:
+.PHONY: e2e-migration-verify
+e2e-migration-verify:
 	@echo "Updating operators and verifying resources..."
 	$(MAKE) execute-tests MEMBER_NS=${MEMBER_NS} MEMBER_NS_2=${MEMBER_NS_2} HOST_NS=${HOST_NS} REGISTRATION_SERVICE_NS=${REGISTRATION_SERVICE_NS} TESTS_TO_EXECUTE="./test/migration/verify"
 	@echo "Migration tests successfully finished"
--- a/make/test.mk
+++ b/make/test.mk
@@ -38,7 +38,7 @@ test-e2e: prepare-e2e verify-migration-and-deploy-e2e e2e-run
 	@echo "To clean the cluster run 'make clean-e2e-resources'"
 
 .PHONY: test-e2e-without-migration
-## Run the e2e tests with migration tests
+## Run the e2e tests without migration tests
 test-e2e-without-migration: prepare-e2e deploy-e2e e2e-run
 	@echo "To clean the cluster run 'make clean-e2e-resources'"
 
--- a/make/test.mk
+++ b/make/test.mk
@@ -47,7 +47,7 @@ cover-migration-and-deploy: prepare-projects e2e-deploy-latest e2e-service-accou
 
 .PHONY: e2e-migration-first-run
 e2e-migration-first-run:
-	@echo "Running before migration tests to setup the environment..."
+	@echo "Setting up the environment before testing the operator migration..."
 	$(MAKE) execute-tests MEMBER_NS=${MEMBER_NS} MEMBER_NS_2=${MEMBER_NS_2} HOST_NS=${HOST_NS} REGISTRATION_SERVICE_NS=${REGISTRATION_SERVICE_NS} TESTS_TO_EXECUTE="./test/migration/setup"
 	@echo "Before migration tests successfuly finished."
 
--- a/make/test.mk
+++ b/make/test.mk
@@ -49,7 +49,7 @@ cover-migration-and-deploy: prepare-projects e2e-deploy-latest e2e-service-accou
 e2e-migration-first-run:
 	@echo "Setting up the environment before testing the operator migration..."
 	$(MAKE) execute-tests MEMBER_NS=${MEMBER_NS} MEMBER_NS_2=${MEMBER_NS_2} HOST_NS=${HOST_NS} REGISTRATION_SERVICE_NS=${REGISTRATION_SERVICE_NS} TESTS_TO_EXECUTE="./test/migration/setup"
-	@echo "Before migration tests successfuly finished."
+	@echo "Environment successfully setup."
 
 .PHONY: e2e-migration-second-run
 e2e-migration-second-run:
--- a/make/test.mk
+++ b/make/test.mk
@@ -55,7 +55,7 @@ e2e-migration-first-run:
 e2e-migration-second-run:
 	@echo "Updating operators and verifying resources..."
 	$(MAKE) execute-tests MEMBER_NS=${MEMBER_NS} MEMBER_NS_2=${MEMBER_NS_2} HOST_NS=${HOST_NS} REGISTRATION_SERVICE_NS=${REGISTRATION_SERVICE_NS} TESTS_TO_EXECUTE="./test/migration/verify"
-	@echo "After migration tests successfuly finished"
+	@echo "Migration tests successfully finished"
 
 .PHONY: e2e-deploy-latest
 e2e-deploy-latest:
--- a/make/test.mk
+++ b/make/test.mk
@@ -33,7 +33,7 @@ SETUP_E2E_SERVICE_ACCOUNTS ?= true
 .PHONY: test-e2e
 ## Run the e2e tests
 test-e2e: INSTALL_OPERATOR=true
-test-e2e: prepare-e2e cover-migration-and-deploy e2e-run
+test-e2e: prepare-e2e verify-migration-and-deploy-e2e e2e-run
 	@echo "The tests successfully finished"
 	@echo "To clean the cluster run 'make clean-e2e-resources'"
 
@@ -42,8 +42,8 @@ test-e2e: prepare-e2e cover-migration-and-deploy e2e-run
 test-e2e-without-migration: prepare-e2e deploy-e2e e2e-run
 	@echo "To clean the cluster run 'make clean-e2e-resources'"
 
-.PHONY: cover-migration-and-deploy
-cover-migration-and-deploy: prepare-projects e2e-deploy-latest e2e-service-account setup-toolchainclusters e2e-migration-first-run get-publish-and-install-operators e2e-migration-second-run
+.PHONY: verify-migration-and-deploy-e2e
+verify-migration-and-deploy-e2e: prepare-projects e2e-deploy-latest e2e-service-account setup-toolchainclusters e2e-migration-first-run get-publish-and-install-operators e2e-migration-second-run
 
 .PHONY: e2e-migration-first-run
 e2e-migration-first-run:
--- a/make/test.mk
+++ b/make/test.mk
@@ -32,13 +32,41 @@ SETUP_E2E_SERVICE_ACCOUNTS ?= true
 
 .PHONY: test-e2e
 ## Run the e2e tests
-test-e2e: deploy-e2e e2e-run
+test-e2e: INSTALL_OPERATOR=true
+test-e2e: prepare-e2e cover-migration-and-deploy e2e-run
 	@echo "The tests successfully finished"
 	@echo "To clean the cluster run 'make clean-e2e-resources'"
 
+.PHONY: test-e2e-without-migration
+## Run the e2e tests with migration tests
+test-e2e-without-migration: prepare-e2e deploy-e2e e2e-run
+	@echo "To clean the cluster run 'make clean-e2e-resources'"
+
+.PHONY: cover-migration-and-deploy
+cover-migration-and-deploy: prepare-projects e2e-deploy-latest e2e-service-account setup-toolchainclusters e2e-migration-first-run get-publish-and-install-operators e2e-migration-second-run
+
+.PHONY: e2e-migration-first-run
+e2e-migration-first-run:
+	@echo "Running before migration tests to setup the environment..."
+	$(MAKE) execute-tests MEMBER_NS=${MEMBER_NS} MEMBER_NS_2=${MEMBER_NS_2} HOST_NS=${HOST_NS} REGISTRATION_SERVICE_NS=${REGISTRATION_SERVICE_NS} TESTS_TO_EXECUTE="./test/migration/setup"
+	@echo "Before migration tests successfuly finished."
+
+.PHONY: e2e-migration-second-run
+e2e-migration-second-run:
+	@echo "Running after migration tests to verify that the migration was successful..."
+	$(MAKE) execute-tests MEMBER_NS=${MEMBER_NS} MEMBER_NS_2=${MEMBER_NS_2} HOST_NS=${HOST_NS} REGISTRATION_SERVICE_NS=${REGISTRATION_SERVICE_NS} TESTS_TO_EXECUTE="./test/migration/verify"
+	@echo "After migration tests successfuly finished"
+
+.PHONY: e2e-deploy-latest
+e2e-deploy-latest:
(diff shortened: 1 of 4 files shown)


make/test.mk
```diff
@@ -32,13 +32,41 @@ SETUP_E2E_SERVICE_ACCOUNTS ?= true
...
+test-e2e: prepare-e2e cover-migration-and-deploy e2e-run
 	@echo "The tests successfully finished"
 	@echo "To clean the cluster run 'make clean-e2e-resources'"
 
+.PHONY: test-e2e-without-migration
+## Run the e2e tests with migration tests
+test-e2e-without-migration: prepare-e2e deploy-e2e e2e-run
+	@echo "To clean the cluster run 'make clean-e2e-resources'"
+
+.PHONY: cover-migration-and-deploy
```
xcoulon, line 45: "cover" may be a bit misleading in the context of tests: it may suggest that there will be some code coverage (which is not the case, AFAICS)

make/test.mk
```diff
@@ -32,13 +32,41 @@ SETUP_E2E_SERVICE_ACCOUNTS ?= true
...
+cover-migration-and-deploy: prepare-projects e2e-deploy-latest e2e-service-account setup-toolchainclusters e2e-migration-first-run get-publish-and-install-operators e2e-migration-second-run
+
+.PHONY: e2e-migration-first-run
+e2e-migration-first-run:
+	@echo "Running before migration tests to setup the environment..."
```
xcoulon, line 50:
```suggestion
	@echo "Setting up the environment before testing the operator migration..."
```

make/test.mk
```diff
@@ -32,13 +32,41 @@ SETUP_E2E_SERVICE_ACCOUNTS ?= true
...
+	$(MAKE) execute-tests MEMBER_NS=${MEMBER_NS} MEMBER_NS_2=${MEMBER_NS_2} HOST_NS=${HOST_NS} REGISTRATION_SERVICE_NS=${REGISTRATION_SERVICE_NS} TESTS_TO_EXECUTE="./test/migration/setup"
+	@echo "Before migration tests successfuly finished."
```
xcoulon, line 52:
```suggestion
	@echo "Environment successfully setup."
```

make/test.mk
```diff
@@ -32,13 +32,41 @@ SETUP_E2E_SERVICE_ACCOUNTS ?= true
...
+
+.PHONY: e2e-migration-second-run
+e2e-migration-second-run:
+	@echo "Running after migration tests to verify that the migration was successful..."
```
xcoulon, line 56:
```suggestion
	@echo "Updating operators and verifying resources..."
```

make/test.mk
```diff
@@ -32,13 +32,41 @@ SETUP_E2E_SERVICE_ACCOUNTS ?= true
...
+	$(MAKE) execute-tests MEMBER_NS=${MEMBER_NS} MEMBER_NS_2=${MEMBER_NS_2} HOST_NS=${HOST_NS} REGISTRATION_SERVICE_NS=${REGISTRATION_SERVICE_NS} TESTS_TO_EXECUTE="./test/migration/verify"
+	@echo "After migration tests successfuly finished"
```
xcoulon, line 58:
```suggestion
	@echo "Migration tests successfully finished"
```

test/migration/verify/verify_migration_test.go
```diff
@@ -0,0 +1,161 @@
...
+			run()
+		}(funcToRun)
+	}
+
+	wg.Wait()
+
+	cleanup.ExecuteAllCleanTasks()
+
+	t.Run("run migration setup", func(t *testing.T) {
+		// We need to run the migration setup part to be sure that when the PR is merged
```
xcoulon, line 59: I appreciate the explanation but I don't understand it: how does running this subtest will help with upcoming PRs? On OpenShift CI, the next PRs will be tested on their own/isolated clusters, so what's the point of setting up migration resources here? 
I believe I'm totally missing the point, though :/

MatousJobanek: Consider this:
1. there is running version of host/member-operator that populates the `UserAccount.Spec.NSTemplateSet` field
2. the `migration setup logic` checks and expects that this field is populated
3. there is a migration PR that drops usage of that `UserAccount.Spec.NSTemplateSet` field so it won't be populated any more
4. there is also paired e2e PR that modifies the `verify migration test` so it doesn't expect the field to be populated
5. the tests are green - so far so good.

but what would happen when the PRs are merged?
The next PR would fail in the `migration setup logic` because it would still contain the logic that would check and expect `UserAccount.Spec.NSTemplateSet` to be populated.

xcoulon: ok, I see. In that case, the next PR will have to update the migration setup logic so that the field is not populated anymore. Or even more radical, it could be a PR with an API update where the field is removed: since the previous PR was already merged, we can assume it was deployed and that the migration took place, so we should be able to move on with the removal of the deprecated field.

MatousJobanek: > In that case, the next PR will have to update the migration setup logic so that the field is not populated anymore.

but who should do that? imagine that in the meantime there can be multiple other PRs and for all of them the e2e tests would start failing. In addition, it would fail also for PRs in other repositories (host, member, reg-serv).
They would all need to wait until the migration setup logic is fixed. I believe that making the logic compatible with both versions is a much easier solution

> it could be a PR with an API update where the field is removed

removal of the deprecated field is a different story. I was talking about the logic that populates the field - so the content matters, not the presence.

Anyway, that was just an example. That use-case could fit to any situation where the setup part relies on a functionality/feature that is present in the current version of operators but won't be present in the next one.
So, to make our lives easier, let's assume that everything that is merged in master works as expected. This means that we can "just" create either `UserSignup` or `Space` and wait for the provisioned status (or any other status/state that signs that the process is done). We don't have to verify the actual content of the resources, nor labels, etc... because it was already verified when the PR was merged. 
If we agree on such a generic setup logic, then we can easily make sure that it's fully compatible with both versions of Dev Sandbox.

make/test.mk
MatousJobanek: is this better? 8a83801

test/migration/verify/verify_migration_test.go
```diff
@@ -0,0 +1,161 @@
...
+func verifyBannedSignup(t *testing.T, awaitilities wait.Awaitilities, signup *toolchainv1alpha1.UserSignup) {
+	hostAwait := awaitilities.Host()
+	cleanup.AddCleanTasks(t, hostAwait.Client, signup)
+
+	// verify that it's still banned
+	_, err := hostAwait.WaitForUserSignup(signup.Name,
+		wait.UntilUserSignupHasConditions(ConditionSet(Default(), ApprovedByAdmin(), Banned())...),
+		wait.UntilUserSignupHasStateLabel(toolchainv1alpha1.UserSignupStateLabelValueBanned))
+	require.NoError(t, err)
+	err = hostAwait.WaitUntilMasterUserRecordDeleted(migration.DeactivatedUser)
```
rajivnathan, line 125:
```suggestion
	err = hostAwait.WaitUntilMasterUserRecordDeleted(migration.BannedUser)
```

rajivnathan (commented):
Thanks for this! Will help me a lot with the MUR-Space migration 🙌

test/migration/verify/verify_migration_test.go
```diff
@@ -0,0 +1,161 @@
...
+			defer wg.Done()
+			run()
+		}(funcToRun)
+	}
+
+	wg.Wait()
+
+	cleanup.ExecuteAllCleanTasks()
+
+	t.Run("run migration setup", func(t *testing.T) {
```
rajivnathan, line 58: What about changing this test name to this? WDYT?
```suggestion
	t.Run("run migration setup with new operator versions for compatibility", func(t *testing.T) {
```

test/migration/verify/verify_migration_test.go
rajivnathan: I missed the point of this until I saw this discussion too. I guess that the other options (making the next PR update the migration logic) would lead to other problems that sound worse than keeping it this way. So my vote would be to keep this approach unless there's a better way. If we decide to keep it this way then maybe we should add Matous' explanation to the comment since it wasn't obvious to two people.

alexeykazakov: The current approach seems to be reasonable to me too but I agree with Rajiv that it's better to add more details to the comment with the reasoning. Some generic example (similar to what you did here in this discussion) would be the best way of explaining the need.

test/migration/verify/verify_migration_test.go
```diff
@@ -0,0 +1,161 @@
...
+func TestAfterMigration(t *testing.T) {
+	// given
+	awaitilities := WaitForDeployments(t)
+
+	// get Signups for the users provisioned in the setup part
+	provisionedSignup := getSignupFromMUR(t, awaitilities.Host(), migration.ProvisionedUser)
+	secondMemberProvisionedSignup := getSignupFromMUR(t, awaitilities.Host(), migration.SecondMemberProvisionedUser)
+	appstudioProvisionedSignup := getSignupFromMUR(t, awaitilities.Host(), migration.AppStudioProvisionedUser)
+	deactivatedSignup := listAndGetSignupWithState(t, awaitilities.Host(), toolchainv1alpha1.UserSignupStateLabelValueDeactivated)
+	bannedSignup := listAndGetSignupWithState(t, awaitilities.Host(), toolchainv1alpha1.UserSignupStateLabelValueBanned)
```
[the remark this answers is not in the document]
MatousJobanek: No, we cannot do that, because:
* we don't know the names of the UserSignups
* we know only the names of the MURs, but there is one account that is deactivated and one that is banned, so we cannot use simple get command - we have to list them using the `state` label
* the functions are executed in parallel and some of them also bans or deactivates the UserSignups to test that everything works fine, so the list could return more UserSignups than only one

Does it makes sense?

test/migration/verify/verify_migration_test.go
MatousJobanek: copy-paste :man_facepalming: Thanks :+1:

make/test.mk
```diff
@@ -32,13 +32,41 @@ SETUP_E2E_SERVICE_ACCOUNTS ?= true
...
+.PHONY: test-e2e-without-migration
+## Run the e2e tests with migration tests
+test-e2e-without-migration: prepare-e2e deploy-e2e e2e-run
+	@echo "To clean the cluster run 'make clean-e2e-resources'"
+
+.PHONY: verify-migration-and-deploy-e2e
+verify-migration-and-deploy-e2e: prepare-projects e2e-deploy-latest e2e-service-account setup-toolchainclusters e2e-migration-first-run get-publish-and-install-operators e2e-migration-second-run
+
+.PHONY: e2e-migration-first-run
+e2e-migration-first-run:
```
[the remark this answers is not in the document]
MatousJobanek: right :smile: - this was the first name that came to my mind when I started hacking the things but forgot to rename it :man_facepalming: 
renamed in d5e7317

make/test.mk
```diff
@@ -32,13 +32,41 @@ SETUP_E2E_SERVICE_ACCOUNTS ?= true
...
+	@echo "Setting up the environment before testing the operator migration..."
+	$(MAKE) execute-tests MEMBER_NS=${MEMBER_NS} MEMBER_NS_2=${MEMBER_NS_2} HOST_NS=${HOST_NS} REGISTRATION_SERVICE_NS=${REGISTRATION_SERVICE_NS} TESTS_TO_EXECUTE="./test/migration/setup"
+	@echo "Environment successfully setup."
+
+.PHONY: e2e-migration-second-run
+e2e-migration-second-run:
```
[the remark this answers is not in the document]
MatousJobanek: renamed d5e7317

test/migration/verify/verify_migration_test.go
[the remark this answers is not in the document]
rajivnathan: Sorry, I'm still a bit confused as to why a function like this wouldn't work:
```
func verifyProvisionedSignup(t *testing.T, awaitilities wait.Awaitilities) {
	signup := getSignupFromMUR(t, awaitilities.Host(), migration.ProvisionedUser)
	cleanup.AddCleanTasks(t, awaitilities.Host().Client, signup)
	VerifyResourcesProvisionedForSignup(t, awaitilities, signup, "base")
	DeactivateAndCheckUser(t, awaitilities, signup)
	ReactivateAndCheckUser(t, awaitilities, signup)
}
```

It would still get the UserSignup using the MUR name.

For the deactivated/banned ones we could also do it within the verify function?
eg.
```
func verifyDeactivatedSignup(t *testing.T, awaitilities wait.Awaitilities) {
	signup := listAndGetSignupWithState(t, awaitilities.Host(), toolchainv1alpha1.UserSignupStateLabelValueDeactivated)
	cleanup.AddCleanTasks(t, awaitilities.Host().Client, signup)

	_, err := awaitilities.Host().WaitForUserSignup(signup.Name,
		wait.UntilUserSignupContainsConditions(ConditionSet(Default(), DeactivatedWithoutPreDeactivation())...),
		wait.UntilUserSignupHasStateLabel(toolchainv1alpha1.UserSignupStateLabelValueDeactivated))
	require.NoError(t, err)
	require.True(t, states.Deactivated(signup), "usersignup should be deactivated")

	err = awaitilities.Host().WaitUntilMasterUserRecordDeleted(migration.DeactivatedUser)
	require.NoError(t, err)

	ReactivateAndCheckUser(t, awaitilities, signup)
}
```

But what I just realized is that with the current way we're following the given, when, then testing pattern so I guess it makes sense to get the usersignups as part of the 'given' phase of the test. So you can disregard my comment.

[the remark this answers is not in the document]
rajivnathan: Actually, we could still follow the given, when, then pattern if we do:
```
func verifyProvisionedSignup(t *testing.T, awaitilities wait.Awaitilities) {
	//given
	signup := getSignupFromMUR(t, awaitilities.Host(), migration.ProvisionedUser)

	// when & then
	cleanup.AddCleanTasks(t, awaitilities.Host().Client, signup)
	VerifyResourcesProvisionedForSignup(t, awaitilities, signup, "base")
	DeactivateAndCheckUser(t, awaitilities, signup)
	ReactivateAndCheckUser(t, awaitilities, signup)
}
```

rajivnathan (approved):
Super cool, this will help me a lot! Thank you! 👍

test/migration/verify/verify_migration_test.go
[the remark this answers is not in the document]
MatousJobanek: so let's take the two functions that you mentioned as an example:
* `verifyProvisionedSignup`
* `verifyDeactivatedSignup`

these functions are executed in parallel. 

Now let's take a look at how the `listAndGetSignupWithState` function (which would be used in `verifyDeactivatedSignup`):
https://github.com/codeready-toolchain/toolchain-e2e/blob/8a83801fe2f3f14f29fd33ed199e6b6335ed2fca/test/migration/verify/verify_migration_test.go#L154-L161
it lists all `UserSignups` with the label `state=deactivated` and expects that there is only one `UserSignup` found.

However, the `verifyProvisionedSignup` - after checking that the `ProvisionedUser` is provisioned - also deactivates the `UserSignup`. 
If the deactivation happens before the `verifyDeactivatedSignup` calls `listAndGetSignupWithState`, then it could get two deactivated `UserSignups` instead of one. 

I know that this is really a corner case that should never happen, but still - I want to eliminate any flakyness.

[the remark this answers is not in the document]
rajivnathan: Ahh, ok now I see what you mean. Ok, that makes sense and I agree it's better to keep it there. We could probably add a little more to the comment to point out why it's better to have it there.

test/migration/verify/verify_migration_test.go
```diff
@@ -0,0 +1,161 @@
...
+
+func TestAfterMigration(t *testing.T) {
+	// given
+	awaitilities := WaitForDeployments(t)
+
+	// get Signups for the users provisioned in the setup part
+	provisionedSignup := getSignupFromMUR(t, awaitilities.Host(), migration.ProvisionedUser)
+	secondMemberProvisionedSignup := getSignupFromMUR(t, awaitilities.Host(), migration.SecondMemberProvisionedUser)
+	appstudioProvisionedSignup := getSignupFromMUR(t, awaitilities.Host(), migration.AppStudioProvisionedUser)
+	deactivatedSignup := listAndGetSignupWithState(t, awaitilities.Host(), toolchainv1alpha1.UserSignupStateLabelValueDeactivated)
```
rajivnathan, line 27:
```suggestion
	// note: listing banned/deactivated UserSignups should be done as part of setup because the tests are run in parallel and there can be multiple banned/deactivated UserSignups at that point which could lead to test flakiness
	deactivatedSignup := listAndGetSignupWithState(t, awaitilities.Host(), toolchainv1alpha1.UserSignupStateLabelValueDeactivated)
```

test/migration/verify/verify_migration_test.go
MatousJobanek: I updated the comment here 3046308
Feel free to improve/fix/change it if needed  :wink:

MatousJobanek:
@xcoulon @rajivnathan @alexeykazakov thanks a lot for your reviews and comments. I addressed all your comments/requests last week. I'll go ahead and merge it, so we can start using it. If anyone has any additional comments, then we can address them in a separate PR.

PR #451, opened 2022-01-19
Update services quota to 30

rajivnathan:
Related PR: https://github.com/codeready-toolchain/host-operator/pull/574

PR #452, opened 2022-01-20
Get the latest copy of space before updating

rajivnathan:
Addresses https://github.com/codeready-toolchain/toolchain-e2e/pull/451#issuecomment-1016797487

### DIFF
--- a/test/e2e/space_test.go
+++ b/test/e2e/space_test.go
@@ -11,7 +11,6 @@ import (
 	"github.com/codeready-toolchain/toolchain-e2e/testsupport/wait"
 	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/types"
 )
 
 func TestCreateSpace(t *testing.T) {
@@ -174,20 +173,10 @@ func TestRetargetSpace(t *testing.T) {
 		// wait until Space has been provisioned on member-1
 		VerifyResourcesProvisionedForSpaceWithTier(t, awaitilities, member1Await, space.Name, "base")
 
-		// retrieve the latest copy of the Space before updating
-		spaceName := space.Name
-		space = &toolchainv1alpha1.Space{}
-		err = hostAwait.Client.Get(context.TODO(),
-			types.NamespacedName{
-				Namespace: hostAwait.Namespace,
-				Name:      spaceName,
-			},
-			space)
-		require.NoError(t, err)
-
 		// when
-		space.Spec.TargetCluster = ""
-		err = hostAwait.Client.Update(context.TODO(), space)
+		space, err = hostAwait.UpdateSpace(space.Name, func(s *toolchainv1alpha1.Space) {
+			s.Spec.TargetCluster = ""
+		})
 		require.NoError(t, err)
 
 		// then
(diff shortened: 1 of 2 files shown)


test/e2e/space_test.go
```diff
@@ -171,7 +172,17 @@ func TestRetargetSpace(t *testing.T) {
 		err := hostAwait.CreateWithCleanup(context.TODO(), space)
 		require.NoError(t, err)
 		// wait until Space has been provisioned on member-1
-		space = VerifyResourcesProvisionedForSpaceWithTier(t, awaitilities, member1Await, space.Name, "base")
+		VerifyResourcesProvisionedForSpaceWithTier(t, awaitilities, member1Await, space.Name, "base")
```
alexeykazakov, line 175: Shouldn't `VerifyResourcesProvisionedForSpaceWithTier()` return the latest copy of the space instead?

test/e2e/space_test.go
```diff
@@ -171,7 +172,18 @@ func TestRetargetSpace(t *testing.T) {
...
+		// retrieve the latest copy of the Space before updating
+		spaceName := space.Name
+		space = &toolchainv1alpha1.Space{}
+		err = hostAwait.Client.Get(context.TODO(),
+			types.NamespacedName{
+				Namespace: hostAwait.Namespace,
+				Name:      spaceName,
+			},
+			space)
+		require.NoError(t, err)
```
MatousJobanek, line 186: Even this code could fail for the conflict and it's impossible to reuse it somewhere else. 
Could you please use the same approach as we use for UserSignup updates? https://github.com/codeready-toolchain/toolchain-e2e/blob/master/testsupport/wait/host.go#L227-L244

testsupport/wait/host.go
```diff
@@ -243,6 +243,27 @@ func (a *HostAwaitility) UpdateUserSignup(userSignupName string, modifyUserSignu
 	return userSignup, err
 }
 
+// UpdateUserSignup tries to update the Spec of the given UserSignup
+// If it fails with an error (for example if the object has been modified) then it retrieves the latest version and tries again
+// Returns the updated UserSignup
```
MatousJobanek, line 248: copy-paste driven development :smile: 
```suggestion
// UpdateSpace tries to update the Spec of the given Space
// If it fails with an error (for example if the object has been modified) then it retrieves the latest version and tries again
// Returns the updated Space
```

MatousJobanek (approved):

testsupport/wait/host.go
xcoulon: ah damn'it! I thought I changed all occurrences of `UserSignup` 🤦‍♂️

test/e2e/space_test.go
rajivnathan: It could. I thought about that but it's not really necessary for `VerifyResourcesProvisionedForSpaceWithTier` itself. If we're not doing an update after calling `VerifyResourcesProvisionedForSpaceWithTier` then the extra get would not be useful. So I thought it made more sense to do a get before update instead.

test/e2e/space_test.go
rajivnathan: oops, forgot about that. Thanks for handling it

PR #453, opened 2022-01-20
chore: verify that SpaceBindingCleanup controller works

MatousJobanek:
verifies that SpaceBinding is deleted when either bound Space or bound MUR is deleted 

paired with: https://github.com/codeready-toolchain/host-operator/pull/575


rajivnathan (approved):
Nice and simple 👍

PR #455, opened 2022-01-24
chore: update linter config

xcoulon:
disable `golint`, use `revive` instead, and fix errors reported by
the latter, plus a `gosimple` suggestion, too.


testsupport/wait/host.go
```diff
@@ -965,7 +965,8 @@ func (a *HostAwaitility) printNotificationWaitCriterionDiffs(actual []toolchainv
 	} else {
 		buf.WriteString("failed to find notifications with matching criteria:\n")
 		buf.WriteString("actual:\n")
-		for _, obj := range actual {
+		for i := range actual {
+			obj := actual[i] // avoids the 'G601: Implicit memory aliasing in for loop. (gosec)' problem
 			y, _ := StringifyObject(&obj)
```
MatousJobanek, line 970: is this really needed for this particular use-case - we should always get the correct values even with the original version...

testsupport/wait/member.go
```diff
@@ -915,7 +915,8 @@ func (a *MemberAwaitility) WaitForPods(namespace string, n int, criteria ...PodW
 			return false, err
 		}
 	pods:
-		for _, p := range foundPods.Items {
+		for i := range foundPods.Items {
+			p := foundPods.Items[i] // avoids the 'G601: Implicit memory aliasing in for loop. (gosec)' problem
 			if !matchPodWaitCriterion(&p, criteria...) {
```
MatousJobanek, line 920: the same question - there shouldn't be a problem with the original version, should it?

testsupport/wait/awaitility.go
```diff
@@ -384,7 +386,8 @@ func (a *Awaitility) DeletePods(criteria ...client.ListOption) error {
 	if err != nil {
 		return err
 	}
-	for _, p := range pods.Items {
+	for i := range pods.Items {
+		p := pods.Items[i] // avoids the 'G601: Implicit memory aliasing in for loop. (gosec)' problem
 		if err := a.Client.Delete(context.TODO(), &p); err != nil {
```
MatousJobanek, line 391: the same for this

testsupport/wait/host.go
xcoulon: reverted in bc57e8a, using a `nolint:gosec` commend instead

testsupport/wait/member.go
xcoulon: reverted in bc57e8a, using a `nolint:gosec` commend instead

testsupport/wait/awaitility.go
xcoulon: reverted in bc57e8a, using a `nolint:gosec` commend instead

testsupport/metrics/metrics.go
```diff
@@ -24,7 +24,7 @@ func GetMetricValue(restConfig *rest.Config, url string, family string, expected
 	client := http.Client{
 		Timeout: time.Duration(10 * time.Second),
 		Transport: &http.Transport{
-			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true}, //nolint:gosec
```
MatousJobanek, line 27: just out of curiosity - what kind of error do we ignore here?

testsupport/tier_setup.go
```diff
@@ -7,7 +7,7 @@ import (
 
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
 	"github.com/codeready-toolchain/toolchain-common/pkg/test"
-	. "github.com/codeready-toolchain/toolchain-e2e/testsupport/wait" // nolint: golint
+	. "github.com/codeready-toolchain/toolchain-e2e/testsupport/wait" // nolint:revive
```
MatousJobanek, line 10: what is ignored here?

MatousJobanek (approved):

testsupport/metrics/metrics.go
xcoulon: `gosec` complains if we explicitly skip cert verification

testsupport/tier_setup.go
xcoulon: dot imports are not recommended. I updated the `nolint` comment since the `golint` linter is now deprecated (and disabled) and replaced by `revive`

PR #456, opened 2022-01-25
test: verify limited appstudio user permissions

MatousJobanek:
verifies https://github.com/codeready-toolchain/host-operator/pull/579


alexeykazakov (approved):

testsupport/has/api/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -0,0 +1,123 @@
+// +build !ignore_autogenerated
```
rajivnathan, line 1: Should we include a make target to regenerate CRD-related files for the has api or you think it's not necessary?

rajivnathan (commented):
Cool, just had one question

testsupport/has/api/v1alpha1/zz_generated.deepcopy.go
MatousJobanek: I don't think that it's needed - this is only as a dummy CRD, we don't need to update it or keep in sync with the real API type / CRD

rajivnathan: Ok, that's what I figured but just thought I'd ask in case we were planning to update it.

testsupport/has/api/v1alpha1/application_types.go
```diff
@@ -0,0 +1,91 @@
...
+limitations under the License.
+*/
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ApplicationSpec defines the desired state of Application
```
alexeykazakov, line 23: Let's mention that it's a dummy CRD we use only for tests, etc. I would also mention it in the test itself.
So no one get confused about e2e owning an Application API and relationship with the actual Application API from HAS.

alexeykazakov (approved):
Looks good! But let's add some comments in the code explaining the purpose of the Application CRD.

testsupport/has/api/v1alpha1/application_types.go
MatousJobanek: added comments in 3316925

alexeykazakov: Good, descriptive comments. Thank you!

PR #457, opened 2022-01-25
feat: use env var to decide if it should use local scripts

MatousJobanek:
I know that the current way the toolchain-cicd scripts are executed is not the best and most transparent. So, I added a var that determines if the scripts should be taken from the local repo or from the remote one. By default, it will take the scripts from the remote repo, which means that it will run the latest version of them.

if we agree on this approach, then I'll change that also in the other repositories.


alexeykazakov (approved):
Looks good to me

xcoulon (approved):

rajivnathan (approved):
It's a good change 👍

PR #459, opened 2022-01-26
Update setup tool to install Gitops primer operator

rajivnathan:
- Updates setup tool to instlal the Gitops primer operator
- Updates README to mention using a fresh cluster when capturing performance metrics for comparison/onboarding


alexeykazakov (approved):
We also need to update our baseline.

MatousJobanek (approved):

PR #461, opened 2022-01-26
test: update space role templateref

xcoulon:
verify that space roles and rolebindings are updated
as expected when the templateref is changed in the
`NSTemplateSet.Spec.SpaceRoles` field

Fixes https://issues.redhat.com/browse/CRT-1339


test/e2e/space_test.go
```diff
@@ -241,6 +242,142 @@ func TestSpaceRoles(t *testing.T) {
...
+				assert.NoError(t, err)
+				err = memberAwait.WaitUntilRoleBindingDeleted(ns, "user1-space-admin")
+				assert.NoError(t, err)
+				err = memberAwait.WaitUntilRoleBindingDeleted(ns, "user2-space-admin")
+				assert.NoError(t, err)
+
+			})
+		})
+
+		t.Run("with same content", func(t *testing.T) {
```
MatousJobanek, line 318: I don't think that we need this test here - it doesn't test anything extra which wouldn't be covered in the previous test. 
But I'll leave it up to you ;-)

MatousJobanek (approved):
looks good :+1:

test/e2e/space_test.go
xcoulon: yeah, probably, but I'd rather keep it for now. We can remove it later if it causes too much maintenance

MatousJobanek: it's not only about the maintenance, but also the execution time :wink:

xcoulon: right :/ let me remove it, then

PR #470, opened 2022-02-03
Provide current user namespace value that can be used in templates

rajivnathan:
Provide user namespace parameter when processing user workload templates in the setup tool.

It can be included in the templates like this:
```
parameters:
- name: CURRENT_USER_NAMESPACE
  required: true
```

### DIFF
--- a/setup/resources/create_resources.go
+++ b/setup/resources/create_resources.go
@@ -13,6 +13,8 @@ import (
 	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const userNSParam = "CURRENT_USER_NAMESPACE"
+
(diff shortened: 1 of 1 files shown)


alexeykazakov (approved):

MatousJobanek:
@rajivnathan just out of curiosity - what do we need this for?

MatousJobanek:
never mind, after having my morning coffee and reading the [slack discussion](https://coreos.slack.com/archives/CHK0J6HT6/p1643905474259269), I understand the usage :-)

jeremyary:
Thank you for implementing this! We (RHODA) will definitely be leveraging right away. ^_^

PR #475, opened 2022-02-10
Test migration from MUR to Spaces

rajivnathan:
Related PR: https://github.com/codeready-toolchain/host-operator/pull/595

Summary of changes:
- Added MUR migration checks to `TestAfterMigration` test
- Fix for `TestE2EFlow` test because useraccount deletion happens very quickly now that the UserAccount deletion does not depend on deletion of the NSTemplateSet and related resources. Instead of waiting for the useraccount to have a deletion timestamp which we may miss, we will ensure the creation timestamp is updated to a newer timestamp than was there previously
- Removed useraccount checks for NSLimit and NSTemplateSet
- Added `tierNameChanged bool` as a parameter to the `verifyResourceUpdatesForUserSignups` func so that the syncIndex can be checked accordingly (it's now expected to change only when tierName is changed, previously cluster and namespace template refs mattered)
- Set `DefaultSpaceTier` config where necessary
- Other minor fixes


MatousJobanek (commented):
Nice :+1: 
Btw, we should also verify the case when the `skip-auto-create-space` annotation is set to `true`. It can be done in separate PR - it's up to you.

test/e2e/nstemplatetier_test.go
```diff
@@ -358,27 +361,26 @@ func verifyResourceUpdatesForUserSignups(t *testing.T, hostAwait *HostAwaitility
 			}
 			require.NoError(t, err, "Failing \nUserSignup: %+v \nUserAccount: %+v \nNSTemplateSet: %+v", usersignup, userAccount, nsTemplateSet)
 		}
+
+		// the syncIndex should be different if the tier changed. if only the template refs changed then the sync index is expected to be the same because UserAccounts no longer have references to the NSTemplateSet
 		mur, err := hostAwait.WaitForMasterUserRecord(usersignup.Status.CompliantUsername,
 			UntilMasterUserRecordHasCondition(Provisioned()), // ignore other conditions, such as notification sent, etc.
-			UntilMasterUserRecordHasNotSyncIndex(syncIndex),
+			UntilMasterUserRecordHasSyncIndex(syncIndex, tierNameChanged),
```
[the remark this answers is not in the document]
rajivnathan: I figured since we still have the sync logic then it doesn't hurt keeping it just to make sure it's behaving as expected until we remove it but if you prefer to stop checking it entirely then I'm fine doing that too. For these tests we aren't changing the tier name, it's just updates to the templates so it's probably not necessary to check that either?

test/e2e/user_management_test.go
```diff
@@ -125,10 +125,11 @@ func (s *userManagementTestSuite) TestUserDeactivation() {
 		require.NoError(t, err)
 
 		// Move the user to the new tier without deactivation enabled
-		murSyncIndex := MoveUserToTier(t, hostAwait, userSignupMember1.Spec.Username, baseDeactivationDisabledTier.Name).Spec.UserAccounts[0].SyncIndex
+		murSyncIndex := murMember1.Spec.UserAccounts[0].SyncIndex
+		MoveUserToTier(t, hostAwait, userSignupMember1.Spec.Username, baseDeactivationDisabledTier.Name)
 		murMember1, err = hostAwait.WaitForMasterUserRecord(murMember1.Name,
 			wait.UntilMasterUserRecordHasCondition(Provisioned()), // ignore other conditions, such as notification sent, etc.
-			wait.UntilMasterUserRecordHasNotSyncIndex(murSyncIndex))
+			wait.UntilMasterUserRecordHasSyncIndex(murSyncIndex, true))
```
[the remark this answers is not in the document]
rajivnathan: Changed it to `wait.UntilMasterUserRecordHasTierName(baseDeactivationDisabledTier.Name)` in 60b80dfbd2aa1c56ee9699ae12af34879782c9d0

test/migration/verify/verify_migration_test.go
```diff
@@ -162,15 +166,46 @@ func verifyBannedSignup(t *testing.T, awaitilities wait.Awaitilities, signup *to
...
+	hash, err := testtier.ComputeTemplateRefsHash(tier) // we can assume the JSON marshalling will always work
+	require.NoError(t, err)
+
+	// space should be assigned and provisioned with correct tier name
+	_, err = hostAwait.WaitForSpace(mur.Name,
+		wait.UntilSpaceHasTier(mur.Spec.TierName),
+		wait.UntilSpaceHasLabelWithValue(fmt.Sprintf("toolchain.dev.openshift.com/%s-tier-hash", mur.Spec.TierName), hash),
+		wait.UntilSpaceHasConditions(Provisioned()),
+		wait.UntilSpaceHasStateLabel(toolchainv1alpha1.SpaceStateLabelValueClusterAssigned),
+		wait.UntilSpaceHasStatusTargetCluster(mur.Spec.UserAccounts[0].TargetCluster))
```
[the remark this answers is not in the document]
rajivnathan: Done in ddb9d357512b337dc776b8f2f6c132d0ab9aa4d3

testsupport/user_assertions.go
```diff
@@ -102,6 +102,19 @@ func VerifyResourcesProvisionedForSignup(t *testing.T, awaitilities wait.Awaitil
...
+	require.NoError(t, err)
+	hash, err := testtier.ComputeTemplateRefsHash(tier) // we can assume the JSON marshalling will always work
+	require.NoError(t, err)
+
+	_, err = hostAwait.WaitForSpace(mur.Name,
+		wait.UntilSpaceHasTier(mur.Spec.TierName),
+		wait.UntilSpaceHasLabelWithValue(fmt.Sprintf("toolchain.dev.openshift.com/%s-tier-hash", mur.Spec.TierName), hash),
+		wait.UntilSpaceHasConditions(Provisioned()),
+		wait.UntilSpaceHasStateLabel(toolchainv1alpha1.SpaceStateLabelValueClusterAssigned),
+		wait.UntilSpaceHasStatusTargetCluster(mur.Spec.UserAccounts[0].TargetCluster))
```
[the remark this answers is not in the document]
rajivnathan: While I was trying to add it to `VerifyResourcesProvisionedForSpaceWithTiers` I wondered whether UserSignup would always be necessary. For some of the space tests we have now, they are not  Or did you mean that just checking that the creator label is set and not empty would be good enough?

testsupport/wait/host.go
```diff
@@ -1588,7 +1614,7 @@ func UntilSpaceHasStatusTargetCluster(expected string) SpaceWaitCriterion {
 func (a *HostAwaitility) WaitUntilSpaceDeleted(name string) error {
 	a.T.Logf("waiting until Space '%s' in namespace '%s' is deleted", name, a.Namespace)
 	var s *toolchainv1alpha1.Space
-	err := wait.Poll(a.RetryInterval, a.Timeout, func() (done bool, err error) {
+	err := wait.Poll(a.RetryInterval, 2*a.Timeout, func() (done bool, err error) {
```
MatousJobanek, line 1617: Do we really need to double the timeout?
```suggestion
	err := wait.Poll(a.RetryInterval, a.Timeout, func() (done bool, err error) {
```

testsupport/wait/member.go
```diff
@@ -216,7 +217,7 @@ func UntilUserAccountContainsCondition(expected toolchainv1alpha1.Condition) Use
 }
 
 // UntilUserAccountIsBeingDeleted returns a `UserAccountWaitCriterion` which checks that the given
-// USerAccount has the deletion timestamp set
+// UserAccount has the deletion timestamp set
 func UntilUserAccountIsBeingDeleted() UserAccountWaitCriterion {
```
MatousJobanek, line 221: not used - can be deleted (can be done in the following PR :wink: )

testsupport/user_assertions.go
```diff
@@ -102,7 +102,23 @@ func VerifyResourcesProvisionedForSignup(t *testing.T, awaitilities wait.Awaitil
...
+		wait.UntilSpaceHasLabelWithValue(toolchainv1alpha1.SpaceCreatorLabelKey, userSignup.Name),
+		wait.UntilSpaceHasLabelWithValue(fmt.Sprintf("toolchain.dev.openshift.com/%s-tier-hash", mur.Spec.TierName), hash),
+		wait.UntilSpaceHasConditions(Provisioned()),
+		wait.UntilSpaceHasStateLabel(toolchainv1alpha1.SpaceStateLabelValueClusterAssigned),
+		wait.UntilSpaceHasStatusTargetCluster(mur.Spec.UserAccounts[0].TargetCluster))
+	require.NoError(t, err)
+
+	VerifySpaceBinding(t, hostAwait, mur.Name, space.Name, "admin")
+
+	tiers.VerifyNsTemplateSet(t, hostAwait, memberAwait, space, tierName)
```
MatousJobanek, line 121: it can be done in separate PR, but is there any difference between your code that the one that is executed in `VerifyResourcesProvisionedForSpace`?

MatousJobanek (commented):
Looks good - I have just a few comments/suggestion, but nothing critical. :+1: :1st_place_medal:

testsupport/wait/host.go
```diff
@@ -415,6 +416,22 @@ func UntilMasterUserRecordHasTierName(expected string) MasterUserRecordWaitCrite
 	}
 }
 
+func UntilMasterUserRecordHasNoTierHashLabel() MasterUserRecordWaitCriterion {
+	return MasterUserRecordWaitCriterion{
+		Match: func(actual *toolchainv1alpha1.MasterUserRecord) bool {
+			if actual.Labels == nil {
```
xcoulon, line 422: I believe this is not needed, `range` will not fail if `actual.Labels` is `nil`. (but it's a minor thing, not needed to be addressed in this PR)

test/e2e/nstemplatetier_test.go
```diff
@@ -334,11 +338,10 @@ func verifyStatus(t *testing.T, hostAwait *HostAwaitility, tierName string, expe
 	}
 }
 
-func verifyResourceUpdatesForUserSignups(t *testing.T, hostAwait *HostAwaitility, memberAwaitility *MemberAwaitility, syncIndexes map[string]string, tierName, aliasTierNamespaces, aliasTierClusterResources string) map[string]string {
-
-	templateRefs, namespacesChecks, clusterResourcesChecks := tiers.GetRefsAndChecksForTiers(t, hostAwait, tierName, aliasTierNamespaces, aliasTierClusterResources)
+func verifyResourceUpdatesForUserSignups(t *testing.T, awaitilities Awaitilities, memberAwaitility *MemberAwaitility, syncIndexes map[string]string, tierName, aliasTierNamespaces, aliasTierClusterResources string, tierNameChanged bool) map[string]string {
```
xcoulon, line 341: in most cases, we pass the `hostAwait *wait.HostAwaitility` arg instead of `awaitilities Awaitilities`, so I would keep it like that (also because `memberAwaitility *MemberAwaitility` can be obtained from `awaitilities Awaitilities` so there's a bit of duplication in the args). 
But it does not have to be changed in this PR, it's purely cosmetic stuff.

testsupport/wait/host.go
```diff
@@ -442,7 +459,7 @@ func (a *HostAwaitility) printUserSignupWaitCriterionDiffs(actual *toolchainv1al
 		buf.WriteString("\n----\n")
 		buf.WriteString("diffs:\n")
 		for _, c := range criteria {
-			if !c.Match(actual) {
+			if !c.Match(actual) && c.Diff != nil {
```
xcoulon, line 462: in which case do we have `c.Diff==nil`?

testsupport/wait/host.go
```diff
@@ -1625,3 +1651,107 @@ func (a *HostAwaitility) WaitUntilSpaceBindingDeleted(name string) error {
...
+
+// UntilSpaceBindingHasMurName returns a `SpaceBindingWaitCriterion` which checks that the given
+// SpaceBinding has the expected MUR name set in its Spec
+func UntilSpaceBindingHasMurName(expected string) SpaceBindingWaitCriterion {
+	return SpaceBindingWaitCriterion{
+		Match: func(actual *toolchainv1alpha1.SpaceBinding) bool {
+			return actual.Spec.MasterUserRecord == expected
+		},
+		Diff: func(actual *toolchainv1alpha1.SpaceBinding) string {
+			return fmt.Sprintf("expected tier name to match:\n%s", Diff(expected, actual.Spec.MasterUserRecord))
```
xcoulon, line 1728:
```suggestion
			return fmt.Sprintf("expected MUR name to match:\n%s", Diff(expected, actual.Spec.MasterUserRecord))
```

testsupport/wait/host.go
```diff
@@ -1625,3 +1651,107 @@ func (a *HostAwaitility) WaitUntilSpaceBindingDeleted(name string) error {
...
+
+// UntilSpaceBindingHasSpaceName returns a `SpaceBindingWaitCriterion` which checks that the given
+// SpaceBinding has the expected MUR name set in its Spec
+func UntilSpaceBindingHasSpaceName(expected string) SpaceBindingWaitCriterion {
+	return SpaceBindingWaitCriterion{
+		Match: func(actual *toolchainv1alpha1.SpaceBinding) bool {
+			return actual.Spec.Space == expected
+		},
+		Diff: func(actual *toolchainv1alpha1.SpaceBinding) string {
+			return fmt.Sprintf("expected tier name to match:\n%s", Diff(expected, actual.Spec.Space))
```
xcoulon, line 1741:
```suggestion
			return fmt.Sprintf("expected Space name to match:\n%s", Diff(expected, actual.Spec.Space))
```

testsupport/wait/host.go
```diff
@@ -1625,3 +1651,107 @@ func (a *HostAwaitility) WaitUntilSpaceBindingDeleted(name string) error {
...
+
+// UntilSpaceBindingHasSpaceRole returns a `SpaceBindingWaitCriterion` which checks that the given
+// SpaceBinding has the expected SpaceRole name set in its Spec
+func UntilSpaceBindingHasSpaceRole(expected string) SpaceBindingWaitCriterion {
+	return SpaceBindingWaitCriterion{
+		Match: func(actual *toolchainv1alpha1.SpaceBinding) bool {
+			return actual.Spec.SpaceRole == expected
+		},
+		Diff: func(actual *toolchainv1alpha1.SpaceBinding) string {
+			return fmt.Sprintf("expected tier name to match:\n%s", Diff(expected, actual.Spec.SpaceRole))
```
xcoulon, line 1754:
```suggestion
			return fmt.Sprintf("expected Space role to match:\n%s", Diff(expected, actual.Spec.SpaceRole))
```

testsupport/wait/member.go
```diff
@@ -108,7 +109,7 @@ func (a *MemberAwaitility) printUserAccountWaitCriterionDiffs(actual *toolchainv
 		buf.WriteString("\n----\n")
 		buf.WriteString("diffs:\n")
 		for _, c := range criteria {
-			if !c.Match(actual) {
+			if !c.Match(actual) && c.Diff != nil {
```
xcoulon, line 112: same question as above. Maybe we should add this nil-check in all other places?

xcoulon (approved):
Thanks for putting all the comments that explain the tests
Approved with a few questions that don't have to be addressed here, and some minor change suggestions.

testsupport/user_assertions.go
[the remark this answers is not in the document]
rajivnathan: Btw, I didn't add creator label check to `VerifyResourcesProvisionedForSpaceWithTiers` because `CreateAndVerifySpace` does not set that label.

test/e2e/e2e_test.go
```diff
@@ -267,8 +270,12 @@ func TestE2EFlow(t *testing.T) {
 
 	t.Run("verify userAccount is not deleted if namespace is not deleted", func(t *testing.T) {
```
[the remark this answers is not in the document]
rajivnathan: Good catch!

test/e2e/e2e_test.go
```diff
@@ -322,17 +329,28 @@ func TestE2EFlow(t *testing.T) {
 		require.NoError(t, err)
 		require.NotEmpty(t, nsTmplSet)
 
-		userAcc, err := memberAwait.WaitForUserAccount(laraUserName, wait.UntilUserAccountIsBeingDeleted(), wait.UntilUserAccountContainsCondition(TerminatingUserAccount()))
+		space, err := hostAwait.WaitForSpace(laraUserName, wait.UntilSpaceIsBeingDeleted(), wait.UntilSpaceHasConditions(TerminatingSpace()))
```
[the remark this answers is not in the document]
rajivnathan: Done in eb19f99895139f662387b32afe35bd206032d411

test/e2e/e2e_test.go
```diff
@@ -428,8 +440,11 @@ func TestE2EFlow(t *testing.T) {
 		err = memberAwait.WaitUntilIdentityDeleted(johnsmithName)
 		assert.NoError(t, err, "Identity is not deleted")
 
+		err = hostAwait.WaitUntilSpaceDeleted(johnsmithName)
+		require.NoError(t, err)
+
```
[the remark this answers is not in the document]
rajivnathan: In c3e952d1ef6abdb6bbfeb243ed54c795fd387e87 I modified the functions to automatically wait until the related SpaceBinding is deleted as well so we won't forget in the future 🙂

testsupport/wait/member.go
rajivnathan: Good catch! Noted, will handle in a separate PR.

testsupport/user_assertions.go
rajivnathan: Noted, will handle in a separate PR.

test/e2e/nstemplatetier_test.go
rajivnathan: Noted, will handle in a separate PR. Thanks!

testsupport/wait/host.go
rajivnathan: I figured it was good to increase the default in this case because the default is 60 seconds and we have to account for 2 things:
1. spacecleanupcontroller won't delete a Space until it is at least 30 seconds old
2. Space won't be deleted until NSTemplateSet and all the namespace resources are deleted

That said, maybe 1 minute is still enough. I'll change it back to 1 minute and if it turns out to not be enough time then maybe we can add another 30 seconds to make it 90 seconds instead of 2 minutes.

testsupport/wait/host.go
rajivnathan: You're right, good catch. Done in d279172377307fd298f562332a8e34f5b5910b0d

testsupport/wait/host.go
rajivnathan: I don't remember now but I only added it because I saw a panic in one of the test runs. I don't think it's always necessary for a criterion to have a Diff so I think it's okay to keep it? WDYT?

testsupport/wait/host.go
rajivnathan: Good catch! Thanks

testsupport/wait/member.go
rajivnathan: Yes, I had the same thought. I'll handle it in a separate PR. Noted.

test/e2e/user_management_test.go
```diff
@@ -195,6 +199,10 @@ func (s *userManagementTestSuite) TestUserDeactivation() {
 		// The non-excluded user should be deactivated
 		err = hostAwait.WaitUntilMasterUserRecordDeleted(murMember1.Name)
 		require.NoError(s.T(), err)
+
+		err = hostAwait.WaitUntilSpaceDeleted(murMember1.Name)
+		require.NoError(t, err)
```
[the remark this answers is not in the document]
rajivnathan: I was confused initially but I guess you meant this comment for the `tests for tiers with automatic deactivation disabled` test (lines 151-156) instead. It's a good suggestion though 👍  Done in 0c718180c962467d3fc48239d8573dd5344238fb

testsupport/wait/host.go
MatousJobanek: @xcoulon it looks like this is an example when it panics https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/codeready-toolchain_toolchain-e2e/475/pull-ci-codeready-toolchain-toolchain-e2e-master-e2e/1500820654616743936

MatousJobanek: @rajivnathan and I found the cause of the panic: https://github.com/codeready-toolchain/toolchain-e2e/blob/0c718180c962467d3fc48239d8573dd5344238fb/testsupport/wait/host.go#L419-L430 it doesn't set the Diff param

PR #486, opened 2022-02-25
refactor e2e tests

xcoulon:
remove combination of tiers in verification, use annotations on the
custom NSTemplateTiers to determine their "alias" and which `TierChecks`
to use to verify the resources.

also, use `corev1` alias instead of `v1`


test/e2e/nstemplatetier_test.go
```diff
@@ -153,54 +152,50 @@ func TestUpdateNSTemplateTier(t *testing.T) {
...
-	updateTemplateTier(t, hostAwait, "chocolate", "", "advanced")
+	// create new NSTemplateTiers (derived from `base`)
+	cheesecakeTier := CreateNSTemplateTier(t, hostAwait, "cheesecake", baseTier)
+	cookieTier := CreateNSTemplateTier(t, hostAwait, "cookie", baseTier)
+	chocolateTier := CreateNSTemplateTier(t, hostAwait, "chocolate", baseTier)
+
+	// first group of users: the "cheesecake users"
+	cheesecakeSyncIndexes := setupAccounts(t, awaitilities, cheesecakeTier, "cheesecakeuser%02d", memberAwait, count)
+	// second group of users: the "cookie users"
+	cookieSyncIndexes := setupAccounts(t, awaitilities, cookieTier, "cookieuser%02d", memberAwait, count)
```
MatousJobanek, line 167: I guess that we don't need the syncIndexes anymore

testsupport/tiers/checks.go
```diff
@@ -47,58 +46,123 @@ var (
...
+	}
+	if tierName, found := tier.Annotations[NamespaceResourcesBaseTierKey]; found {
+		checks, err := newChecks(tierName)
+		if err != nil {
+			return nil, err
+		}
+		c.getNamespaceObjectChecks = checks.GetNamespaceObjectChecks
+	}
+	return c, nil
+}
```
MatousJobanek, line 149: using the annotation for storing the information on what tier was used for the creation of the new custom tier is additional hidden complexity that is not needed at all.
I mean - why do we need that in the first place? only to remove a few parameters from `verifyResourceUpdatesForUserSignups` function? If so, then I don't think that it's really worth it. Just compare the complexity/amount of code between this additional logic and the additional parameters in `verifyResourceUpdatesForUserSignups`.
Or is there anything else that I'm missing?

If you want to somehow somewhere store the information on what was used and create the checks based on that, then you could create a new type - let's say `CustomTier`:
```go
type CustomTier struct {
	tier                       *toolchainv1alpha1.NSTemplateTier
	namespacesSourceTier       *toolchainv1alpha1.NSTemplateTier
	clusterResourcesSourceTier *toolchainv1alpha1.NSTemplateTier
}
```
that would be returned from `CreateNSTemplateTier` function
The same could be then used in `updateNSTemplateTier` so it would update it with the actual tier that was used as the source of the templates.

and the same could be then passed to `verifyResourceUpdatesForUserSignups` for creating the needed checks. The main benefit compared to the annotation is that this is fully transparent.

test/e2e/nstemplatetier_test.go
```diff
@@ -153,54 +152,50 @@ func TestUpdateNSTemplateTier(t *testing.T) {
...
+	WaitUntilBaseNSTemplateTierIsUpdated(t, hostAwait)
+	baseTier, err := hostAwait.WaitForNSTemplateTier("base")
+	require.NoError(t, err)
 
-	// when updating the "cheesecakeTier" tier with the "advanced" template refs for ClusterResources but keep the Namespaces refs
-	updateTemplateTier(t, hostAwait, "cheesecake", "", "advanced")
-	// and when updating the "cookie" tier to the "base" template refs
-	updateTemplateTier(t, hostAwait, "cookie", "base", "base")
-	// and when updating the "chocolate" tier to the "advanced" template for Namespace refs
-	updateTemplateTier(t, hostAwait, "chocolate", "", "advanced")
```
MatousJobanek, line 185: I noticed that you dropped the second batch of updates, but I haven't found anything that wouldn't be already covered in the first step. Could you please confirm that?

testsupport/space.go
```diff
@@ -167,16 +163,16 @@ func VerifyResourcesProvisionedForSpaceWithTiers(t *testing.T, awaitilities wait
...
-	nsTemplateSet, err := targetCluster.WaitForNSTmplSet(spaceName, wait.UntilNSTemplateSetHasTier(tierName), wait.UntilNSTemplateSetHasConditions(Provisioned()))
+	nsTemplateSet, err := targetCluster.WaitForNSTmplSet(spaceName, wait.UntilNSTemplateSetHasTier(tier.Name), wait.UntilNSTemplateSetHasConditions(Provisioned()))
 	require.NoError(t, err)
 
 	// verify NSTemplateSet with namespace & cluster scoped resources
-	tiers.VerifyGivenNsTemplateSet(t, targetCluster, nsTemplateSet, namespacesChecks, clusterResourcesChecks, templateRefs)
+	tiers.VerifyNSTemplateSet(t, targetCluster, nsTemplateSet, checks, templateRefs)
 
-	if aliasTierNamespaces == "appstudio" {
+	if tier.Name == "appstudio" || tier.Annotations[tiers.BaseTierKey] == "appstudio" {
```
MatousJobanek, line 175: there is another concern I have with using some additional annotations: It's not used in prod so we don't keep that in our minds that we would need look for it and check if it's set or not - that means that we could easily miss that in other checks plus it doesn't make any sense why we should check that for the first view.

MatousJobanek (commented):
I appreciate your effort to make the code simpler, but I'm afraid that using the extra annotation is not the best approach - I would rather use a new type that is fully transparent in what it contains and what it should be used for.

xcoulon:
> I appreciate your effort to make the code simpler, but I'm afraid that using the extra annotation is not the best approach - I would rather use a new type that is fully transparent in what it contains and what it should be used for.

it's really great to have such feedback 13 days after the PR was opened.

MatousJobanek:
> it's really great to have such a feedback 13 days after the PR was opened.

I understand your feeling. I know that it's much better to get the feedback in a much shorter time so you can continue working on it when you still have the mindset. 
I really wanted to make the PR review earlier, but there was literally no time. I was focused on more critical tasks and PRs. I think that you know that I'm trying my best in making the PR reviews as quickly as possible - you saw that in your other PRs. I'm just a human and I cannot work all day long.
After all, it's not only me in our team who you could ask for a review.

xcoulon:
> > it's really great to have such a feedback 13 days after the PR was opened.
> 
> I understand your feeling. I know that it's much better to get the feedback in a much shorter time so you can continue working on it when you still have the mindset. 

yes, for one thing

> I really wanted to make the PR review earlier, but there was literally no time. I was focused on more critical tasks and PRs. I think that you know that I'm trying my best in making the PR reviews as quickly as possible - you saw that in your other PRs. I'm just a human and I cannot work all day long. After all, it's not only me in our team who you could ask for a review.

yes, I know the circumstances. The whole team was tagged and no one took time to look at it.

test/e2e/nstemplatetier_test.go
xcoulon: I don't understand your remark

MatousJobanek: My point was that the syncIndexes were used for checking that there was an update of the MURs, but we don't use MURs for updating the namespaces anymore. So checking the syncIndexes here is not needed.

Anyway, if it works then you can keep it as it is - it can be dropped in one of the following cleanup PRs :+1:

xcoulon: yes, that thing was not in the scope of the refactoring I wanted to propose. If these `syncIndexes` are not needed anymore, they should have been removed in a previous PR (when their usage was effectively dropped)

testsupport/tiers/checks.go
xcoulon: > I mean - why do we need that in the first place? only to remove a few parameters from verifyResourceUpdatesForUserSignups function? 

yes, because IMO the code is not easy to understand. It causes overhead/friction if one has to figure out what those parameters are and what they are used for. The goal of the PR was to remove this complexity. And since we've recently decided to opt for annotations in [some](https://github.com/codeready-toolchain/host-operator/pull/607) [specific](https://github.com/codeready-toolchain/host-operator/pull/595) [cases](https://github.com/codeready-toolchain/host-operator/pull/596), I thought we could also use this mechanism here, so we just need to pass the `tier` as an argument and let the `newCustomChecks` func figure out what actual checks are needed.

MatousJobanek: There is one big difference between the cases your referenced and this one. The annotations are used to alter the behavior of the controller and are used in production. E2e tests then verify the expected behavior based on the given annotation.

However, in your case, you add annotations that are used purely in e2e tests. They don't alter anything in the controller behavior, they "only" mark what tiers were used when for the creation or for the update. As I mentioned in the example above - this can be easily solved by introducing a new type(s).
Using the new type would also solve the problem with the number of parameters - you would only pass one parameter - the `CustomTier` to be verified. 
I haven't tried to implement it, but I can imagine that in some cases you could have problem with re-using the same function for the cases when you need to pass `CustomTier` or `NSTemplateTier` - in such cases you can think of other concepts like taking not the `NSTemplateTier`, but directly the `TierChecks`...

What I'm trying to say - if the code is confusing, then let's try to make it simpler either via a new types or via fluent API or variadic function, but in a way that it's clear and transparent what it actually does. 
I have a feeling that with the annotation approach we just keep the code more or less the same (with less parameters), but hiding the complexity at places that are not transparent nor clear.

xcoulon: I'm indeed looking at your `CustomTier` suggestion and indeed, it causes troubles with other funcs which expect an `NSTemplateTier` parameter. 
Once again, we disagree on what is clear and what is not.

MatousJobanek: I understand that. 
So it's not only between you and me, let's ask the rest of the team what they would prefer:
* using annotation
* building either fluent API or custom types

@ranakan19 @rajivnathan @sbryzak @alexeykazakov 
But please, read the discussion and ideally also check the PR so you know the context.

MatousJobanek: btw, another option that just came to my mind would be replacing the three parameters with one that would be called for example `TierContext` which could be created via fluent API:
```go
tierCtx := ForTier("cheesecake").NamespaceTemplatesBasedOn("advanced").ClusterResourcesBasedOn("base")
```
which would be used in:
```go
cheesecakeSyncIndexes = verifyResourceUpdatesForUserSignups(t, awaitilities, memberAwait, cheesecakeSyncIndexes, tierContext, false)
```

This shouldn't require so many changes and should be pretty straightforward. 
I just spent 15 minutes trying to quickly implement it to check if there is some hidden complication and it seems to be working fine (from the implementation point of view):
https://github.com/codeready-toolchain/toolchain-e2e/compare/master...MatousJobanek:tiercontext?expand=1

alexeykazakov: I like the `TierContext` option!
I would try to minimize the refactoring as much as possible.

xcoulon: @MatousJobanek you suggested al alternative solution with `CustomTier` so please let me take care of that now. And I prefer variadics to fluent api, but that's just my taste

MatousJobanek: Based on the [previous comments](https://github.com/codeready-toolchain/toolchain-e2e/pull/486#discussion_r823779438) I got a feeling that you didn't want to spend the time with the `CustomTier` as it causes troubles with other funcs, so I proposed a much simpler solution that doesn't require so many changes and should solve the problem with the number of variables.
So I'm confused now... :man_shrugging: 

> I prefer variadics to fluent api, but that's just my taste

sure, no problem. Variadics is simpler from the implementation point of view. Fluent API is better as you know what functions/options are available and could be used. I'm fine with either way.

alexeykazakov: I agree Matous. I'm still voting for the TierContext option.

xcoulon: awesome.
how is the `TierContext` proposal better than the `CustomTier` one that Matous suggested above?
Also, `TierContext` is not a good name IMO.

MatousJobanek: I missed the last comment, so sorry for the late response.
I thought that the `TierContext` would be better mainly because it was much simpler and it still solved the same problem. Since it would be simpler, then it would mean:
* fewer code changes
* shorter and quicker PR reviews
* faster feedback loop

which means that we would all spend less time with the refactoring effort. 
But I see that you already changed it in another way, so let me try to find some time for a review, but since it's bigger, then it will take a bit more time.

xcoulon: I believe that the `CustomNSTemplateTier` type in my PR is quite close to the `TierContext` you suggested
There are other changes related to moving some funcs from `testsupport` to `testsupport/tiers` pkg, so it's only a matter of imports.

test/e2e/nstemplatetier_test.go
xcoulon: yes, that was my feeling as well. But we can keep this second batch of update/test if it's really worth it.

test/e2e/nstemplatetier_test.go
rajivnathan: I planned to remove them in an upcoming PR

test/e2e/nstemplatetier_test.go
```diff
@@ -153,54 +153,70 @@ func TestUpdateNSTemplateTier(t *testing.T) {
...
+	// we need to increase the timeouts on assertions/awaitilities to allow for all resources to be updated
+	hostAwaitTimeout := hostAwait.Timeout
+	memberAwaitTimeout := memberAwait.Timeout
+	defer func() {
+		// restore original timeouts
+		hostAwait.Timeout = hostAwaitTimeout
+		memberAwait.Timeout = memberAwaitTimeout
+	}()
+	hostAwait.Timeout = hostAwait.Timeout + time.Second*time.Duration(3*count*2)     // 3 batches of `count` accounts, with 2s of interval between each update
+	memberAwait.Timeout = memberAwait.Timeout + time.Second*time.Duration(3*count*2) // 3 batches of `count` accounts, with 2s of interval between each update
```
rajivnathan, line 166: Correct me if I'm wrong, but couldn't this be simplified to
```suggestion
hostAwait = hostAwait.WithRetryOptions(TimeoutOption(hostAwait.Timeout + time.Second*time.Duration(3*count*2)))       // 3 batches of `count` accounts, with 2s of interval between each update
	memberAwait = memberAwait.WithRetryOptions(TimeoutOption(memberAwait.Timeout + time.Second*time.Duration(3*count*2))) // 3 batches of `count` accounts, with 2s of interval between each update
```

testsupport/tiers/nstemplateset.go
```diff
@@ -65,23 +66,18 @@ func UntilNSTemplateSetHasTemplateRefs(expectedRevisions TemplateRefs) wait.NSTe
 	return wait.NSTemplateSetWaitCriterion{
 		Match: func(actual *toolchainv1alpha1.NSTemplateSet) bool {
 			actualNamespaces := actual.Spec.Namespaces
-			if len(actualNamespaces) != len(expectedRevisions.Namespaces) ||
-				expectedRevisions.ClusterResources == nil ||
+			if expectedRevisions.ClusterResources == nil ||
```
rajivnathan, line 69: Wouldn't the length still need to match? So wouldn't the check for `len(actualNamespaces) != len(expectedRevisions.Namespaces)` still be valid as a quick check?

testsupport/tiers/nstemplateset.go
```diff
@@ -65,23 +66,18 @@ func UntilNSTemplateSetHasTemplateRefs(expectedRevisions TemplateRefs) wait.NSTe
...
-				}
-				return false
+			actualNamespaceTmplRefs := make([]string, len(actualNamespaces))
+			for i, r := range actualNamespaces {
+				actualNamespaceTmplRefs[i] = r.TemplateRef
 			}
-			return true
+			sort.Strings(actualNamespaceTmplRefs)
+			sort.Strings(expectedRevisions.Namespaces)
+			return reflect.DeepEqual(actualNamespaceTmplRefs, expectedRevisions.Namespaces)
```
rajivnathan, line 80: Nice, this is more simple to follow 👍

testsupport/wait/host.go
```diff
@@ -1479,7 +1484,7 @@ func (a *HostAwaitility) printSpaceWaitCriterionDiffs(actual *toolchainv1alpha1.
 		}
 	}
 	// also include Spaces resources in the host namespace, to help troubleshooting
-	a.listAndPrint("Spaces", a.Namespace, &toolchainv1alpha1.SpaceList{})
+	// a.listAndPrint("Spaces", a.Namespace, &toolchainv1alpha1.SpaceList{})
```
rajivnathan, line 1487: Did you mean to uncomment this?

test/e2e/nstemplatetier_test.go
xcoulon: I did not remember about the `WithRetryOptions`, but I believe we should still reset the timeouts to their initial value, shouldn't we?

testsupport/tiers/nstemplateset.go
xcoulon: I believe that check on the length of the namespaces is implied in the `reflect.DeepEqual(actualNamespaceTmplRefs, expectedRevisions.Namespaces)` call a few lines below

testsupport/tiers/checks.go
rajivnathan: Sorry for being late, I am catching up on PR reviews. Are we still deciding what to do? I guess Matous' `TierContext` idea would have been more focused on just combining the checks? Perhaps there's a better name but I do see how it can be simpler because we wouldn't have both `CustomNSTemplateTier` and `NSTemplateTier` to deal with it would just be about checks? That said, Xavier has already done the work for CustomTier and it's working now so 🤷‍♂️  If the work wasn't already done, I would have voted for the `TierContext` idea but now that it's done I'm not sure it'd be worth spending more time on this stuff.

test/e2e/nstemplatetier_test.go
xcoulon: oh, ok, then it's much simpler, indeed

testsupport/tiers/checks.go
xcoulon: ok, let me see if there's a way to simply/factorize the way these `TierChecks` are initialized...

testsupport/tiers/nstemplateset.go
rajivnathan: yeah, it is but I meant it would return immediately, without having to sort the strings, etc. Just slightly more efficient to do that check up front.

xcoulon: right, but we're talking about a slice of 1 or 2 elements, so the benefit of making this check upfront is minimal IMO. Also, this will have to be expended with checks on the space roles in the future, so I believe that checking each section (cluster resource, namespaces and space roles) individually will be simpler to maintain.
But if it's really a concern, I can restore it.

testsupport/wait/host.go
xcoulon: done in 598b35a

testsupport/tiers/nstemplateset.go
rajivnathan: Good point, didn't think about space roles too. Agree it's simpler to keep it as is. 👍

testsupport/tiers/checks.go
xcoulon: @MatousJobanek @rajivnathan I refactored the checks in 23cefec. What I like about this solution is that we can more easily compare the checks between tiers, and the setup of the checks for the "CustomNSTemplateTier" is more straightforward. 
Please let me know what you think about it.

testsupport/tiers/checks.go
```diff
@@ -48,333 +47,234 @@ var (
...
+				clusterResourceQuotaDeployments(),
+				clusterResourceQuotaReplicas(),
+				clusterResourceQuotaRoutes(),
+				clusterResourceQuotaJobs(),
+				clusterResourceQuotaServices(),
+				clusterResourceQuotaBuildConfig(),
+				clusterResourceQuotaSecrets(),
+				clusterResourceQuotaConfigMap(),
+				numberOfClusterResourceQuotas(),
+				idlers(43200, ""))
```
rajivnathan, line 103: I think we can reduce even more duplication by including the common checks in the `clusterObjectsChecks` func and only provide the differentiating checks here
```suggestion
			return clusterObjectsChecks(
				clusterResourceQuotaCompute(cpuLimit, "1750m", "7Gi", "15Gi"),
				idlers(43200, ""))
```
```suggestion
			return clusterObjectsChecks(
				clusterResourceQuotaCompute(cpuLimit, "1750m", "7Gi", "15Gi"),
				clusterResourceQuotaDeployments(),
				clusterResourceQuotaReplicas(),
				clusterResourceQuotaRoutes(),
				clusterResourceQuotaJobs(),
				clusterResourceQuotaServices(),
				clusterResourceQuotaBuildConfig(),
				clusterResourceQuotaSecrets(),
				clusterResourceQuotaConfigMap(),
				numberOfClusterResourceQuotas(),
				idlers(43200, ""))
```

testsupport/tiers/checks.go
```diff
@@ -48,333 +47,234 @@ var (
...
+				clusterResourceQuotaServices(),
+				clusterResourceQuotaBuildConfig(),
+				clusterResourceQuotaSecrets(),
+				clusterResourceQuotaConfigMap(),
+				numberOfClusterResourceQuotas(),
+				idlers(0, "dev", "stage"))
+		case testtier:
+			return []clusterObjectsCheck{}
+		default:
+			return clusterObjectsChecks(
```
rajivnathan, line 159: I think it's safer for the default to be an error so that the tests must be explicit.

rajivnathan, line 159: I think it's safer for the default to be an error so that the tests must be explicit.

testsupport/tiers/checks.go
```diff
@@ -48,333 +47,234 @@ var (
...
+				rbacEditRole(),
+				crtadminPodsRoleBinding(),
+				crtadminViewRoleBinding(),
+				execPodsRole(),
+				numberOfToolchainRoles(2),
+				numberOfToolchainRoleBindings(4))
+			checks = append(checks, commonNetworkPolicyChecks...)
+			checks = append(checks, networkPolicyAllowFromCRW(), numberOfNetworkPolicies(6))
+			return checks
+		default:
```
rajivnathan, line 218: Same comment about default here.

rajivnathan, line 218: Same comment about default here.

testsupport/tiers/checks.go
```diff
@@ -48,333 +47,234 @@ var (
...
+		switch tierName {
+		case appstudio:
+			templateRefs := GetTemplateRefs(hostAwait, tierName)
+			verifyNsTypes(hostAwait.T, tierName, templateRefs, "appstudio")
+			return templateRefs
+		case base1ns:
+			templateRefs := GetTemplateRefs(hostAwait, tierName)
+			verifyNsTypes(hostAwait.T, tierName, templateRefs, "dev")
+			return templateRefs
+		default:
```
rajivnathan, line 253: Same comment about default here.

rajivnathan, line 253: Same comment about default here.

testsupport/tiers/checks.go
```diff
@@ -48,333 +47,234 @@ var (
...
+			return []tierObjectCheck{nsTemplateTier(tierName, 0)}
+		case baselarge:
+			return []tierObjectCheck{nsTemplateTier(tierName, 90)}
+		case baseextended:
+			return []tierObjectCheck{nsTemplateTier(tierName, 180)}
+		case hackathon:
+			return []tierObjectCheck{nsTemplateTier(tierName, 80)}
+		case testtier:
+			return []tierObjectCheck{}
+		default:
```
rajivnathan, line 274: Same comment about default here.

rajivnathan, line 274: Same comment about default here.

rajivnathan (commented):
Nice. I'm actually thinking that something like a `CustomTierChecks` was probably all we needed since our Custom NSTemplateTiers right now are really just a composition of existing NSTemplateTiers that we only use in the tests. But maybe we could expand on this idea of `CustomNSTemplateTier` in the future. eg. Further separation of namespace, cluster and other templates that could be used by some kind of Composable Tier in production.

Just left some minor comments on the tier checks refactoring.

testsupport/tiers/checks.go
xcoulon: good point! done in a013cb3

testsupport/tiers/checks.go
xcoulon: done in a013cb3

xcoulon: done in a013cb3

testsupport/tiers/checks.go
xcoulon: refactored here, but not returning an error, because we may want to change the deactivation timeout for custom tiers

testsupport/tiers/checks.go
xcoulon: refactored, but instead of returning an error, the actual NSTemplateTier is used to return the templateRefs

xcoulon: duplicate of the comment below?

testsupport/tiers/checks.go
xcoulon: done in a013cb3

xcoulon: duplicate of the comment below?

rajivnathan (commented):
Thanks for addressing my comments 🙌  Sorry about the duplicate comments, I think there was something weird going on with github.

xcoulon:
> Thanks for addressing my comments 🙌 Sorry about the duplicate comments, I think there was something weird going on with github.

thanks for your feedback @rajivnathan and no worries about the duplicate comments, that's also what I thought, but I just wanted to make sure I responded to all of them. But rest assured that no offence was implied when I responded with "duplicate of the comment below?" ;)

xcoulon:
closing as we are not able to reach a consensus on the value of the changes proposed in this PR.

PR #488, opened 2022-03-01
Fix flaky test

rajivnathan:
The `TestSpaceAndSpaceBindingCleanup` test becomes flaky after the changes from https://github.com/codeready-toolchain/host-operator/pull/595 are in place. These changes ensure the tests pass with and without the changes from that PR.

Summary of Changes:
1. The deletion of the MUR happens much faster with the changes from https://github.com/codeready-toolchain/host-operator/pull/595 and I suspect the UserSignup is recreating it right away so deleting the MUR directly and then verifying the SpaceBinding is deleted becomes flaky so to work around that we can instead deactivate the UserSignup to ensure the MUR stays deleted and then we can verify the SpaceBinding is deleted.
2. The `for Spaces` sub-tests fail because the previous test executes in under 30 seconds. To work around this I changed  the deletionThreshold timestamp to a duration instead and pass that into the `UntilSpaceHasCreationTimestampOlderThan` function so that we can compute the time as we wait and we should reach the point where 30 seconds has elapsed since the spaces were created.

### DIFF
--- a/test/e2e/space_cleanup_test.go
+++ b/test/e2e/space_cleanup_test.go
@@ -5,7 +5,8 @@ import (
 	"testing"
 	"time"
 
-	"github.com/codeready-toolchain/api/api/v1alpha1"
+	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
+	"github.com/codeready-toolchain/toolchain-common/pkg/states"
 	. "github.com/codeready-toolchain/toolchain-e2e/testsupport"
 	"github.com/codeready-toolchain/toolchain-e2e/testsupport/wait"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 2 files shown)


test/e2e/space_cleanup_test.go
```diff
@@ -37,11 +38,15 @@ func TestSpaceAndSpaceBindingCleanup(t *testing.T) {
 
 		t.Run("when mur is deleted", func(t *testing.T) {
 			// given
-			_, mur, spaceBinding := setupForSpaceBindingCleanupTest(t, awaitilities, memberAwait, "lara", "ibm")
+			_, userSignup, spaceBinding := setupForSpaceBindingCleanupTest(t, awaitilities, memberAwait, "lara", "ibm")
 
 			// when
-			err := hostAwait.Client.Delete(context.TODO(), mur)
```
MatousJobanek, line 43: that was a bad mistake :man_facepalming: 
no idea how I could think that it could work :smile: :man_shrugging: 
Thanks for fixing it :+1:

MatousJobanek (approved):
Nice catch! Thanks for fixing it :pray: :+1:

PR #499, opened 2022-03-09
Tests for new single namespace tier

rajivnathan:
Related PR: https://github.com/codeready-toolchain/host-operator/pull/612


testsupport/tiers/checks.go
```diff
@@ -152,6 +156,53 @@ func (a *baseTierChecks) GetClusterObjectChecks() []clusterObjectsCheck {
...
+		numberOfToolchainRoles(2),
+		numberOfToolchainRoleBindings(4))
+
+	checks = append(checks, commonNetworkPolicyChecks()...)
+
+	switch nsType {
+	case "dev":
+		checks = append(checks, networkPolicyAllowFromCRW(), numberOfNetworkPolicies(6))
+	}
+	return checks
```
alexeykazakov, line 180: I wonder if we should fail here if the type is something else but `dev`?

MatousJobanek: this is not needed - this method is called based on the values returned from `GetExpectedTemplateRefs` which is defined a few lines below. So it should not really happen that the method would be called with another type than `dev`.

Anyway, since there is only one ns type expected then this "switch case" is not really needed at all - you can drop it :-)

rajivnathan: Good point. Dropped in fdc2ae4e8880df89d8ee50f9ab96ee4e9f73a930

alexeykazakov (approved):

rajivnathan:
/retest
infra

alexeykazakov:
infra issue

/retest

PR #515, opened 2022-03-23
fix: get rid of reg-service flakiness

MatousJobanek:
this PR brings two improvements to get rid of reg-service flakiness:
* it doesn't force-restart the reg-service pods if it's part of e2e test execution
* lists all pods of the deployment the test is waiting for and checks that:
   * the number of pod matches the expected number of replicase
   * none of the pod is being deleted
   * all pods are ready

### DIFF
--- a/make/test.mk
+++ b/make/test.mk
@@ -28,7 +28,7 @@ else
 PUBLISH_OPERATOR := true
 endif
 
-SETUP_E2E_SERVICE_ACCOUNTS ?= true
+E2E_TEST_EXECUTION ?= true
 
 ifeq ($(IS_OSD),true)
 LETS_ENCRYPT_PARAM := --lets-encrypt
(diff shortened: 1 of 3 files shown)


testsupport/wait/awaitility.go
```diff
@@ -458,6 +460,19 @@ func (a *Awaitility) WaitForDeploymentToGetReady(name string, replicas int) {
...
+			return false, nil
+		}
+		for i := range pods.Items {
+			if util.IsBeingDeleted(&pods.Items[i]) {
+				return false, nil
+			}
+			if !podutils.IsPodReady(&pods.Items[i]) {
+				return false, nil
+			}
+		}
```
xcoulon, line 475:
```suggestion
		for _, pod := range pods.Items {
			if util.IsBeingDeleted(&pod) || !podutils.IsPodReady(&pod) {
				return false, nil
			}
		}
```

xcoulon (approved):
Looks good at first sight, just a minor suggestion in the logic to way for pods to be ready.

testsupport/wait/awaitility.go
MatousJobanek: yeah, why not - just the linter will fail with this change because of the implicit memory aliasing in for loop

xcoulon: ah yes, probably 🤦‍♂️ 
but at least the 2 "if" blocks can be combined, can't they? anyways, I'll leave that to you ;)

MatousJobanek: yes, I noticed the if statement change - thanks for it - I had already applied that, and then I updated it so the linter is ignored ;-)

make/test.mk
```diff
@@ -303,8 +303,10 @@ create-host-resources:
 		echo "{\"spec\":{\"members\":{\"specificPerMemberCluster\":{\"member-$${TOOLCHAIN_CLUSTER_NAME}2\":{\"webhook\":{\"deploy\":false}}}}}}" > $$PATCH_FILE; \
 		oc patch toolchainconfig config -n $(HOST_NS) --type=merge --patch "$$(cat $$PATCH_FILE)"; \
 	fi;
-	# delete registration-service pods in case they already exist
+ifneq ($(E2E_TEST_EXECUTION),true)
+	# if it's not part of e2e test execution, then delete registration-service pods in case they already exist
```
rajivnathan, line 307:
```suggestion
	# if it's not part of e2e test execution, then delete registration-service pods in case they already exist so that the ToolchainConfig will be reloaded
```

rajivnathan (approved):
Cool, this should help 👍

alexeykazakov:
/retest

infra issue

alexeykazakov (approved):

PR #521, opened 2022-03-30
cicd: verify that deletion of jobs help

MatousJobanek:
uses https://github.com/codeready-toolchain/toolchain-cicd/pull/61


MatousJobanek:
The workaround was verified - it fixes the problem

PR #535, opened 2022-04-07
test: create SocialEvent

xcoulon:
e2e tests for https://github.com/codeready-toolchain/host-operator/pull/624


test/e2e/socialevent_test.go
```diff
@@ -0,0 +1,72 @@
...
+	t.Run("create socialevent with valid tiername", func(t *testing.T) {
+		// given
+		se := testsupport.NewSocialEvent("lab-", "base")
+
+		// when
+		err := hostAwait.CreateWithCleanup(context.TODO(), se)
+
+		// then
+		require.NoError(t, err)
+		_, err = hostAwait.WaitForSocialEvent("lab", UntilSocialEventHasConditions(toolchainv1alpha1.Condition{
```
MatousJobanek, line 32: correct me if I'm wrong, but when creating the SocialEvent CR you used the `lab-` with the dash suffix, which would mean that it should use the generated name, so you cannot pass the `lab` name here, because it would be different.
BTW, couldn't we just use generated names by default and pass them via `se.Name`?

testsupport/wait/host.go
```diff
@@ -1744,3 +1744,77 @@ func UntilSpaceBindingHasSpaceRole(expected string) SpaceBindingWaitCriterion {
...
+}
+
+// WaitForSocialEvent waits until the SocialEvent is available with the provided criteria, if any
+func (a *HostAwaitility) WaitForSocialEvent(name string, criteria ...SocialEventWaitCriterion) (*toolchainv1alpha1.SocialEvent, error) {
+	var se *toolchainv1alpha1.SocialEvent
+
+	err := wait.Poll(a.RetryInterval, 2*a.Timeout, func() (done bool, err error) {
+		// retrieve the SocialEvent from the host namespace
+		seList := &toolchainv1alpha1.SocialEventList{}
+		if err = a.Client.List(context.TODO(), seList, client.InNamespace(a.Namespace)); err != nil {
```
MatousJobanek, line 1769: ah, you don't use the name here at all - you just list all of them. But this is not a robust approach - what if there are SocialEvents that were created in previous tests?

xcoulon: yeah, let me fix that, I didn't notice this bug

xcoulon: this was clearly a combination of poor copy-paste and lack of attention 😬
addressed in 75d3f6d

MatousJobanek (commented):
We could run the tests in parallel with the other ones - could you please move it to `test/e2e/parallel` and add `t.Parallel()` at the beginning of the test?

xcoulon:
> We could run the tests in parallel with the other ones - could you please move it to `test/e2e/parallel` and add `t.Parallel()` at the beginning of the test?

good point, let me do that!

xcoulon:
> > We could run the tests in parallel with the other ones - could you please move it to `test/e2e/parallel` and add `t.Parallel()` at the beginning of the test?
> 
> good point, let me do that!

done in b629588

MatousJobanek (approved):

rajivnathan (approved):
Nice and simple! 👍

PR #538, opened 2022-04-11
feat: store cleanup tassks by T

MatousJobanek:
this functionality is the prerequisite for running e2e tests in parallel. It stores the cleanup tasks in a map mapped by the `testing.T` instance, so it cleans only those resources for the particular test.


rajivnathan (approved):

alexeykazakov (approved):

testsupport/has/api/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -1,3 +1,4 @@
+//go:build !ignore_autogenerated
```
xcoulon, line 1: why do we have such a file here now?

MatousJobanek: do you mean file or line? 
the line was added by my version of go (1.17.x)
the file has been there for a long time already - it's a workaround for proxy e2e tests that use the appstudio tier which has a limited list of resources that users can modify.

PR #545, opened 2022-04-21
Test user tiers

rajivnathan:
Related PR: https://github.com/codeready-toolchain/host-operator/pull/632

Summary of changes:
- Test user tier update mechanism
- Verify all user tiers are created with correct values

### DIFF
--- a/testsupport/init.go
+++ b/testsupport/init.go
@@ -138,6 +138,11 @@ func WaitForDeployments(t *testing.T) wait.Awaitilities {
 		err = initHostAwait.WaitUntilBaseNSTemplateTierIsUpdated()
 		require.NoError(t, err)
 
+		// check that the default user tier exists and is updated to the current version, an outdated version is applied from deploy/e2e-tests/usertier-base.yaml as
+		// part of the e2e test setup make target for the purpose of verifying user tiers are updated on startup of the host operator
+		err = initHostAwait.WaitUntilBaseUserTierIsUpdated()
+		require.NoError(t, err)
+
 		t.Log("all operators are ready and in running state")
 	})
 
--- a/testsupport/init.go
+++ b/testsupport/init.go
@@ -140,8 +140,8 @@ func WaitForDeployments(t *testing.T) wait.Awaitilities {
 
 		// check that the default user tier exists and is updated to the current version, an outdated version is applied from deploy/e2e-tests/usertier-base.yaml as
 		// part of the e2e test setup make target for the purpose of verifying user tiers are updated on startup of the host operator
-		err = initHostAwait.WaitUntilBaseUserTierIsUpdated()
-		require.NoError(t, err)
+		// err = initHostAwait.WaitUntilBaseUserTierIsUpdated()
+		// require.NoError(t, err)
 
 		t.Log("all operators are ready and in running state")
 	})
--- a/testsupport/init.go
+++ b/testsupport/init.go
@@ -140,8 +140,8 @@ func WaitForDeployments(t *testing.T) wait.Awaitilities {
 
 		// check that the default user tier exists and is updated to the current version, an outdated version is applied from deploy/e2e-tests/usertier-base.yaml as
 		// part of the e2e test setup make target for the purpose of verifying user tiers are updated on startup of the host operator
-		// err = initHostAwait.WaitUntilBaseUserTierIsUpdated()
-		// require.NoError(t, err)
+		err = initHostAwait.WaitUntilBaseUserTierIsUpdated()
+		require.NoError(t, err)
 
 		t.Log("all operators are ready and in running state")
 	})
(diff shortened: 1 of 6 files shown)


testsupport/wait/host.go
```diff
@@ -704,6 +713,99 @@ func containsUserAccountStatus(uaStatuses []toolchainv1alpha1.UserAccountStatusE
...
+		buf.WriteString("diffs:\n")
+		for _, c := range criteria {
+			if !c.Match(actual) {
+				buf.WriteString(c.Diff(actual))
+				buf.WriteString("\n")
+			}
+		}
+	}
+	// also include other resources relevant in the host namespace, to help troubleshooting
+	buf.WriteString(a.sprintAllResources())
```
MatousJobanek, line 775: I don't think that we need to print all resources when the UserTier is not found. We could maybe all UserTiers which can be done in the `WaitForUserTier` function:
```
if err != nil {
	a.printUserTierWaitCriterionDiffs(tier, criteria...)
	a.listAndPrint("UserTier", a.Namespace, &toolchainv1alpha1.UserTierList{})
}
```
```suggestion
```

TBH. I think that we can drop this `sprintAllResources` completely and replace it with the `listAndPrint` approach, but that can be done as a separate effort.

testsupport/wait/host.go
```diff
@@ -704,6 +713,99 @@ func containsUserAccountStatus(uaStatuses []toolchainv1alpha1.UserAccountStatusE
...
+			return actual.Generation == expected
+		},
+		Diff: func(actual *toolchainv1alpha1.UserTier) string {
+			return fmt.Sprintf("expected generation value %d. Actual: %d", expected, actual.Generation)
+		},
+	}
+}
+
+func (a *HostAwaitility) WaitUntilBaseUserTierIsUpdated() error {
+	_, err := a.WaitForUserTier("deactivate30", UntilUserTierHasDeactivationTimeoutDays(30), UntilUserTierHasGeneration(2))
```
MatousJobanek, line 805: Is checking the `Generation` number necessary? isn't it enough to check that the deactivation value has been updated? The only thing that we care about is that the deactivation value is set to the expected one, we don't care how many times the resource was updated.
```suggestion

func (a *HostAwaitility) WaitUntilBaseUserTierIsUpdated() error {
	_, err := a.WaitForUserTier("deactivate30", UntilUserTierHasDeactivationTimeoutDays(30))
```

testsupport/init.go
```diff
@@ -138,6 +138,11 @@ func WaitForDeployments(t *testing.T) wait.Awaitilities {
 		err = initHostAwait.WaitUntilBaseNSTemplateTierIsUpdated()
 		require.NoError(t, err)
 
+		// check that the default user tier exists and is updated to the current version, an outdated version is applied from deploy/e2e-tests/usertier-base.yaml as
+		// part of the e2e test setup make target for the purpose of verifying user tiers are updated on startup of the host operator
+		err = initHostAwait.WaitUntilBaseUserTierIsUpdated()
+		require.NoError(t, err)
+
```
MatousJobanek, line 145: the e2e tests are failing because this code is executed in the "migration setup part" - which means before the UserTier CRD is created.

MatousJobanek (approved):
I like how the `TestVerifyUserTiers` test is so straightforward :-) 
Just a few comments, but nothing critical. To make it work you need to fix the migration test.

testsupport/init.go
rajivnathan: yeah, I'm aware of the failure. Do you have any suggestion on how to fix this? I was thinking of just commenting this out for the initial PR and then uncommenting it in a separate PR. WDYT?

MatousJobanek: yup, that would work :+1: 
another option would be:
1. comment it out
2. let the host-operator PR pass
3. merge host-operator PR
4. uncomment it and let it pass as well
5. merge this PR

I know that ideally we should follow the approach when we merge both PRs together, but there is no breaking change, so it's not a big deal. 
I'll leave it up to you - either way is fine.

rajivnathan: cool, I'll use your way since it'll save me the trouble of having to create a separate PR 🙂

rajivnathan: hmm, looks like the branch pairing fails if the host operator PR is merged so let me create a separate PR to merge these tests

rajivnathan: new PR: https://github.com/codeready-toolchain/toolchain-e2e/pull/548

rajivnathan:
Closing in favour of https://github.com/codeready-toolchain/toolchain-e2e/pull/548

testsupport/init.go
MatousJobanek: ah, right - I totally forgot about this. Sorry for causing that trouble :-)

testsupport/wait/host.go
MatousJobanek: I see, then maybe change the check so it's not equal to `1`. But it's not super important - I guess that we don't do that in NSTemplateTier either.

PR #546, opened 2022-04-22
feat: run registration_service_test in parallel

MatousJobanek:
This is another PR in the effort of running e2e tests in parallel. This time it is for `registration_service_test.go`
Changes:
* moved `registration_service_test.go` to `parallel` dir
* transformed the test from suites to normal standalone tests (the testing framework doesn't have full support for running suites in parallel https://github.com/stretchr/testify/issues/187)
* introduce & use `WaitAndVerifyThatUserSignupIsNotCreated` to verify that a UserSignup is not created

### DIFF
--- a/test/e2e/parallel/registration_service_test.go
+++ b/test/e2e/parallel/registration_service_test.go
@@ -287,7 +287,7 @@ func TestSignupFails(t *testing.T) {
 		require.Equal(t, float64(403), response["code"])
 
 		hostAwait := await.Host()
-		hostAwait.WaitAndVerifyThatUserSignupIsNotCreated(identity.ID.String())
+		hostAwait.WithRetryOptions(wait.TimeoutOption(time.Second*15)).WaitAndVerifyThatUserSignupIsNotCreated(identity.ID.String())
 	})
 }
 
--- a/test/e2e/parallel/registration_service_test.go
+++ b/test/e2e/parallel/registration_service_test.go
@@ -1,4 +1,4 @@
-package e2e
+package parallel
 
 import (
 	"bytes"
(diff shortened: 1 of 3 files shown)


test/e2e/parallel/registration_service_test.go
```diff
@@ -273,94 +278,100 @@ func (s *registrationServiceTestSuite) TestSignupFails() {
...
-		require.Nil(s.T(), userSignup)
-		require.Error(s.T(), err)
-		require.EqualError(s.T(), err, "timed out waiting for the condition")
+		response := invokeEndpoint(t, "POST", route+"/api/v1/signup", token, "", http.StatusForbidden)
+		require.Equal(t, "forbidden: failed to create usersignup for test-crtadmin", response["message"])
+		require.Equal(t, "error creating UserSignup resource", response["details"])
+		require.Equal(t, float64(403), response["code"])
+
+		hostAwait := await.Host()
+		hostAwait.WaitAndVerifyThatUserSignupIsNotCreated(identity.ID.String())
```
alexeykazakov, line 290: We usually use custom (shorter) timeouts for checks which verify that the resource does **not** exist to speed up the test execution. An example: https://github.com/codeready-toolchain/toolchain-e2e/blob/c288b650eb6f8df5cbb6800f59bbc0b1ebae3856/test/metrics/metrics_test.go#L241

So something like 
```suggestion
		hostAwait.WithRetryOptions(wait.TimeoutOption(time.Second*15)).WaitAndVerifyThatUserSignupIsNotCreated(identity.ID.String())
```
The default timeout is 60 seconds. So otherwise this test will hang for 60 seconds if we do not use a custom timeout.

alexeykazakov (approved):
Looks good. Just one timeout suggestion.

test/e2e/parallel/registration_service_test.go
MatousJobanek: Ah, sure - just forgot to add the shorter timeout again when solving the conflict locally.
Good catch - thanks :+1:

PR #555, opened 2022-05-09
refactor(test): simplify token generation

xcoulon:
wrap call to toolchain-common helper func


sbryzak (approved):
Very nice, this is a good improvement.

PR #556, opened 2022-05-12
Chore: Add empty metrics instance for appstudio

sawood14012:
Hello, we are using the metrics pkg as an dependency for load testing AppStudio , well we just needed one empty metrics instance that we will add the workloads later from our script PTAL .

### DIFF
--- a/setup/metrics/gather.go
+++ b/setup/metrics/gather.go
@@ -73,6 +73,7 @@ func New(t terminal.Terminal, cl client.Client, token string, interval time.Dura
 	return g
 }
 
+//nolint:golint,unused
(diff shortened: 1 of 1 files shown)


MatousJobanek (approved):
This looks fine, but keep in mind that the plan for the service-preview has been changed since the time we had a call regarding the load testing. Service-preview won't be based on the pre-KCP Sandbox solution but directly on the KCP.

sawood14012:
> This looks fine, but keep in mind that the plan for the service-preview has been changed since the time we had a call regarding the load testing. Service-preview won't be based on the pre-KCP Sandbox solution but directly on the KCP.

Thanks ! will discuss and let you know the next plans

PR #560, opened 2022-05-16
Add notification sent check in e2e test

ranakan19:
fix wrt : https://issues.redhat.com/browse/CRT-1418

Add check that Idler Triggered notification is created and sent. And Also the status of Idler is updated with Condition to reflect the same. 

paired with : https://github.com/codeready-toolchain/member-operator/pull/357 
Note: IdlerTriggeredNotion does not have the owner label as Controller Owner Reference cannot be set.


testsupport/wait/host.go
```diff
@@ -1064,6 +1064,30 @@ func (a *HostAwaitility) WaitForNotifications(username, notificationType string,
 	return notifications, err
 }
 
+// WaitForNotificationWithName waits until there is an expected Notifications available for the provided user and with the notification type and which match the conditions (if provided).
```
rajivnathan, line 1067:
```suggestion
// WaitForNotificationWithName waits until there is an expected Notifications available with the provided name and with the notification type and which match the conditions (if provided).
```

testsupport/wait/host.go
```diff
@@ -1064,6 +1064,30 @@ func (a *HostAwaitility) WaitForNotifications(username, notificationType string,
...
+func (a *HostAwaitility) WaitForNotificationWithName(notificationName, notificationType string, criteria ...NotificationWaitCriterion) ([]toolchainv1alpha1.Notification, error) {
+	a.T.Logf("waiting for notification with name '%s'", notificationName)
+	var notifications []toolchainv1alpha1.Notification
```
rajivnathan, line 1070: Since we're using Get we should only expect a single notification so no need for the slice

testsupport/wait/host.go
```diff
@@ -1064,6 +1064,30 @@ func (a *HostAwaitility) WaitForNotifications(username, notificationType string,
...
+		notification := &toolchainv1alpha1.Notification{}
+		if err := a.Client.Get(context.TODO(), types.NamespacedName{Name: notificationName, Namespace: a.Namespace}, notification); err != nil {
+			return false, err
+		}
+		if typeFound, found := notification.GetLabels()[toolchainv1alpha1.NotificationTypeLabelKey]; !found {
+			return false, fmt.Errorf("notification found with name does not have type label")
+		} else if typeFound != notificationType {
+			return false, fmt.Errorf("notification found with name does not have the expected type")
+		}
+		notifications = append(notifications, *notification)
```
rajivnathan, line 1081: If we find a notification and append it here but the call to `matchNotificationWaitCriterion` returns false then `wait.Poll` will run again and we'd end up appending the notification multiple times in `notifications`? If we get rid of the slice then no need to worry about this.

test/e2e/user_workloads_test.go
```diff
@@ -71,6 +72,11 @@ func (s *userWorkloadsTestSuite) TestIdlerAndPriorityClass() {
 		err := memberAwait.WaitUntilPodsDeleted(p.Namespace, wait.WithPodName(p.Name))
 		require.NoError(s.T(), err)
 	}
+	// check notification was created and Idler condition was updated
+	_, err = memberAwait.WaitForIdler("test-idler-dev", wait.IdlerConditions(Running(), IdledNotificationCreated()))
```
rajivnathan, line 76: Wouldn't there be a Notification created for the `test-idler-stage` namespace too?

rajivnathan (commented):
Nice and simple! Just a couple comments added. Also, it would be good to test that the Notifications haven't changed in the 2nd part of the test when it creates another pod see the comment `Create another pod and make sure it's deleted.`

test/e2e/user_workloads_test.go
ranakan19: no, the short timeOut which makes the pod long running is only set for `test-idler-dev` namespace.

testsupport/wait/host.go
ranakan19: That's a good point! Thank you for pointing that out. I was returning a slice before because the functions `matchNotificationWaitCriterion` and `printNotificationWaitCriterionDiffs` take a slice as input parameter, and I was trying to minimize change. But as you pointed out, that is a good reason to create a new function which take single notification as input. 
done in [f571b95](https://github.com/codeready-toolchain/toolchain-e2e/pull/560/commits/f571b955373036242fc903480344d27435a26df1)

testsupport/wait/host.go
ranakan19: done in [f571b95](https://github.com/codeready-toolchain/toolchain-e2e/pull/560/commits/f571b955373036242fc903480344d27435a26df1)

testsupport/wait/host.go
```diff
@@ -1064,6 +1073,30 @@ func (a *HostAwaitility) WaitForNotifications(username, notificationType string,
...
+		if err := a.Client.Get(context.TODO(), types.NamespacedName{Name: notificationName, Namespace: a.Namespace}, notification); err != nil {
+			return false, err
+		}
+		if typeFound, found := notification.GetLabels()[toolchainv1alpha1.NotificationTypeLabelKey]; !found {
+			return false, fmt.Errorf("notification found with name does not have type label")
+		} else if typeFound != notificationType {
+			return false, fmt.Errorf("notification found with name does not have the expected type")
+		}
+
+		return matchNotificationWaitCriterion(*notification, criteria...), nil
```
rajivnathan, line 1091:
```suggestion
		return matchNotificationsWaitCriterion([]toolchainv1alpha1.Notification{*notification}, criteria...), nil
```

testsupport/wait/host.go
```diff
@@ -1012,6 +1012,15 @@ func matchNotificationWaitCriterion(actual []toolchainv1alpha1.Notification, cri
...
 
+func matchNotificationWaitCriterion(notification toolchainv1alpha1.Notification, criteria ...NotificationWaitCriterion) bool {
+	for _, c := range criteria {
+		if !c.Match(notification) {
+			return false
+		}
+	}
+	return true
+}
+
```
rajivnathan, line 1023:
```suggestion
```
We don't really need an additional function since we can just put the single notification in a slice and pass it to the `matchNotificationsWaitCriterion` function.

test/e2e/user_workloads_test.go
rajivnathan: Ah ok, that makes sense 👍 But in that case we should verify no notification was created for `test-idler-stage`

ranakan19: done in [7ccb6b7](https://github.com/codeready-toolchain/toolchain-e2e/pull/560/commits/7ccb6b74640b9be103ebc902249245d76fcfc7d5)

test/e2e/user_workloads_test.go
```diff
@@ -71,22 +74,31 @@ func (s *userWorkloadsTestSuite) TestIdlerAndPriorityClass() {
...
 	// Now we want to verify that the idler reconcile is triggered without modifying the Idler resource.
+	// Notification shouldn't be created again.
 	pod := s.createStandalonePod(idler.Name, "idler-test-pod-2")    // create just one standalone pod. No need to create all possible pod controllers which may own pods.
 	_, err = memberAwait.WaitForPod(idler.Name, "idler-test-pod-2") // pod was created
 	require.NoError(s.T(), err)
 	time.Sleep(time.Duration(2*idler.Spec.TimeoutSeconds) * time.Second)
 	err = memberAwait.WaitUntilPodDeleted(pod.Namespace, pod.Name)
 	require.NoError(s.T(), err)
+	_, err = hostAwait.WaitForNotificationWithName("test-idler-dev-idled", toolchainv1alpha1.NotificationTypeIdled, wait.UntilNotificationHasConditions(Sent()))
+	require.True(s.T(), errors.IsNotFound(err))
```
rajivnathan, line 101: I'm a bit confused. Why would we expect the notification to not be found at this point when the same check finds it on line 80? I understand that it shouldn't be created again but wouldn't we still find the same notification from line 80?

test/e2e/user_workloads_test.go
```diff
@@ -71,22 +74,31 @@ func (s *userWorkloadsTestSuite) TestIdlerAndPriorityClass() {
...
 
-	// make sure that "noise" pods are still there
+	// make sure that "noise" pods are still there, and notification is not created for stage namespace
 	_, err = memberAwait.WaitForPods(idlerNoise.Name, len(podsNoise), wait.PodRunning(), wait.WithPodLabel("idler", "idler"), wait.WithSandboxPriorityClass())
 	require.NoError(s.T(), err)
 	_, err = memberAwait.WaitForPods("workloads-noise", len(externalNsPodsNoise), wait.PodRunning(), wait.WithPodLabel("idler", "idler"), wait.WithOriginalPriorityClass())
 	require.NoError(s.T(), err)
-
+	_, err = hostAwait.WaitForNotificationWithName("test-idler-stage-idled", toolchainv1alpha1.NotificationTypeIdled, wait.UntilNotificationHasConditions(Sent()))
+	require.True(s.T(), errors.IsNotFound(err))
```
rajivnathan, line 89:
```suggestion
	require.True(s.T(), errors.IsNotFound(err))
	
```
Let's add a new line here to make the code a bit more readable

test/e2e/user_workloads_test.go
```diff
@@ -71,22 +74,31 @@ func (s *userWorkloadsTestSuite) TestIdlerAndPriorityClass() {
...
+	require.NoError(s.T(), err)
 
-	// make sure that "noise" pods are still there
+	// make sure that "noise" pods are still there, and notification is not created for stage namespace
 	_, err = memberAwait.WaitForPods(idlerNoise.Name, len(podsNoise), wait.PodRunning(), wait.WithPodLabel("idler", "idler"), wait.WithSandboxPriorityClass())
 	require.NoError(s.T(), err)
 	_, err = memberAwait.WaitForPods("workloads-noise", len(externalNsPodsNoise), wait.PodRunning(), wait.WithPodLabel("idler", "idler"), wait.WithOriginalPriorityClass())
 	require.NoError(s.T(), err)
-
+	_, err = hostAwait.WaitForNotificationWithName("test-idler-stage-idled", toolchainv1alpha1.NotificationTypeIdled, wait.UntilNotificationHasConditions(Sent()))
```
rajivnathan, line 88:
```suggestion
	_, err = hostAwait.WithRetryOptions(wait.TimeoutOption(10*time.Second)).WaitForNotificationWithName("test-idler-stage-idled", toolchainv1alpha1.NotificationTypeIdled, wait.UntilNotificationHasConditions(Sent()))
```
By default the timeout is 1 minute, but we don't really need to wait that long in the tests so let's set a shorter timeout option just for this check

test/e2e/user_workloads_test.go
ranakan19: good point! Its because its deletion default time for e2e tests is set to 5sec (instead of the default 24hr for actual mailgun service) here - https://github.com/codeready-toolchain/toolchain-e2e/blob/master/deploy/host-operator/e2e-tests/toolchainconfig.yaml#L11

ranakan19: fair point, done in 63dc881

rajivnathan (approved):
Nice job! 👍

test/e2e/user_workloads_test.go
```diff
@@ -71,22 +73,36 @@ func (s *userWorkloadsTestSuite) TestIdlerAndPriorityClass() {
 		err := memberAwait.WaitUntilPodsDeleted(p.Namespace, wait.WithPodName(p.Name))
 		require.NoError(s.T(), err)
 	}
+	// check notification was created and Idler condition was updated
+	_, err = memberAwait.WaitForIdler("test-idler-dev", wait.IdlerConditions(Running(), IdledNotificationCreated()))
+	require.NoError(s.T(), err)
```
alexeykazakov, line 78: The condition is an implementation detail here. What you need is to check that the notification is created (and this is what you already do below). How it's done (via conditions or something else) is an implementation detail we are not interested in the e2e tests.

```suggestion
	// check notification was created
```

test/e2e/user_workloads_test.go
```diff
@@ -71,22 +73,36 @@ func (s *userWorkloadsTestSuite) TestIdlerAndPriorityClass() {
...
 	// In the tests above the Idler reconcile was triggered after we changed the Idler resource (to set a short timeout).
 	// Now we want to verify that the idler reconcile is triggered without modifying the Idler resource.
+	// Notification shouldn't be created again.
 	pod := s.createStandalonePod(idler.Name, "idler-test-pod-2")    // create just one standalone pod. No need to create all possible pod controllers which may own pods.
 	_, err = memberAwait.WaitForPod(idler.Name, "idler-test-pod-2") // pod was created
 	require.NoError(s.T(), err)
 	time.Sleep(time.Duration(2*idler.Spec.TimeoutSeconds) * time.Second)
 	err = memberAwait.WaitUntilPodDeleted(pod.Namespace, pod.Name)
 	require.NoError(s.T(), err)
+	_, err = hostAwait.WaitForNotificationWithName("test-idler-dev-idled", toolchainv1alpha1.NotificationTypeIdled, wait.UntilNotificationHasConditions(Sent()))
```
alexeykazakov, line 104: Use a custom timeout (`wait.TimeoutOption(10*time.Second)`) to save some time on the test execution.

ranakan19: done in [43d63bd](https://github.com/codeready-toolchain/toolchain-e2e/pull/560/commits/43d63bdc215c35b00065d694a62354fbce6aa696)

test/e2e/user_workloads_test.go
ranakan19: done in [43d63bd](https://github.com/codeready-toolchain/toolchain-e2e/pull/560/commits/43d63bdc215c35b00065d694a62354fbce6aa696)

PR #572, opened 2022-06-08
Remove UserTier migration tests

rajivnathan:
Removes the temporary UserTier migration test

Paired with https://github.com/codeready-toolchain/host-operator/pull/652


test/e2e/nstemplatetier_test.go
```diff
@@ -42,7 +42,7 @@ func TestNSTemplateTiers(t *testing.T) {
 		Resources()
 
 	// all tiers to check - keep the base as the last one, it will verify downgrade back to the default tier at the end of the test
-	tiersToCheck := []string{"advanced", "basedeactivationdisabled", "baseextended", "baseextendedidling", "baselarge", "hackathon", "test", "appstudio", "base1ns", "base"}
+	tiersToCheck := []string{"advanced", "baseextendedidling", "baselarge", "test", "appstudio", "base1ns", "base"}
```
MatousJobanek, line 45: why did you drop them - there are still used in Spaces right? we would need to:
1. promote the particular Spaces so they use `base` tier
2. remove the tiers from host-operator
3. drop them also from the e2e tests

as for step 1. - as I wrote, manual "promotion" should be enough since there are only a few Spaces that use these tiers. There is no need for another migration code.

MatousJobanek (approved):

test/e2e/nstemplatetier_test.go
rajivnathan: As discussed, will drop them in a separate PR when the tiers from host-operator are removed.

rajivnathan: Reverted the change 7aac7120e3720a650361fc22cd1b3cda7031467f

PR #606, opened 2022-09-26
Provide more operator installation information during perf setup

rajivnathan:
- include Subscription information and messaging to identify when a Subscription can be updated to make the first time setup of operators faster on subsequent tries
- also output installation duration information

### DIFF
--- a/setup/operators/operators.go
+++ b/setup/operators/operators.go
@@ -23,6 +23,7 @@ const (
 )
 
 var Templates = []string{
+	"devspaces.yaml",
 	"rhoda.yaml",
 	"rhods.yaml",
 	"aikit.yaml",
--- a/setup/operators/operators.go
+++ b/setup/operators/operators.go
@@ -28,7 +28,7 @@ var Templates = []string{
 	"rhods.yaml",
 	"aikit.yaml",
 	"openvino.yaml",
-	"devworkspace-operator.yaml",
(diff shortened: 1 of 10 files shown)


setup/operators/operators.go
```diff
@@ -91,28 +91,47 @@ func EnsureOperatorsInstalled(cl client.Client, s *runtime.Scheme, templatePaths
...
 		err = wait.ForSubscriptionWithCriteria(cl, subscriptionResource.GetName(), subscriptionResource.GetNamespace(), func(subscription *v1alpha1.Subscription) bool {
 			currentCSV = subscription.Status.CurrentCSV
 			if currentCSV == "" {
 				return false
 			}
-			// waiting for csv should fail quickly so that the currentCSV can be reloaded in case it was changed
+
+			if len(lastCSVs) == 0 || currentCSV != lastCSVs[len(lastCSVs)-1] { // subscription's current CSV has changed
+				lastCSVs = append(lastCSVs, currentCSV)
+				fmt.Printf("CurrentCSV of subscription: '%s'\n", currentCSV)
```
alexeykazakov, line 108: We could use some logger for nicer output.

alexeykazakov (approved):

setup/operators/operators.go
rajivnathan: Yeah, I should probably do that 🙂 I think I'll do it in a separate PR

PR #609, opened 2022-09-29
Removed migration test

sbryzak:
e2e tests for https://github.com/codeready-toolchain/host-operator/pull/689


xcoulon (commented):
what about the setup and execution of the migration tests in `make/test.mk`? Do we still need it now that the migration is done?

xcoulon (approved):

PR #613, opened 2022-10-03
Use working set memory metric for perf testing

rajivnathan:
- Use `container_memory_working_set_bytes` because `container_memory_usage_bytes` includes the usage of all kinds of memory (including cache that can be evicted at any time) while container_memory_working_set_bytes  only includes the memory that cannot be evicted. So it's a better representation of the actual memory usage.
- Continue gathering metrics for 15m after provisioning users because memory continues to grow for some time after provisioning is completed


alexeykazakov (approved):

ranakan19 (approved):
LGTM

PR #615, opened 2022-10-06
Update imagestream in default user-workloads

khrm:
Imagestream doesn't exists which leads to generation lot of events w.r.t. buildconfig.


rajivnathan (approved):
Thanks for submitting this change @khrm

PR #622, opened 2022-11-11
Added tests for notifications

sbryzak:
Added a few tests to ensure that all notification types created correctly populate the user email property.

Fixes https://issues.redhat.com/projects/CRT/issues/CRT-1221?filter=myopenissues

### DIFF
--- a/test/e2e/user_management_test.go
+++ b/test/e2e/user_management_test.go
@@ -164,6 +164,37 @@ func (s *userManagementTestSuite) TestUserDeactivation() {
 		require.NoError(t, err)
 	})
 
+	s.T().Run("verify notification created and has correct usersignup email on user deactivation", func(t *testing.T) {
+
+		// User on member cluster 1
+		userToDeactivate, _ := NewSignupRequest(t, s.Awaitilities).
+			Username("usertodeactivate3").
+			Email("usertodeactivate3@redhat.com").
+			ManuallyApprove().
+			EnsureMUR().
+			TargetCluster(memberAwait).
+			RequireConditions(ConditionSet(Default(), ApprovedByAdmin())...).
+			Execute().Resources()
+
+		// Set the user to deactivated
+		userSignup, err := hostAwait.UpdateUserSignup(userToDeactivate.Name, func(us *toolchainv1alpha1.UserSignup) {
+			states.SetDeactivated(us, true)
+		})
+		require.NoError(s.T(), err)
+		s.T().Logf("user signup '%s' set to deactivated", userSignup.Name)
+
+		_, err = hostAwait.WaitForUserSignup(userSignup.Name,
+			wait.UntilUserSignupHasConditions(ConditionSet(Default(), ApprovedByAdmin(), DeactivatedWithoutPreDeactivation())...))
+		require.NoError(t, err)
+
+		notifications, err := hostAwait.WaitForNotifications(userSignup.Name, toolchainv1alpha1.NotificationTypeDeactivated, 1)
+		require.NoError(t, err)
+
+		require.NotEmpty(t, notifications)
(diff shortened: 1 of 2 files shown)


test/e2e/usersignup_test.go
```diff
@@ -136,6 +136,33 @@ func (s *userSignupIntegrationTest) TestAutomaticApproval() {
...
+	// when
+	userSignup, _ := NewSignupRequest(s.T(), s.Awaitilities).
+		Username("bartsimpson").
+		Email("bartsimpson@redhat.com").
+		RequireConditions(ConditionSet(Default(), ApprovedAutomatically())...).
+		Execute().Resources()
+
+	// then
+	userSignup, err := s.Host().WaitForUserSignup(userSignup.Name,
+		wait.UntilUserSignupHasConditions(ConditionSet(Default(), ApprovedAutomatically())...),
```
MatousJobanek, line 152: isn't the presence of the conditions already ensured by `RequireConditions(ConditionSet(Default(), ApprovedAutomatically())...).` ?

test/e2e/user_management_test.go
```diff
@@ -314,10 +345,19 @@ func (s *userManagementTestSuite) TestUserDeactivation() {
 			murMember1.Status.ProvisionedTime.String())
 
 		// The user should be set to deactivating, but not deactivated
-		_, err = hostAwait.WaitForUserSignup(userSignupMember1.Name, wait.UntilUserSignupHasConditions(
+		userSignupMember1, err = hostAwait.WaitForUserSignup(userSignupMember1.Name, wait.UntilUserSignupHasConditions(
 			ConditionSet(Default(), ApprovedByAdmin(), Deactivating())...))
 		require.NoError(s.T(), err)
 
+		// Verify that the deactivating notification was created
+		notifications, err := hostAwait.WaitForNotifications(userSignupMember1.Name, toolchainv1alpha1.NotificationTypeDeactivating, 1)
```
MatousJobanek, line 353: IIRC, we create notifications using the "compliant username" and not usersignup name.

test/e2e/usersignup_test.go
```diff
@@ -136,6 +136,33 @@ func (s *userSignupIntegrationTest) TestAutomaticApproval() {
...
+		wait.UntilUserSignupHasStateLabel(toolchainv1alpha1.UserSignupStateLabelValueApproved))
+	require.NoError(s.T(), err)
+
+	notifications, err := s.Host().WaitForNotifications(userSignup.Name, toolchainv1alpha1.NotificationTypeProvisioned, 1)
```
MatousJobanek, line 156: the same - you should use compliant username instead

test/e2e/usersignup_test.go
```diff
@@ -136,6 +136,33 @@ func (s *userSignupIntegrationTest) TestAutomaticApproval() {
...
+	require.NoError(s.T(), err)
+
+	VerifyResourcesProvisionedForSignup(s.T(), s.Awaitilities, userSignup, "deactivate30", "base")
```
MatousJobanek, line 159: we don't need to verify the created resources - let's focus only on the notification.

MatousJobanek (commented):
I think that for the purpose of the Jira issue:
> Update e2e tests to verify that recipient is always set in Notification CRs

it should be sufficient (and maybe even better) to move the assertion
```
require.NotEmpty(t, notifications.Spec.Recipient)
require.Equal(t,<recipient>, notifications.Spec.Recipient)
```
into 
```
hostAwait.WaitForNotifications
``` 
function so it makes sure that the recipient field is always correctly set,  no matter what kind of notification it is.

alexeykazakov:
This PR is almost one year old. Please reopen it if/when you still want to work on it.

PR #627, opened 2022-11-29
(fix): test metrics when users banned not cleaning up usersignup before next test

mfrancisc:
This PR fixes an issue with test metrics assertion failing because of usersignup and metrics where not in synch.

The `unban the banned user` scenario was exiting with a counter.SpaceCount exceeding by 1 the real number of spaces on the cluster causing the next test (`TestMetricsWhenUserDisabled`) to fail.

Following was the error (observed rarely also in CI):

```
awaitility.go:348: waiting for metric 'sandbox_spaces_current{[cluster_name member-fmuntean-devsandbox-cluster.devcluster.openshift.com]}' to reach '2'
    awaitility.go:356:
        	Error Trace:	/Users/fmuntean/go/src/github.com/codeready-toolchain/toolchain-e2e/test/metrics/awaitility.go:356
        	            				/Users/fmuntean/go/src/github.com/codeready-toolchain/toolchain-e2e/test/metrics/metrics.go:94
        	            				/Users/fmuntean/go/src/github.com/codeready-toolchain/toolchain-e2e/test/metrics/metrics_test.go:327
        	Test:       	TestMetricsWhenUserDisabled
        	Messages:   	waited for metric 'sandbox_spaces_current{[cluster_name member-fmuntean-devsandbox-cluster.devcluster.openshift.com]}' to reach '2'. Current value: 0
    clean.go:112: deleting UserSignup: janedoe ...
    clean.go:127: waiting until UserSignup: janedoe is completely deleted
    clean.go:152: the related MasterUserRecord: janedoe is deleted as well
--- FAIL: TestMetricsWhenUserDisabled (131.59s)

```


test/metrics/metrics_test.go
```diff
@@ -291,6 +294,20 @@ func TestMetricsWhenUsersBanned(t *testing.T) {
...
+		// wait for original usersignup to be deleted
+		err = hostAwait.WaitUntilUserSignupDeleted(userSignup.GetName())
+		require.NoError(t, err)
+		err = hostAwait.WaitUntilSpaceAndSpaceBindingsDeleted(userSignup.GetName())
+		require.NoError(t, err)
+		// and verify metrics reset
+		metricsAssertion.WaitForMetricDelta(UserAccountsMetric, 0, "cluster_name", memberAwait.ClusterName)
+		metricsAssertion.WaitForMetricDelta(UserAccountsMetric, 0, "cluster_name", memberAwait2.ClusterName)
+		metricsAssertion.WaitForMetricDelta(SpacesMetric, 0, "cluster_name", memberAwait.ClusterName)
+		metricsAssertion.WaitForMetricDelta(SpacesMetric, 0, "cluster_name", memberAwait2.ClusterName)
```
MatousJobanek, line 310: this shouldn't be needed as all users provisioned are cleaned at the end of the test https://github.com/codeready-toolchain/toolchain-e2e/blob/master/testsupport/cleanup/clean.go
But what you could do is to add a wait for cleanup of Spaces in that file as well - something similar as we do for MURs https://github.com/codeready-toolchain/toolchain-e2e/blob/master/testsupport/cleanup/clean.go#L144

MatousJobanek (commented):
good catch :+1:  I have one comment/proposal regarding cleanup at the end of the test.

test/metrics/metrics_test.go
mfrancisc: I saw that the provisioned usersignup are cleaned between one test and the other, but still I've noticed that there is a synchronization issue with the metric for space counts being updated by the time next test starts.

ok let me try and create a similar function for spaces and see if it still passes. thanks for your suggestion!

testsupport/cleanup/clean.go
```diff
@@ -174,3 +183,34 @@ func (c *cleanTask) verifyMurDeleted(isUserSignup bool, userSignup *toolchainv1a
...
 }
+
+func (c *cleanTask) verifySpaceDeleted(isUserSignup bool, userSignup *toolchainv1alpha1.UserSignup, delete bool) (bool, error) {
+	// only applicable for UserSignups
+	if !isUserSignup {
+		return true, nil
+	}
+
+	space := &toolchainv1alpha1.Space{}
+	if err := c.client.Get(context.TODO(), test.NamespacedName(userSignup.GetNamespace(), userSignup.GetName()), space); err != nil {
```
MatousJobanek, line 194: We don't use UserSignup names for Space names. 
But instead of looking for the Spaces by name, I would rather use the "creator" label. 
See: https://github.com/codeready-toolchain/host-operator/blob/master/controllers/usersignup/space.go#L11

mfrancisc: ok I see, thanks! it's better then using the `userSignup.Status.CompliantUsername` used here https://github.com/codeready-toolchain/host-operator/blob/7186f30efd0ad8273ed7eff02a996df35a4a5ec2/controllers/usersignup/space.go#L17 ?

mfrancisc: If I'm seeing it correctly seems like spaces are created in different ways during e2e tests. 
For example running this test : https://github.com/codeready-toolchain/toolchain-e2e/blob/855cd504a25d35a4a9382cfc95b9f4ab69516270/test/e2e/parallel/space_test.go#L16

The resources are creating without that expected label:

```
➜  toolchain-e2e git:(fix/test-stabilization) ✗ k get space,usersignups -o yaml
apiVersion: v1
items:
  kind: Space
  metadata:
    creationTimestamp: "2022-12-02T15:51:56Z"
    finalizers:
    - finalizer.toolchain.dev.openshift.com
    generateName: testcreatespace-
    generation: 1
    labels:
      toolchain.dev.openshift.com/appstudio-tier-hash: 28fcd1b930571badaaf23b8e10c0585e
      toolchain.dev.openshift.com/state: cluster-assigned
    name: testcreatespace-xgx9n
    namespace: toolchain-host-02123953
    ........
    
  kind: UserSignup
  metadata:
    annotations:
      toolchain.dev.openshift.com/activation-counter: "1"
      toolchain.dev.openshift.com/last-target-cluster: member-fmuntean-sandbox-dev1.devcluster.openshift.com2
      toolchain.dev.openshift.com/skip-auto-create-space: "true"
      toolchain.dev.openshift.com/user-email: for-space-testcreatespace-xgx9n@acme.com
      toolchain.dev.openshift.com/verification-counter: "0"
    creationTimestamp: "2022-12-02T15:51:57Z"
    generation: 2
    labels:
      toolchain.dev.openshift.com/email-hash: 671a2426f9aaf785bdfefc545c0f35d2
      toolchain.dev.openshift.com/state: approved
    name: for-space-testcreatespace-xgx9n
```
but maybe it's because the signup is created using this function https://github.com/codeready-toolchain/toolchain-e2e/blob/855cd504a25d35a4a9382cfc95b9f4ab69516270/testsupport/space.go#L192

that issues the signup request with `.NoSpace()` .

So in brief , not sure it will always be able to find spaces associated with usersignups, but maybe I'm missing something.

mfrancisc:
With the code that looks for spaces with creator label I see that following parallel tests are failing:

1. TestSpaceAndSpaceBindingCleanup/for_SpaceBinding/when_space_is_deleted
2. TestCreateSpace/failed_to_create_space_-_unknown_target_member_cluster 

- For the second failure my suspicious is that the listing spaces in the new `verifySpacesDeleted` function triggers a reconcile  of the space controller which set's a different state and causes the test to fail: 
https://github.com/codeready-toolchain/host-operator/blob/838e40c372d3f960a15a66f1da2d824160a4c836/controllers/space/space_controller.go#L438
see diff failure:
```
  -expected
        +actual
          []v1alpha1.Condition{
          	{
          		Type:               "Ready",
          		Status:             "False",
        - 		LastTransitionTime: v1.Time{},
        + 		LastTransitionTime: v1.Time{Time: s"2022-12-04 21:05:47 +0000 UTC"},
        - 		Reason:             "UnableToProvision",
        + 		Reason:             "UnableToTerminate",
        - 		Message:            "unknown target member cluster 'unknown'",
        + 		Message:            "cannot delete NSTemplateSet: unknown target member cluster: 'unknown'",
          		LastUpdatedTime:    nil,
          	},
```

- For the first test I'm not understanding  why NSTemplateSet is missing.

I've tried running the test suites one at the time locally and they pass, so it's seems to be something that happens only when there is parallelism involved ( plus the new function that I'm trying to add, which does something wrong at this point).

@MatousJobanek could you please help me figure out why is this happening ?

MatousJobanek:
interesting - both cases caused the same problem.
We have a cleanup controller that deletes a Space if:
* there is no SpaceBinding for the given Space
* the Space is at least 30s old: https://github.com/codeready-toolchain/host-operator/blob/master/controllers/spacecleanup/space_cleanup_controller.go#L41

I analyzed the operator logs and in both cases it took [this function](https://github.com/codeready-toolchain/toolchain-e2e/blob/master/testsupport/space.go#L192-L214) more than 30 seconds to:
1. create UserSignup
2. wait until MUR is present
3. create SpaceBinding

it's definitely given by the nature of the parallelized tests - the other test executions bere blocked by another one and it got to the creation of SpaceBinding too late. 

Anyway, the root cause is the short timeout in the Space cleanup controller. We wanted to have it parametrized at the tier level - we haven't just got to implementing it yet. I'll create an issue for this.
In the meantime, you can just easily increase the timeout for example to 60 or 90 seconds (to have some time cushion). You can also pair it with this PR ;-)

testsupport/cleanup/clean.go
```diff
@@ -174,3 +183,51 @@ func (c *cleanTask) verifyMurDeleted(isUserSignup bool, userSignup *toolchainv1a
...
+				break
+			}
+		}
+		// return error found during space deletion loop
+		if foundError {
+			return false, err
+		}
+	}
+
+	c.t.Logf("waiting until spaces with label %s:%s are completely deleted", toolchainv1alpha1.SpaceCreatorLabelKey, userSignup.GetName())
```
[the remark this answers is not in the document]
mfrancisc: Added printing of the list in this [commit](https://github.com/codeready-toolchain/toolchain-e2e/pull/627/commits/c3258244b323c113ea062e99000049113b4904d3)

MatousJobanek (approved):

PR #635, opened 2022-12-11
Store host and member operator logs in artifacts directory on OpenShift CI

mfrancisc:
Dump operator logs to files during openshift ci.

The logs will be saved in the in the `artifacts` folder with the following naming convention `${DEPLOYMENT_NAME}.log`.

Example job with logs saved to files:
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/codeready-toolchain_toolchain-e2e/635/pull-ci-codeready-toolchain-toolchain-e2e-master-e2e/1602062639536541696/artifacts/e2e/test/artifacts/

Jira ticket: [CRT-1469](https://issues.redhat.com/browse/CRT-1469)

### DIFF
--- a/make/test.mk
+++ b/make/test.mk
@@ -126,7 +126,7 @@ e2e-run:
 .PHONY: execute-tests
 execute-tests:
 	@echo "Starting test $(shell date)"
-	MEMBER_NS=${MEMBER_NS} MEMBER_NS_2=${MEMBER_NS_2} HOST_NS=${HOST_NS} REGISTRATION_SERVICE_NS=${REGISTRATION_SERVICE_NS} go test ${TESTS_TO_EXECUTE} -p 1 -parallel ${E2E_PARALLELISM} -v -timeout=90m -failfast && \
+	MEMBER_NS=${MEMBER_NS} MEMBER_NS_2=${MEMBER_NS_2} HOST_NS=${HOST_NS} REGISTRATION_SERVICE_NS=${REGISTRATION_SERVICE_NS} go test ${TESTS_TO_EXECUTE} -p 1 -parallel ${E2E_PARALLELISM} -v -timeout=90m -failfast || \
 	($(MAKE) print-logs HOST_NS=${HOST_NS} MEMBER_NS=${MEMBER_NS} MEMBER_NS_2=${MEMBER_NS_2} REGISTRATION_SERVICE_NS=${REGISTRATION_SERVICE_NS} && exit 1)
 
 .PHONY: print-logs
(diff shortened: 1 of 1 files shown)


sbryzak (approved):

make/test.mk
```diff
@@ -201,6 +201,7 @@ print-operator-logs:
 	@echo "======================= ${DEPLOYMENT_NAME} deployment logs - Namespace: ${NAMESPACE} =========================="
 	@echo "==============================================================================================================="
 	-oc logs deployment.apps/${DEPLOYMENT_NAME} ${ADDITIONAL_PARAMS} --namespace ${NAMESPACE}
+	-oc logs deployment.apps/${DEPLOYMENT_NAME} ${ADDITIONAL_PARAMS} --namespace ${NAMESPACE} > ${ARTIFACT_DIR}/${DEPLOYMENT_NAME}.log
```
xcoulon, line 204: AFAICS, the Jira ticket suggested that the logs should be kept in the artifacts directory _instead_, but here you're making a copy. My point is, the logs will still be very heavy, won't they?

mfrancisc: The logs are now stored both in :

1. https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/codeready-toolchain_toolchain-e2e/635/pull-ci-codeready-toolchain-toolchain-e2e-master-e2e/1602062639536541696/artifacts/e2e/test/build-log.txt (as single file with all the logs from the operators)
2. https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/codeready-toolchain_toolchain-e2e/635/pull-ci-codeready-toolchain-toolchain-e2e-master-e2e/1602062639536541696/artifacts/e2e/test/artifacts/ (each operator has it's own log file)

If I well understood your concern, you are suggesting to drop the operator logs from `build-log.txt` or the new files are not looking fine ? 

I can remove line 203 which stores the operator logs in `build-log.txt`, but I was unsure if someone still prefers to have all the logs in one file.

xcoulon: I'm fine with having the operator logs in a separate file instead and making `build-log.txt` less heavy, as the Jira ticket says:

> It would be more convenient if those logs were stored in individual files in the "artifacts" directory of OpenShift CI, instead.

I guess we could have a poll on Slack to decide?

rajivnathan: I vote for having the operator logs in separate files only and not as part of `build-log.txt`. We could see how it goes with that and if we find it's better to have all the logs in `build-log.txt` then we can put it back.

MatousJobanek (approved):

PR #638, opened 2022-12-14
refactor(test): introduce new HTTPRequest abstraction

mfrancisc:
This PR's introduces:

1. removal of the global variable `httpClient` 
2. move `WaitForUserSignupReadyInRegistrationService` to `host.go` 
3. create new `HTTPRequest` abstraction similar to `SignupRequest`

### DIFF
--- a/testsupport/wait/http_request.go
+++ b/testsupport/wait/http_request.go
@@ -12,6 +12,10 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// ContentTypeApplicationJSON is the default content type set during
+// HTTP request, but can be configured using the `c.ContentType` method.
+const ContentTypeApplicationJSON = "application/json"
+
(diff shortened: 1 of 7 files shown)


testsupport/wait/http_request.go
```diff
@@ -0,0 +1,211 @@
+package wait
```
xcoulon, line 1: I believe it would be better located in the `testsupport` package. The `wait` package is for func with wait/poll logic, so reading `wait.NewHTTPRequest()` is a bit confusing :(

mfrancisc: I think I will need to create a new package, since `testsupport` uses `wait` and `NewHTTPRequest` is being used in host.go by `WaitForUserSignupReadyInRegistrationService`, this will create a circular dependency error.

Any suggestion for the new package name or an existing one (different then `testsupport`) ?

PR #642, opened 2022-12-16
Update space role checks

rajivnathan:
Paired with https://github.com/codeready-toolchain/host-operator/pull/719


rajivnathan:
Skipping the TestDoNotOverrideServiceAccount test instead of deleting it because we will need to create SAs as part of the environment sub-workspaces so the test & logic will be useful to keep.

alexeykazakov (approved):

PR #646, opened 2023-01-10
small fix on creating the toolchainCluster with expected type label

mfrancisc:
create `toolchainCluster` resource in `await.Namespace` with `otherAwait.Type` .

In the scenario where `await.Namespace` is `toolchain-host-ns`, the type label for the `toolchainCluster` resource should be `otherAwait.Type` (so `member` in this case)

### DIFF
--- a/test/e2e/toolchaincluster_test.go
+++ b/test/e2e/toolchaincluster_test.go
@@ -34,7 +34,7 @@ func verifyToolchainCluster(t *testing.T, await *wait.Awaitility, otherAwait *wa
 
 	t.Run("create new ToolchainCluster with correct data and expect to be ready for cluster type "+string(await.Type), func(t *testing.T) {
 		// given
-		name := "new-ready-" + string(await.Type)
+		name := "new-ready-" + string(otherAwait.Type)
 		toolchainCluster := newToolchainCluster(await.Namespace, name,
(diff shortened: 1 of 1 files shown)


xcoulon (approved):

mfrancisc:
/retest

infra issue

test/e2e/toolchaincluster_test.go
```diff
@@ -36,7 +36,7 @@ func verifyToolchainCluster(t *testing.T, await *wait.Awaitility, otherAwait *wa
 		// given
 		name := "new-ready-" + string(await.Type)
```
MatousJobanek, line 37: this one as well right? 
```suggestion
		name := "new-ready-" + string(otherAwait.Type)
```

MatousJobanek (commented):
Thanks for fixing it.
could you please fix it also in the following test? 
https://github.com/codeready-toolchain/toolchain-e2e/blob/5549f388e2a0bd72ab0c33fdbe1dad2a4f8b1504/test/e2e/toolchaincluster_test.go#L67-L71

test/e2e/toolchaincluster_test.go
mfrancisc: ah right! missed that! thx!

PR #649, opened 2023-01-16
ASC-247 avoid clean up of subspaces

mfrancisc:
Implement e2e space test that ensures sub-spaces are not being cleaned up by the space clean up controller, when there is no spacebinding with the sub-space label.

It's covering following acceptance criteria from Jira https://issues.redhat.com/browse/ASC-247:

The cleanup space controller is updated so it cleans the Spaces up only if:
- there is no SpaceBinding for the given Space present
- at the same time, it either doesn't have parentSpace set or the parentSpace is gone or being deleted

Paired with:
- host-operator: https://github.com/codeready-toolchain/host-operator/pull/728

### DIFF
--- a/test/e2e/parallel/space_cleanup_test.go
+++ b/test/e2e/parallel/space_cleanup_test.go
@@ -70,6 +70,33 @@ func TestSpaceAndSpaceBindingCleanup(t *testing.T) {
 		space2, err := hostAwait.WaitForSpace(t, space2.Name, wait.UntilSpaceHasCreationTimestampOlderThan(deletionThreshold))
 		require.NoError(t, err)
 
+		t.Run("when space has parent-space, then it should not be deleted even if without spacebinding", func(t *testing.T) {
+			// when
+			parentSpace, err := hostAwait.WaitForSpace(t, space1.Name)
+			subSpace := CreateSubSpace(t, awaitilities, parentSpace.Name)
+			require.NoError(t, err)
+
+			// then
+			// parent-space should be provisioned
+			parentSpace, _ = VerifyResourcesProvisionedForSpace(t, awaitilities, parentSpace.Name)
+			// sub-space should be provisioned
+			VerifyResourcesProvisionedForSpace(t, awaitilities, subSpace.Name,
+				wait.UntilSpaceHasLabelWithValue(toolchainv1alpha1.ParentSpaceLabelKey, parentSpace.Name)) // check that parent-space label was added accordingly
+
+			t.Run("when SpaceBinding is inherited from parent-space, then sub-space should not be deleted", func(t *testing.T) {
+				// when
+				// there is no spaceBinding with the sub-space name as label
+				err := hostAwait.WaitUntilSpaceBindingsWithLabelDeleted(t, toolchainv1alpha1.SpaceBindingSpaceLabelKey, subSpace.Name)
+				require.NoError(t, err)
+
+				// then
+				// sub-space is provisioned
+				VerifyResourcesProvisionedForSpace(t, awaitilities, subSpace.Name,
+					wait.UntilSpaceHasLabelWithValue(toolchainv1alpha1.ParentSpaceLabelKey, parentSpace.Name)) // check that parent-space label is present
+			})
+
+		})
+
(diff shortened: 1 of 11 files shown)


test/e2e/parallel/space_cleanup_test.go
```diff
@@ -70,6 +70,33 @@ func TestSpaceAndSpaceBindingCleanup(t *testing.T) {
 		space2, err := hostAwait.WaitForSpace(t, space2.Name, wait.UntilSpaceHasCreationTimestampOlderThan(deletionThreshold))
 		require.NoError(t, err)
 
+		t.Run("when space has parent-space, then it should not be deleted even if without spacebinding", func(t *testing.T) {
+			// when
+			parentSpace, err := hostAwait.WaitForSpace(t, space1.Name)
+			subSpace := CreateSubSpace(t, awaitilities, parentSpace.Name)
+			require.NoError(t, err)
```
xcoulon, line 77:
```suggestion
			require.NoError(t, err)
			subSpace := CreateSubSpace(t, awaitilities, parentSpace.Name)
```

xcoulon: checking `err` before creating the subspace

test/e2e/parallel/space_cleanup_test.go
```diff
@@ -70,6 +70,33 @@ func TestSpaceAndSpaceBindingCleanup(t *testing.T) {
...
+
+			t.Run("when SpaceBinding is inherited from parent-space, then sub-space should not be deleted", func(t *testing.T) {
+				// when
+				// there is no spaceBinding with the sub-space name as label
+				err := hostAwait.WaitUntilSpaceBindingsWithLabelDeleted(t, toolchainv1alpha1.SpaceBindingSpaceLabelKey, subSpace.Name)
+				require.NoError(t, err)
+
+				// then
+				// sub-space is provisioned
+				VerifyResourcesProvisionedForSpace(t, awaitilities, subSpace.Name,
```
xcoulon, line 94: should we also check that it's the same subspace? for example by checking that the resource UID is still the same

xcoulon: Sorry but I'm a bit lost: why are you referring to `NewSpace`?

mfrancisc: I was referring to the fact that `subSpace` is created using `CreateSubSpace` function https://github.com/mfrancisc/toolchain-e2e/blob/cea00168e100f77ec2c95e6ca76fab70250a1d21/test/e2e/parallel/space_cleanup_test.go#L77

which uses `NewSpace` under the hood https://github.com/mfrancisc/toolchain-e2e/blob/cea00168e100f77ec2c95e6ca76fab70250a1d21/testsupport/space.go#L22

If I'm seeing it correctly, the latter function should generate a new name for each new Space, it has a fixed prefix using the name of the test (truncated) and a "random" suffix appended by kubernetes ?

mfrancisc: Ok , I think I understood your concern:
The _space cleanup controller_ deletes the `subSpace` and the _space controller_ recreates it with the same name in the meantime ? 

I've pushed a small assertion to cover this scenario 👍 
See: [e4525d7](https://github.com/codeready-toolchain/toolchain-e2e/pull/649/commits/e4525d7b7c4658fea1f12fa985f28d34cae1a01e)

xcoulon: yes, or to rephrase: the space cleanup controller should _not_ delete the subspace, hence, we should have the same resource (by UID) at the end of the day.

xcoulon (approved):
thanks for addressing my comments!

test/e2e/parallel/space_cleanup_test.go
mfrancisc: ok , good point! thanks!

mfrancisc:
/retest

infra issue

mfrancisc:
/retest

infra issue

mfrancisc:
/retest

infra issue

MatousJobanek (approved):

PR #657, opened 2023-01-24
Increase PVC size and count for AppStudio

Roming22:
Pipeline Service keeps the latest 10 PipelineRuns, so the quota is increased to 12 to allow for 2 PipelineRuns to be created before the pruner makes more space for additional PipelineRuns. Storage size is increased to 50Gi.

See https://github.com/codeready-toolchain/host-operator/pull/737

Signed-off-by: Romain Arnaud <rarnaud@redhat.com>

### DIFF
--- a/testsupport/tiers/checks.go
+++ b/testsupport/tiers/checks.go
@@ -216,7 +216,7 @@ func (a *base1nsTierChecks) GetNamespaceObjectChecks(_ string) []namespaceObject
 	checks := []namespaceObjectsCheck{
 		resourceQuotaComputeDeploy("20", "14Gi", "1750m", "14Gi"),
 		resourceQuotaComputeBuild("20", "10Gi", "2", "10Gi"),
-		resourceQuotaStorage("50Gi", "50Gi", "50Gi", "12"),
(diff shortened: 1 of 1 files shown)


testsupport/tiers/checks.go
```diff
@@ -216,7 +216,7 @@ func (a *base1nsTierChecks) GetNamespaceObjectChecks(_ string) []namespaceObject
 	checks := []namespaceObjectsCheck{
 		resourceQuotaComputeDeploy("20", "14Gi", "1750m", "14Gi"),
 		resourceQuotaComputeBuild("20", "10Gi", "2", "10Gi"),
-		resourceQuotaStorage("15Gi", "15Gi", "15Gi", "5"),
+		resourceQuotaStorage("50Gi", "50Gi", "50Gi", "12"),
```
alexeykazakov, line 219: We are not changing (and we don't want to) the quota for other tiers. Only for the appstudio one.
```suggestion
		resourceQuotaStorage("15Gi", "15Gi", "15Gi", "5"),
```

alexeykazakov (approved):

PR #658, opened 2023-01-26
Increase pod quota for AppStudio to 150

Roming22:
Signed-off-by: Romain Arnaud <rarnaud@redhat.com>

C.f. https://github.com/codeready-toolchain/host-operator/pull/739

### DIFF
--- a/testsupport/tiers/checks.go
+++ b/testsupport/tiers/checks.go
@@ -196,7 +196,7 @@ func (a *baseTierChecks) GetExpectedTemplateRefs(t *testing.T, hostAwait *wait.H
 func (a *baseTierChecks) GetClusterObjectChecks() []clusterObjectsCheck {
 	return clusterObjectsChecks(
 		clusterResourceQuotaCompute(cpuLimit, "1750m", "7Gi", "15Gi"),
-		clusterResourceQuotaDeployments(),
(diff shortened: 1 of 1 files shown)


testsupport/tiers/checks.go
```diff
@@ -883,7 +883,7 @@ func clusterResourceQuotaDeployments() clusterObjectsCheckCreator {
 			require.NoError(t, err)
 			hard[count("deploymentconfigs.apps")], err = resource.ParseQuantity("30")
 			require.NoError(t, err)
-			hard[count(corev1.ResourcePods)], err = resource.ParseQuantity("50")
+			hard[count(corev1.ResourcePods)], err = resource.ParseQuantity("150")
```
alexeykazakov, line 886: This won't work... `clusterResourceQuotaDeployments()` is used for testing all the tiers. We have it because until now all tiers used the same pod quota. But now we are changing the appstudio tier quota only. So you will have to modify line 476:
https://github.com/codeready-toolchain/toolchain-e2e/blob/master/testsupport/tiers/checks.go#L476
And probably have a parameterized version of  `clusterResourceQuotaDeployments()` to be used in line 476.

alexeykazakov (approved):

alexeykazakov (approved):

PR #661, opened 2023-02-07
Test Proxy Space Lister

rajivnathan:
Tests the proxy space lister API

Paired with https://github.com/codeready-toolchain/registration-service/pull/301


test/e2e/proxy_test.go
```diff
@@ -231,14 +233,114 @@ func TestProxyFlow(t *testing.T) {
...
+				)
+
+				t.Run("successful list", func(t *testing.T) {
+					// when
+					workspaces := &toolchainv1alpha1.WorkspaceList{}
+					err = proxyCl.List(context.TODO(), workspaces)
+
+					// then
+					require.NoError(t, err)
+					require.Len(t, workspaces.Items, 1)
```
alexeykazakov, line 267: sWe need to cover the case when there is actually some workspace sharing and the list of available workspaces contains more than one workspace.

Let's make sure we test a real life scenarios.
Create multiple spaces & spacebindings. One user has access to a single workspace only. And at least two more users which share their spaces with each other. Then test that:
- each user gets expected list of spaces
- doesn't get the workspaces which the user doesn't have access to included into the result (so no noise workspaces included)
- same for GET specific workspace request.

rajivnathan: @alexeykazakov e2e tests have been updated

alexeykazakov:
/retest

after merging https://github.com/codeready-toolchain/member-operator/pull/408

rajivnathan:
/retest
after merging https://github.com/codeready-toolchain/host-operator/pull/742

PR #667, opened 2023-02-13
feat: add list of provisioned namespaces into Space status

mfrancisc:
Tests to check that `Space.Status.ProvisionedNamespace` list is populated as expected, containing the namespace name and type from the `NSTemplateSet` of the given Space object.

Jira Story: https://issues.redhat.com/browse/ASC-252

Related PRs:
- api: https://github.com/codeready-toolchain/api/pull/341

Paired with:
- https://github.com/codeready-toolchain/host-operator/pull/746


test/e2e/parallel/space_test.go
```diff
@@ -429,30 +429,34 @@ func TestSpaceNamespaces(t *testing.T) {
...
+		},
+		{
+			Name: space.Name + "-stage",
+		},
+	}
 
 	// then
 	// wait until NSTemplateSet has expected list of provisioned namespaces
-	// todo check also list of provisioned namespaces from space once functionality will be available
 	_, err := memberAwait.WaitForNSTmplSet(t, nsTmplSet.Name,
```
alexeykazakov, line 452: Do we really care about verifying the NSTemplateSets here?
Isn't the NSTemplateSet is just an implementation detail?
What we need is to make sure:
- All the end resources are provisioned for the space (Namespaces with all expected resources there, cluster scoped resources, etc). And this is what is already done in line 439 (`VerifyResourcesProvisionedForSpace()`), right?
- Check the Space status including the list of namespaces.

It doesn't heart to check the NsTemplateSet too but do we need it in our e2e tests for Spaces?
WDYT?

alexeykazakov (approved):
Overall looks good. I have one question though.

test/e2e/parallel/space_test.go
rajivnathan: Would it make sense to add a check in https://github.com/codeready-toolchain/toolchain-e2e/blob/master/testsupport/tiers/nstemplateset.go#L17 instead so that anything that calls the `VerifyNSTemplateSet` func will check the provisioned namespaces status too? That function has the slice of namespaces `expectedTemplateRefs.Namespaces` already so it should be easy to verify there?

testsupport/space.go
```diff
@@ -193,7 +193,11 @@ func verifyResourcesProvisionedForSpace(t *testing.T, hostAwait *wait.HostAwaiti
 	}
 
 	// get NSTemplateSet
-	nsTmplSet, err := targetCluster.WaitForNSTmplSet(t, spaceName, wait.UntilNSTemplateSetHasTier(tier.Name), wait.UntilNSTemplateSetHasConditions(Provisioned()))
+	nsTmplSet, err := targetCluster.WaitForNSTmplSet(t, spaceName,
+		wait.UntilNSTemplateSetHasTier(tier.Name),
+		wait.UntilNSTemplateSetHasConditions(Provisioned()),
+		wait.UntilNSTemplateSetProvisionedNamespacesIsNotEmpty(),
```
rajivnathan, line 199: If we go with the suggestion from https://github.com/codeready-toolchain/toolchain-e2e/pull/667/files#r1105280439 then would we still need this check?

rajivnathan (approved):
Looks good! Just the minor question related to Alexey's comment and then whether we still need the `UntilNSTemplateSetProvisionedNamespacesIsNotEmpty` func

test/e2e/parallel/space_test.go
mfrancisc: You are right, I simplified (hopefully) a lot the version of the tests, please see: [3a8e178](https://github.com/codeready-toolchain/toolchain-e2e/pull/667/commits/3a8e17812b60eafa5d2891bafce9f933a3dc9e55)

Basically as @rajivnathan suggested (thanks for this, haven't noticed that function that returns the list of namespaces provisioned from the tier template), now I check the provisioned namespaces in `expectedTemplateRefs.Namespaces` in the `VerifyNSTemplateSet` function , and I've removed all the "hard coded" checks for provisioned namesapces in all other places.

So in brief now,  if I did it correctly, `VerifyResourcesProvisionedForSpace` function checks the provisioned namespaces for both the `NSTemplateSet` and the `Space` , it checks this list against the namespaces that are coming from the tier template. No need to check those in the individual tests in theory. For the same reason I've deleted the previously created test `TestSpaceNamespaces`.

Thanks for your suggestions and let me know WDYT!

testsupport/space.go
```diff
@@ -205,14 +208,11 @@ func verifyResourcesProvisionedForSpace(t *testing.T, hostAwait *wait.HostAwaiti
...
-		// check that namespace list is available in the NSTemplate status
-		nsTmplSet, err = targetCluster.WaitForNSTmplSet(t, spaceName, wait.UntilNSTemplateSetHasProvisionedNamespaces([]toolchainv1alpha1.SpaceNamespace{{
-			Name: spaceName + "-tenant",
-			Type: "default", // default type should be there
-		}}))
-		require.NoError(t, err)
 	}
+	// wait for space to have list of provisioned namespaces in status
+	_, err = hostAwait.WaitForSpace(t, spaceName,
+		wait.UntilSpaceHasProvisionedNamespaces(nsTmplSet.Status.ProvisionedNamespaces))
```
rajivnathan, line 214: Let's add a comment here to mention the nsTmplSet ProvisionedNamespaces is checked as part of VerifyNSTemplateSet?

rajivnathan (approved):
Nice work!

testsupport/space.go
mfrancisc: makes sense! done [a7ebfe3](https://github.com/codeready-toolchain/toolchain-e2e/pull/667/commits/a7ebfe353db58e3c285fc10f9579800bd91e5401)

Thanks!

alexeykazakov (approved):

mfrancisc:
/retest

SpaceLister failure reported

PR #668, opened 2023-02-15
update proxy test to verify NotEmpty fields in Workspace objects

mfrancisc:
Some e2e jobs where failing while checking the match between `Workspace.ObjectMeta` fields and `Space.ObjectMeta` fields:

```
  host.go:1640: waiting for Space 'bus' with matching criteria
=== RUN   TestSpaceLister/bicycle_gets_workspaces/can_get_car_workspace
    host.go:1640: waiting for Space 'car' with matching criteria
    proxy_test.go:740: 
        	Error Trace:	/tmp/toolchain-e2e/test/e2e/proxy_test.go:740
        	            				/tmp/toolchain-e2e/test/e2e/proxy_test.go:471
        	Error:      	Not equal: 
        	            	expected: "75067"
        	            	actual  : "74984"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-75067
        	            	+74984
        	Test:       	TestSpaceLister/bicycle_gets_workspaces/can_get_car_workspace
```

Jira task: https://issues.redhat.com/browse/ASC-296

### DIFF
--- a/test/e2e/proxy_test.go
+++ b/test/e2e/proxy_test.go
@@ -744,6 +744,7 @@ func verifyHasExpectedWorkspace(t *testing.T, expectedWorkspace toolchainv1alpha
 			return
 		}
 	}
+
 	t.Errorf("expected workspace %s not found", expectedWorkspace.Name)
 }
 
--- a/test/e2e/proxy_test.go
+++ b/test/e2e/proxy_test.go
@@ -744,7 +744,6 @@ func verifyHasExpectedWorkspace(t *testing.T, expectedWorkspace toolchainv1alpha
 			return
 		}
 	}
-
 	t.Errorf("expected workspace %s not found", expectedWorkspace.Name)
 }
 
--- a/test/e2e/proxy_test.go
+++ b/test/e2e/proxy_test.go
@@ -736,7 +736,7 @@ func setStoneSoupConfig(t *testing.T, hostAwait *wait.HostAwaitility, memberAwai
 func verifyHasExpectedWorkspace(t *testing.T, expectedWorkspace toolchainv1alpha1.Workspace, actualWorkspaces ...toolchainv1alpha1.Workspace) {
 	for _, actualWorkspace := range actualWorkspaces {
 		if actualWorkspace.Name == expectedWorkspace.Name {
-			assert.NotEmpty(t, actualWorkspace.Status, "Workspace.Status field is empty", actualWorkspace, expectedWorkspace)
+			assert.Equal(t, expectedWorkspace.Status, actualWorkspace.Status)
 			assert.NotEmpty(t, actualWorkspace.ObjectMeta.ResourceVersion, "Workspace.ObjectMeta.ResourceVersion field is empty: %#v", actualWorkspace)
 			assert.NotEmpty(t, actualWorkspace.ObjectMeta.Generation, "Workspace.ObjectMeta.Generation field is empty: %#v", actualWorkspace)
 			assert.NotEmpty(t, actualWorkspace.ObjectMeta.CreationTimestamp, "Workspace.ObjectMeta.CreationTimestamp field is empty: %#v", actualWorkspace)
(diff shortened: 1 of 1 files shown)


test/e2e/proxy_test.go
```diff
@@ -736,11 +736,11 @@ func setStoneSoupConfig(t *testing.T, hostAwait *wait.HostAwaitility, memberAwai
 func verifyHasExpectedWorkspace(t *testing.T, expectedWorkspace toolchainv1alpha1.Workspace, actualWorkspaces ...toolchainv1alpha1.Workspace) {
 	for _, actualWorkspace := range actualWorkspaces {
 		if actualWorkspace.Name == expectedWorkspace.Name {
-			assert.Equal(t, expectedWorkspace.Status, actualWorkspace.Status)
-			assert.Equal(t, expectedWorkspace.ObjectMeta.ResourceVersion, actualWorkspace.ObjectMeta.ResourceVersion)
-			assert.Equal(t, expectedWorkspace.ObjectMeta.Generation, actualWorkspace.ObjectMeta.Generation)
-			assert.Equal(t, expectedWorkspace.ObjectMeta.CreationTimestamp, actualWorkspace.ObjectMeta.CreationTimestamp)
-			assert.Equal(t, expectedWorkspace.ObjectMeta.UID, actualWorkspace.ObjectMeta.UID)
+			assert.NotEmpty(t, actualWorkspace.Status, "Workspace.Status field is empty", actualWorkspace, expectedWorkspace)
```
alexeykazakov, line 739: We can't just check if the status is not empty. We need to check the actual values. The status has the list of namespaces and other critical information we need to test.

test/e2e/proxy_test.go
```diff
@@ -736,11 +736,11 @@ func setStoneSoupConfig(t *testing.T, hostAwait *wait.HostAwaitility, memberAwai
...
+			assert.NotEmpty(t, actualWorkspace.ObjectMeta.ResourceVersion, "Workspace.ObjectMeta.ResourceVersion field is empty: %#v", actualWorkspace)
```
rajivnathan, line 740: In these cases the actual value would be empty if it fails so there's not much value in having an explicit string for the failure case. Can simplify it to:
```suggestion
			assert.NotEmpty(t, actualWorkspace.ObjectMeta.ResourceVersion)
```

Same for the other checks below

test/e2e/proxy_test.go
mfrancisc: makes sense, I've restored the check in [ad0fb7e](https://github.com/codeready-toolchain/toolchain-e2e/pull/668/commits/ad0fb7e58930af89a4713957f287010f1a66888f)

test/e2e/proxy_test.go
mfrancisc: I'm printing to whole `actualWorkspace` object in case of an assertion error, so that all fields and values can be checked. 
But if it's not useful I can remove that.

test/e2e/proxy_test.go
mfrancisc: As I was trying to understand in the discussion on the slack channel, since the tests is first fetching the list of workspaces https://github.com/codeready-toolchain/toolchain-e2e/blob/614010ed3f9ce916589dc6d5036fa143b9f381ac/test/e2e/proxy_test.go#L384 
and then waiting for the space to match condition https://github.com/codeready-toolchain/toolchain-e2e/blob/614010ed3f9ce916589dc6d5036fa143b9f381ac/test/e2e/proxy_test.go#L751

I am afraid, that by the time the Space is retrieved, it may have a difference also in the `.Status` object. But maybe that's less frequent or I misunderstood how it works. 🤷‍♂️

rajivnathan: The Workspace.Status mirrors the following values from the Space resource: `Namespaces`, `Owner`, `Role`, and Type. The test calls `VerifyResourcesProvisionedForSignup` after creating each user and does not modify any of those Space values so I believe none of those should change after `listWorkspaces` is called. So it should be safe to keep the check.

The conversation on slack was more specifically about the `ObjectMeta` fields because those can be changed in many cases and is not really possible for us to predict so it's better to just check that they're not empty.

rajivnathan (approved):
Thanks for covering this! 👍

alexeykazakov (approved):

test/e2e/proxy_test.go
mfrancisc: Ok, thanks for the explanation, and makes sense that not changing the users after creation may not modify the Status object. I wanted to perform some other test runs, before I merge, so I'll push few "empty" commits into this PR.

test/e2e/proxy_test.go
mfrancisc: I personally found it useful to see all the values in case of an assertion failure , and not just the generic error about the field that was empty, with the line number. But we can remove those at any point if needed.

PR #677, opened 2023-03-08
fix toolchaincluster name logic

mfrancisc:
This PR is part of work done for Jira Story: [ASC-299](https://issues.redhat.com/browse/ASC-299)
Which is about the ability to run our e2e tests on hypershift clusters.

Changes applied: 
- The logic for generating toolchaincluster name needs to be updated so that it handles the case in which cluster URLS are longer then 63 chars, the naming mismatch was causing the deployment of 2 webhooks instances (one per each member namespace) and an invalid WebHookConfig . 

- hypershift cluster do not have `master` node usage metrics

Realted PR:
- https://github.com/codeready-toolchain/toolchain-cicd/pull/83


mfrancisc:
/retest

infra

make/test.mk
```diff
@@ -333,13 +333,16 @@ create-host-resources:
...
 		API_ENDPOINT=`oc get infrastructure cluster -o jsonpath='{.status.apiServerURL}'`; \
-		TOOLCHAIN_CLUSTER_NAME=`echo "$${API_ENDPOINT}" | sed 's/.*api\.\([^:]*\):.*/\1/'`; \
 		echo "API_ENDPOINT $${API_ENDPOINT}"; \
+		JOINING_CLUSTER_NAME=`echo "$${API_ENDPOINT}" | sed 's/^[^/]*\/\/\([^:/]*\)\(:.*\)\{0,1\}\(\/.*\)\{0,1\}$$/\1/' | sed 's/^api\.//'`; \
+		echo "Joining cluster name: $${JOINING_CLUSTER_NAME}"; \
+		TOOLCHAIN_CLUSTER_NAME=`echo "member-$${JOINING_CLUSTER_NAME:0:55}2"`; \
 		echo "TOOLCHAIN_CLUSTER_NAME $${TOOLCHAIN_CLUSTER_NAME}"; \
 		PATCH_FILE=/tmp/patch-toolchainconfig_${DATE_SUFFIX}.json; \
-		echo "{\"spec\":{\"members\":{\"specificPerMemberCluster\":{\"member-$${TOOLCHAIN_CLUSTER_NAME}2\":{\"webhook\":{\"deploy\":false}}}}}}" > $$PATCH_FILE; \
+		echo "{\"spec\":{\"members\":{\"specificPerMemberCluster\":{\"$${TOOLCHAIN_CLUSTER_NAME}\":{\"webhook\":{\"deploy\":false}}}}}}" > $$PATCH_FILE; \
```
MatousJobanek, line 345: what if we swapped the targets in https://github.com/codeready-toolchain/toolchain-e2e/blob/7c7b289e8b9b2f7ec3c735493f7ce5363493d9dc/make/test.mk#L261 so the `create-host-resources` is called after `setup-toolchainclusters`
In that case, we wouldn't need to replicate the logic of getting the name, but instead of it we would:
1. list ToolchainCluster CRs of the member type with the role tenant
2. find the one which ends with `2`
3. take the name out of it

mfrancisc: Good point! Let me try that out !

testsupport/wait/member.go
```diff
@@ -1562,14 +1562,14 @@ func UntilMemberStatusHasUsageSet() MemberStatusWaitCriterion {
...
-			return fmt.Sprintf("expected MemberStatus to have 'master' and 'worker' usages set: %v", actual.Status.ResourceUsage.MemoryUsagePerNodeRole)
+			return fmt.Sprintf("expected MemberStatus to have 'master' and/or 'worker' usages set: %v", actual.Status.ResourceUsage.MemoryUsagePerNodeRole)
 		},
 	}
 }
 
 func hasMemberStatusUsageSet(status toolchainv1alpha1.MemberStatusStatus) bool {
 	usage := status.ResourceUsage.MemoryUsagePerNodeRole
-	return len(usage) == 2 && usage["worker"] > 0 && usage["master"] > 0
+	return len(usage) >= 1 && usage["worker"] > 0
```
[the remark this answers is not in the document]
mfrancisc: Done please see: [6f34f7f](https://github.com/codeready-toolchain/toolchain-e2e/pull/677/commits/6f34f7faf418bb225155f9ae92a5003018fe3ccf)

make/test.mk
```diff
@@ -333,13 +333,12 @@ create-host-resources:
 	-oc create -f deploy/host-operator/${ENVIRONMENT}/ -n ${HOST_NS}
 	# patch toolchainconfig to prevent webhook deploy for 2nd member, a 2nd webhook deploy causes the webhook verification in e2e tests to fail
 	# since e2e environment has 2 member operators running in the same cluster
+	# for details on how the TOOLCHAINCLUSTER_NAME is composed see https://github.com/codeready-toolchain/toolchain-cicd/blob/master/scripts/add-cluster.sh
 	if [[ ${SECOND_MEMBER_MODE} == true ]]; then \
-		API_ENDPOINT=`oc get infrastructure cluster -o jsonpath='{.status.apiServerURL}'`; \
-		TOOLCHAIN_CLUSTER_NAME=`echo "$${API_ENDPOINT}" | sed 's/.*api\.\([^:]*\):.*/\1/'`; \
-		echo "API_ENDPOINT $${API_ENDPOINT}"; \
+		TOOLCHAIN_CLUSTER_NAME=`oc get toolchaincluster -l type=member -A --no-headers -o custom-columns=":metadata.name" | grep -E "2\>"`; \
```
MatousJobanek, line 338: we need to list only in host operator ns, plus I would filter only  should get only the `tenant` ToolchainCluster CRs - just in case someone would rerun the target against an already existing deployment/namespaces that would already contain the e2e ToolchainCluster CRs.
and as for the grep - doesn't it select a name that has `2` anywhere in the name? shouldn't it be:
```suggestion
		TOOLCHAIN_CLUSTER_NAME=`oc get toolchaincluster -l type=member,cluster-role.toolchain.dev.openshift.com/tenant=""  -n ${HOST_NS} --no-headers -o custom-columns=":metadata.name" | grep "2$"`; \
```

mfrancisc: Good point about adding the namespace.

About the `cluster-role.toolchain.dev.openshift.com/tenant=""` I'm not sure it's 100% required , if you are referring to `e2e-fmuntean-devsandbox-cluster1.devcluster.openshift.com2` type of `toolchaincluster` they do not have the label `type=member` but `type=e2e` :

```
 labels:
    namespace: toolchain-member-14130231
    ownerClusterName: member-fmuntean-devsandbox-cluster1.devcluster.openshift.com
    type: e2e
```
but probably I've misunderstood.

As per the `grep` part seems to be working the same way:
```
✗ oc get toolchaincluster -l type=member  -n toolchain-host-14130231 --no-headers -o custom-columns=":metadata.name" | grep "2$"
member-fmuntean-devsandbox-cluster1.devcluster.openshift.com2

✗ oc get toolchaincluster -l type=member  -n toolchain-host-14130231 --no-headers -o custom-columns=":metadata.name" | grep -E "2\>"
member-fmuntean-devsandbox-cluster1.devcluster.openshift.com2
```
 I can change it since your version seems more intuitive probably.

MatousJobanek: > About the cluster-role.toolchain.dev.openshift.com/tenant="" I'm not sure it's 100% required , if you are referring to e2e-fmuntean-devsandbox-cluster1.devcluster.openshift.com2 type of toolchaincluster they do not have the label type=member but type=e2e :

you are right :+1: , my bad :man_facepalming: I forgot that we use different label

as for grep:
```bash
$ echo "member-sandbox.openshiftapps.com2
member-sandbox-m2.openshiftapps.com" | grep -E "2\>"
member-sandbox.openshiftapps.com2
member-sandbox-m2.openshiftapps.com
```
vs
```bash
$ echo "member-sandbox.openshiftapps.com2
member-sandbox-m2.openshiftapps.com" | grep -E "2$"
member-sandbox.openshiftapps.com2
```

mfrancisc: I thought I tested that, good catch! 
I've hopefully reflected all the required changes in [29fe04d](https://github.com/codeready-toolchain/toolchain-e2e/pull/677/commits/29fe04dc3d43f0f5d04fca32f5186f221c963cdd)

Thanks for great review!

mfrancisc: hey @MatousJobanek I now remember why I ended up using `grep -E 2\>` instead of  `grep 2$` command. 
It's because was not working initially in make target, since it needs the extra $ simbol 😕 

Please see: [5927024](https://github.com/codeready-toolchain/toolchain-e2e/pull/677/commits/5927024164a577226c4d58533f6194871e2f0cee)

MatousJobanek (approved):

make/run-cicd-script.mk
```diff
@@ -1,5 +1,5 @@
 
-OWNER_AND_BRANCH_LOCATION=codeready-toolchain/toolchain-cicd/master
+OWNER_AND_BRANCH_LOCATION=mfrancisc/toolchain-cicd/feature/ASC-299_hypershift_cluster
```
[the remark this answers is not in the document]
mfrancisc: fixed in [d8055e5](https://github.com/codeready-toolchain/toolchain-e2e/pull/677/commits/d8055e53858dfb5693e551ad60321a12eea3981f)

PR #689, opened 2023-03-27
avoid space deletion when spacebinding is not created yet 

mfrancisc:
Observed a series of failures regarding e2e test regarding `TestCreateSpace`.

This PR updates `CreateSpace` helper function so that space and spacebinding creation is triggered in sequence without too much delay between them. 

failure log example https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/codeready-toolchain_host-operator/770/pull-ci-codeready-toolchain-host-operator-master-e2e/1639611710845227008/build-log.txt

```
        expected:
        ----
        [{Ready False 0001-01-01 00:00:00 +0000 UTC UnableToProvision unknown target member cluster 'unknown' <nil>}]----
        actual:
        ----
        [{Ready False 2023-03-25 13:48:57 +0000 UTC UnableToTerminate cannot delete NSTemplateSet: unknown target member cluster: 'unknown' <nil>}]----
        -expected
        +actual
          []v1alpha1.Condition{
          	{
          		Type:               "Ready",
          		Status:             "False",
        - 		LastTransitionTime: v1.Time{},
        + 		LastTransitionTime: v1.Time{Time: s"2023-03-25 13:48:57 +0000 UTC"},
        - 		Reason:             "UnableToProvision",
        + 		Reason:             "UnableToTerminate",
          		Message: strings.Join({
        + 			"cannot delete NSTemplateSet: ",
          			"unknown target member cluster",
        + 			":",
          			" 'unknown'",
          		}, ""),
          		LastUpdatedTime: nil,
          	},
          }
```

### DIFF
--- a/openshift-ci/Dockerfile.tools
+++ b/openshift-ci/Dockerfile.tools
@@ -9,8 +9,8 @@ ENV LANG=en_US.utf8 \
     PATH=$PATH:$GOPATH/bin \
     GIT_COMMITTER_NAME=devtools \
     GIT_COMMITTER_EMAIL=devtools@redhat.com \
-    GOLANG_VERSION=go1.17.13 \
-    GOLANG_SHA256=4cdd2bc664724dc7db94ad51b503512c5ae7220951cac568120f64f8e94399fc
+    GOLANG_VERSION=go1.18.8 \
+    GOLANG_SHA256=4d854c7bad52d53470cf32f1b287a5c0c441dc6b98306dea27358e099698142a
 
 ARG GO_PACKAGE_PATH=github.com/codeready-toolchain/toolchain-e2e
 


testsupport/space.go
```diff
@@ -84,16 +85,23 @@ func WithTierNameAndHashLabel(tierName, hash string) SpaceOption {
 	}
 }
 
+func WithCreationTimestamp(creationTimestamp time.Time) SpaceOption {
+	return func(s *toolchainv1alpha1.Space) {
+		s.CreationTimestamp = metav1.NewTime(creationTimestamp)
```
alexeykazakov, line 90: Does it actually work? I mean does kube allow specifying the creation timestamp? Are you sure it's not ignored?

test/e2e/parallel/space_test.go
```diff
@@ -47,9 +48,11 @@ func TestCreateSpace(t *testing.T) {
 
 	t.Run("failed to create space - unknown target member cluster", func(t *testing.T) {
 		// given & when
-		s, _, _ := CreateSpace(t, awaitilities, WithTierName("appstudio"), func(space *toolchainv1alpha1.Space) {
-			space.Spec.TargetCluster = "unknown"
-		})
+		s, _, _ := CreateSpace(t, awaitilities,
+			WithCreationTimestamp(time.Now().Add(time.Second*60)), // we set a time in the future, so that spacecleanup_controller doesn't delete the space before we actually create the spacebinding
```
alexeykazakov, line 52: As far as I remember there is already a delay in the space garbage collector which waits for 60 seconds (?) before deleting spaces without any space bindings... If I'm not mistaken.

testsupport/space.go
```diff
@@ -84,16 +85,23 @@ func WithTierNameAndHashLabel(tierName, hash string) SpaceOption {
...
 // CreateSpace initializes a new Space object using the NewSpace function, and then creates it in the cluster
 // It also automatically provisions MasterUserRecord and creates SpaceBinding for it
 func CreateSpace(t *testing.T, awaitilities wait.Awaitilities, opts ...SpaceOption) (*toolchainv1alpha1.Space, *toolchainv1alpha1.UserSignup, *toolchainv1alpha1.SpaceBinding) {
 	space := NewSpace(t, awaitilities, opts...)
 	err := awaitilities.Host().CreateWithCleanup(t, space)
 	require.NoError(t, err)
-	space, err = awaitilities.Host().WaitForSpace(t, space.Name, wait.UntilSpaceHasAnyTargetClusterSet(), wait.UntilSpaceHasAnyTierNameSet())
-	require.NoError(t, err)
 	// we also need to create a MUR & SpaceBinding, otherwise, the Space could be automatically deleted by the SpaceCleanup controller
 	signup, mur, spaceBinding := CreateMurWithAdminSpaceBindingForSpace(t, awaitilities, space, true)
```
alexeykazakov, line 101: I wander if we can create a space binding first and then a space...

alexeykazakov (approved):
Looks good. However I have a few theoretical questions :)

testsupport/space.go
mfrancisc: right! locally was working but as you said the creationTimestamp it's being changed by the apiServer just before storing the object into etcd, so was just luck :D

test/e2e/parallel/space_test.go
mfrancisc: I had to increase the deletionThreshold in the spacecleanup_controller see https://github.com/codeready-toolchain/host-operator/pull/776 , but not sure it's the same delay you are referring to.

testsupport/space.go
mfrancisc: Actually I tried doing that but if I'm not wrong, it doesn't seem possible because `NewSpaceBinding` needs the name of the space https://github.com/codeready-toolchain/toolchain-e2e/blob/210fe8e9eeb171b1bb476197922e1094010ec144/testsupport/spacebinding.go#L43

and that only get's generated by the api server once the space creation was triggered (since some times we use `GenerateName` function)

test/e2e/parallel/space_test.go
alexeykazakov: Yes, it's the same delay.

mfrancisc:
Sorry for the confusion, but I had to update this PR, I got some false positives initially while running the tests locally.
Once I've merged the changes to my local [branch where I'm adding more parallel tests](https://github.com/codeready-toolchain/toolchain-e2e/pull/685) I've noticed that it was still failing.
 After trying multiple things, this is apparently the most reliable solution that I could get. Which also implies a change in `spacecleanup_controller.go` in host-operator.

testsupport/space.go
```diff
@@ -87,13 +87,26 @@ func WithTierNameAndHashLabel(tierName, hash string) SpaceOption {
...
+		RequireConditions(ConditionSet(Default(), ApprovedByAdmin())...).
+		NoSpace().
+		WaitForMUR().Execute(t).Resources()
+	t.Logf("The UserSignup %s and MUR %s were created", signup.Name, mur.Name)
+	// then create the actual space
 	space := NewSpace(t, awaitilities, opts...)
 	err := awaitilities.Host().CreateWithCleanup(t, space)
 	require.NoError(t, err)
+	// create spacebinding request immediately after
+	spaceBinding := CreateSpaceBinding(t, awaitilities.Host(), mur, space, "admin")
```
mfrancisc, line 105: An alternative to this logic, could be to try and issue a dryRun creation of the space, so that kube api server returns an actual unique space object generated name, and use that name for the usersignup and spacebinding, and only after those two are created trigger the actual space creation request.

alexeykazakov: Let's stick with the current solution. If it doesn't help then we could try to use dryRun...

mfrancisc:
/retest

infra

mfrancisc:
@MatousJobanek, as discussed in our 1:1 call, I've added the polling logic to space and space binding creation, and also closed the paired PR in host operator that was increasing the deletion threshold in spacecleanup_controller. 

I'm noticing that a metric assertion related to spaces has failed:
`waited for metric 'sandbox_spaces_current{[cluster_name member-a7f79a01e0b0a415fb33dc99c1130eb0-0c759b3c94722f31.elb.u1]}' to reach '1'. Current value: 0`

Maybe it's due to some small delay between the metric being updated in Prometheus and the test reading it. 

Or maybe I should try and introduce in this PR also the space cleanup in `clean.go` functions that we talked about ?

mfrancisc:
One thing I've noticed locally is that, in case of test failure, there are no spaces left behind, only few usersignups and murs:

```
➜ k get space
No resources found in toolchain-host-29093938 namespace.
➜  k get usersignup
NAME                                                  USERNAME                                              COMPLETE   REASON   COMPLIANTUSERNAME                                     EMAIL
for-space-migration-appstudio-provisioned-space       for-space-migration-appstudio-provisioned-space       True                for-space-migration-appstudio-provisioned-space       for-space-migration-appstudio-provisioned-space@acme.com
for-space-migration-second-member-provisioned-space   for-space-migration-second-member-provisioned-space   True                for-space-migration-second-member-provisioned-space   for-space-migration-second-member-provisioned-space@acme.com
➜ k get mur
NAME                                                  READY   REASON        CLUSTER                                                         TIER
for-space-migration-appstudio-provisioned-space       True    Provisioned   member-fmuntean-devsandbox-cluster15.devcluster.openshift.com   deactivate30
for-space-migration-second-member-provisioned-space   True    Provisioned   member-fmuntean-devsandbox-cluster15.devcluster.openshift.com   deactivate30
```
which should be cleaned up by the migration test from what I can see in the code, but apparently this is not happening.

testsupport/wait/host.go
```diff
@@ -2166,3 +2166,40 @@ func EncodeUserIdentifier(subject string) string {
...
+		space, err = a.WaitForSpace(t, space.Name, UntilSpaceHasAnyTargetClusterSet(), UntilSpaceHasAnyTierNameSet())
+		if err != nil {
+			return false, err
+		}
+		// let's see if spacebinding was provisioned as expected
+		spaceBinding, err = a.WaitForSpaceBinding(t, mur.Name, space.Name,
+			UntilSpaceBindingHasMurName(mur.Name),
+			UntilSpaceBindingHasSpaceName(space.Name),
+			UntilSpaceBindingHasSpaceRole(spaceRole),
+		)
```
MatousJobanek, line 2197: Let's only check the existence in this poll loop - because this is what matters now (if it was successfully created or unexpectedly deleted). And move the verification of the content out of this function.

MatousJobanek:
>Maybe it's due to some small delay between the metric being updated in Prometheus and the test reading it.

@mfrancisc don't forget to delete this branch https://github.com/mfrancisc/host-operator/tree/fixcreatespace from your fork.
until it will be in your fork present, the PR pairing logic will keep using it, which means that it will also use the longer deletion threshold.

> Or maybe I should try and introduce in this PR also the space cleanup in clean.go functions that we talked about ?

you can try that for sure.

> One thing I've noticed locally is that, in case of test failure, there are no spaces left behind, only few usersignups and murs:

yeah, they should have been definitely cleaned up

MatousJobanek:
ok, I see where the issue is with the uncleaned MURs & UserSignups

there is hard-coded `false` in this call
https://github.com/codeready-toolchain/toolchain-e2e/blob/master/test/migration/setup_runner.go#L71
but should be:
```
test.CreateMurWithAdminSpaceBindingForSpace(t, r.Awaitilities, space, r.WithCleanup)
```

mfrancisc:
@MatousJobanek 
> ok, I see where the issue is with the uncleaned MURs & UserSignups
> there is hard-coded false in this call
> https://github.com/codeready-toolchain/toolchain-e2e/blob/master/test/migration/setup_runner.go#L71

Good point, let me try that out! 

I now see that `setup_migration_test` is supposed to have `WithCleanup=false` but `verify_migration_test` is expected to run after with `WithCleanup=true` in order to cleanup everything, which was not happening.

Thanks!

testsupport/wait/host.go
MatousJobanek: Yes, let's check:
* it's present and not terminating
* if it's terminating, then wait until it is gone and do the loop again.

mfrancisc: I did that plus the other fixes (cleanup function for spaces and fixed the `r.WithCleanup` param in setup_runner.go)

Please check:  https://github.com/codeready-toolchain/toolchain-e2e/pull/689/commits/5654e56aad5992b6caa3f7de34f1cb9f76911220

I've still faced one issue locally, this time with the `user_workloads_test` 😕 :
```
r_workloads_test.go:78:
        	Error Trace:	/Users/fmuntean/go/src/github.com/codeready-toolchain/toolchain-e2e/test/e2e/user_workloads_test.go:78
        	            	notifications.toolchain.dev.openshift.com "test-idler-dev-idled" not found
        	Test:       	TestUserWorkloads/TestIdlerAndPriorityClass
```
but let's see if this happens also in openshift-ci job. Meantime I'm running more tests locally to check the overall stability.

mfrancisc:
Unbelievable!!!! first passing `e2e` job I see since days 😲

testsupport/wait/host.go
```diff
@@ -2166,3 +2166,40 @@ func EncodeUserIdentifier(subject string) string {
...
+// We are creating both of them (Space and SpaceBinding) at the same time , with polling logic, so that we mitigate the issue with spacecleanup_controller deleting the Space before we create it's SpaceBinding.
+func (a *HostAwaitility) CreateSpaceAndSpaceBinding(t *testing.T, mur *toolchainv1alpha1.MasterUserRecord, space *toolchainv1alpha1.Space, spaceRole string, terminatingSpace toolchainv1alpha1.Condition) (*toolchainv1alpha1.Space, *toolchainv1alpha1.SpaceBinding, error) {
+	var spaceBinding *toolchainv1alpha1.SpaceBinding
+	err := wait.Poll(a.RetryInterval, a.Timeout, func() (done bool, err error) {
+		// create the space
+		if err := a.CreateWithCleanup(t, space); err != nil {
+			if errors.IsNotFound(err) {
+				return false, nil
+			}
+			return false, err
```
MatousJobanek, line 2180: when we create the Space, then I don't think that we could get "not found" error
we should rather check if it already exists
```suggestion
			if !errors.IsAlreadyExists(err) {
				return false, err
			}
```

testsupport/wait/host.go
```diff
@@ -2166,3 +2166,40 @@ func EncodeUserIdentifier(subject string) string {
...
+		}
+		// create spacebinding request immediately after ...
+		spaceBinding = spacebinding.NewSpaceBinding(mur, space, spaceRole)
+		if err := a.CreateWithCleanup(t, spaceBinding); err != nil {
+			return false, err
+		}
+		// let's see if space was provisioned as expected
+		space, err = a.WaitForSpace(t, space.Name)
+		if err != nil {
+			return false, err
```
MatousJobanek, line 2190: but we should check the "not found" here
```suggestion
			if errors.IsNotFound(err) {
				return false, nil
			}
			return false, err
```

testsupport/wait/host.go
```diff
@@ -2166,3 +2166,40 @@ func EncodeUserIdentifier(subject string) string {
...
+		}
+		if test.ContainsCondition(space.Status.Conditions, terminatingSpace) {
+			// space is in terminating let's wait until is gone and recreate it ...
+			return false, nil
+		}
+		// let's see if spacebinding was provisioned as expected
+		spaceBinding, err = a.WaitForSpaceBinding(t, mur.Name, space.Name)
+		if err != nil {
+			return false, err
```
MatousJobanek, line 2199:
```suggestion
			if errors.IsNotFound(err) {
				return false, nil
			}
			return false, err
```

testsupport/wait/host.go
```diff
@@ -2166,3 +2166,40 @@ func EncodeUserIdentifier(subject string) string {
...
+		}
+		// let's see if space was provisioned as expected
+		space, err = a.WaitForSpace(t, space.Name)
+		if err != nil {
+			return false, err
+		}
+		if test.ContainsCondition(space.Status.Conditions, terminatingSpace) {
+			// space is in terminating let's wait until is gone and recreate it ...
+			return false, nil
+		}
```
MatousJobanek, line 2195: let's rather rely on the deletion timestamp than the condition.
and you can right away wait until it's gone
```suggestion
		if util.IsBeingDeleted(space) {
			// space is in terminating let's wait until is gone and recreate it ...
			return false, a.WaitUntilSpaceAndSpaceBindingsDeleted(t, space.Name)
		}
```

testsupport/wait/host.go
mfrancisc: Good catch! fixed in https://github.com/codeready-toolchain/toolchain-e2e/pull/689/commits/6810c96b1fd58e73f4637c474d4a36135a1419c7

testsupport/wait/host.go
mfrancisc: applied in https://github.com/codeready-toolchain/toolchain-e2e/pull/689/commits/6810c96b1fd58e73f4637c474d4a36135a1419c7

testsupport/wait/host.go
mfrancisc: applied in https://github.com/codeready-toolchain/toolchain-e2e/pull/689/commits/6810c96b1fd58e73f4637c474d4a36135a1419c7

testsupport/wait/host.go
mfrancisc: Nice ! applied in https://github.com/codeready-toolchain/toolchain-e2e/pull/689/commits/6810c96b1fd58e73f4637c474d4a36135a1419c7

Thanks!

MatousJobanek (approved):
Thanks a lot for fixing the flaky tests :+1

PR #690, opened 2023-03-29
Add e2e test for web console plugin

sbryzak:
This PR adds tests to confirm that the web console plugin is deployed correctly and serving its resources as expected.

Current state: still a WIP as the web console endpoints require authorization and the token used to sign up the user isn't sufficient for invoking the content server.


test/e2e/web_console_plugin_test.go
```diff
@@ -0,0 +1,64 @@
...
+		Execute(s.T()).GetSignupResponse(s.T())
+
+	// Response should contain a ConsoleURL with a value something like:
+	// https://console-openshift-console.apps.99b682869228f7464338-mgmt.ci.hypershift.devcluster.openshift.com/
+	require.Contains(s.T(), response, "consoleURL")
+
+	consoleURL := response["consoleURL"]
+	manifestURL := fmt.Sprintf("%s%s", consoleURL, "api/plugins/toolchain-member-web-console-plugin/plugin-manifest.json")
+
+	resp, err := http.Get(manifestURL)
```
alexeykazakov, line 57: I think before calling the service endpoint we should wait for the health endpoint returning 200 to make sure the service is fully deployed. Just to avoid any test flakiness.

sbryzak: Have added a check that will retry a few times if the health check endpoint fails.

test/e2e/web_console_plugin_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+	// First perform a health check - we will attempt up to 5 times to invoke the health check endpoint without error
+	var err error
+	var resp *http.Response
+
+	req, err := http.NewRequest("GET", healthCheckURL, nil)
+	require.NoError(s.T(), err)
+	req.Header.Set("Authorization", signupRequest.GetToken())
+
+	pollResult, err := testutil.Poll(time.Second*5, time.Minute, func() (bool, error) {
```
mfrancisc, line 81: very minor thing -  I saw in other tests we usually used `wait.Poll` function from `k8s.io/apimachinery/pkg/util/wait`.

I'm not sure if there is any difference between those.

test/e2e/web_console_plugin_test.go
```diff
@@ -0,0 +1,104 @@
...
+	})
+	require.NoError(s.T(), err)
+	require.True(s.T(), pollResult)
+	require.Equal(s.T(), http.StatusOK, resp.StatusCode)
+
+	req, err = http.NewRequest("GET", manifestURL, nil)
+	require.NoError(s.T(), err)
+	req.Header.Set("Authorization", signupRequest.GetToken())
+
+	resp, err = httpClient.Do(req)
```
mfrancisc, line 96: should we wrap also this http call in a Poll loop, or should this one work at first attempt? I'm thinking about potential networking issues like connection drops ...

test/e2e/web_console_plugin_test.go
sbryzak: Good catch, looks like i accidentally imported the wrong one.

test/e2e/web_console_plugin_test.go
sbryzak: i think if it gets past the first health check ok then this should be fine too, I wouldn't expect any connection issues once we've confirmed that the route is active.

test/e2e/web_console_plugin_test.go
```diff
@@ -0,0 +1,153 @@
...
+		tr = &http.Transport{
+			TLSClientConfig: &tls.Config{
+				InsecureSkipVerify: true,
+			},
+		}
+	}
+
+	httpClient := &http.Client{Transport: tr}
+
+	// First perform a health check - we will attempt up to 5 times to invoke the health check endpoint without error
```
alexeykazakov, line 123:
```suggestion
	// First perform a health check - if the health check doesn't pass within one minute then the test will fail
```

alexeykazakov (approved):
Looks good! Just one minor correction in the comment.

test/e2e/web_console_plugin_test.go
```diff
@@ -0,0 +1,153 @@
...
+	wait.Awaitilities
+}
+
+func TestRunWebConsolePluginTest(t *testing.T) {
+	suite.Run(t, &webConsolePluginTest{})
+}
+
+func (s *webConsolePluginTest) SetupSuite() {
+	s.Awaitilities = WaitForDeployments(s.T())
+}
```
MatousJobanek, line 33: Please, stop using the test suites - it makes impossible to parallelize the tests. It also makes pretty hard to debug it and investigate any issues.

test/e2e/web_console_plugin_test.go
```diff
@@ -0,0 +1,153 @@
...
+
+func (s *webConsolePluginTest) TearDownTest() {
+	hostAwait := s.Host()
+	memberAwait := s.Member1()
+	memberAwait2 := s.Member2()
+	hostAwait.Clean(s.T())
+	memberAwait.Clean(s.T())
+	memberAwait2.Clean(s.T())
```
MatousJobanek, line 41: I'm not sure if we need to call the clean when the clean functions are already registered as part of the creation step

test/e2e/web_console_plugin_test.go
```diff
@@ -0,0 +1,153 @@
...
+		Execute(s.T())
+	response := signupRequest.GetSignupResponse(s.T())
+
+	_, mur := signupRequest.Resources()
+
+	memberAwait := GetMurTargetMember(s.T(), s.Awaitilities, mur)
+
+	// Response should contain a ConsoleURL with a value something like:
+	// https://console-openshift-console.apps.99b682869228f7464338-mgmt.ci.hypershift.devcluster.openshift.com/
+	require.Contains(s.T(), response, "consoleURL")
```
MatousJobanek, line 63: Could you please clarify what you verify here within the context of this test?

test/e2e/web_console_plugin_test.go
```diff
@@ -0,0 +1,153 @@
...
+			TLS: &v1.TLSConfig{
+				Termination: "reencrypt",
+				Certificate: string(secret.Data["tls.crt"]),
+				Key:         string(secret.Data["tls.key"]),
+			},
+			WildcardPolicy: "None",
+		},
+	}
+
+	err = memberAwait.Create(s.T(), route)
```
MatousJobanek, line 100: let's create it with the cleanup - we don't need the route to stay there after finishing the test
```go
err = memberAwait.CreateWithCleanup(s.T(), route)
```

test/e2e/web_console_plugin_test.go
```diff
@@ -0,0 +1,153 @@
...
+	httpClient := &http.Client{Transport: tr}
+
+	// First perform a health check - if the health check doesn't pass within one minute then the test will fail
+	var healthCheckResponse *http.Response
+
+	req, err := http.NewRequest("GET", healthCheckURL, nil)
+	require.NoError(s.T(), err)
+	req.Header.Set("Authorization", signupRequest.GetToken())
+
+	err = wait2.Poll(time.Second*5, time.Minute, func() (bool, error) {
```
MatousJobanek, line 130: why setting  interval to 5 seconds? why not one second?

test/e2e/web_console_plugin_test.go
```diff
@@ -0,0 +1,153 @@
...
+		healthCheckResponse, err = httpClient.Do(req) //nolint
+		if err != nil {
+			return false, err
+		}
+		defer healthCheckResponse.Body.Close()
+		return true, nil
+	})
```
MatousJobanek, line 137: what is the point of running the `Poll` if the only values that you return are:
* `false, err` (which means that it immediately ends because of the error)
* `true, nil` (which means that it immediately ends because you return true)

in other words, there will be only one execution of the logic

test/e2e/web_console_plugin_test.go
```diff
@@ -0,0 +1,153 @@
...
+	require.NoError(s.T(), err)
+	require.Equal(s.T(), http.StatusOK, healthCheckResponse.StatusCode, "error polling health check endpoint", route)
```
MatousJobanek, line 139: you maybe wanted to add this into the poll function

testsupport/signup_request.go
```diff
@@ -112,6 +112,39 @@ func (r *SignupRequest) Resources() (*toolchainv1alpha1.UserSignup, *toolchainv1
...
+	if r.noSpace {
+		queryParams["no-space"] = "true"
+	}
+
+	// Call the signup GET endpoint
+	response := invokeEndpoint(t, "GET", hostAwait.RegistrationServiceURL+"/api/v1/signup",
+		r.token, "", http.StatusOK, queryParams)
+
+	return response
+}
```
MatousJobanek, line 146: why do we need to execute this here? How is it different compared to what we have in `Execute()` method?
If we are missing test that verifies the response of the `GET` request, then it should go into the `registration_service_test.go` not here. However, I believe that we have all needed verification already there, right?

MatousJobanek (commented):
Please, add new tests into the parallelized ones if there is no big reason for not doing so.

PR #691, opened 2023-03-29
Update tests for storage limit increase

rajivnathan:
Related PR: https://github.com/codeready-toolchain/host-operator/pull/778


alexeykazakov (approved):
Looks good but it looks like you missed line 340.

rajivnathan:
Good catch! Fixed

PR #692, opened 2023-03-30
verify appstudio environment tier

MatousJobanek:
https://github.com/codeready-toolchain/host-operator/pull/777


alexeykazakov (approved):

rajivnathan (approved):

testsupport/wait/templateref_parser.go
```diff
@@ -0,0 +1,19 @@
+package wait
+
+import (
+	"fmt"
+	"strings"
+)
+
+// TierAndType parses the templateRef and extracts the `tier` and `type` values out of it
+// returns an error if this TemplateRef's format is invalid
+func TierAndType(templateRef string) (string, string, error) { // nolint:unparam
```
mfrancisc, line 10: I like more the new function name and file name! 👍 

Was just curious if there was any issue with the existing logic of the previously called `Split` function which was using `strings.SplintN` to split the string in N tokens?

mfrancisc (approved):
Nice work! 👍 

just left one small question.

testsupport/wait/templateref_parser.go
MatousJobanek: I had to change the function because previously it expected that the tier name shouldn't contain a hyphen within its name. It split the template ref by the hyphens:
```
<tier>-<type>-<based-on-tier-revision>-<template-revision>
```
which caused issues when I introduced the `appstudio-env` tier. I changed it so it allows hyphens to be part of the tier name.

PR #701, opened 2023-04-06
feat(STONEINTG-343): set ResourceQuota for Appstudio CRDs

dirgim:
* Add e2e tests for the new resource quotas

### DIFF
--- a/testsupport/tiers/checks.go
+++ b/testsupport/tiers/checks.go
@@ -428,6 +428,13 @@ func (a *appstudioTierChecks) GetNamespaceObjectChecks(_ string) []namespaceObje
 		memberOperatorSaReadRoleBinding(),
 		gitOpsServiceLabel(),
 		environment("development"),
+		resourceQuotaAppstudioCrds("512", "512", "512"),
+		resourceQuotaAppstudioCrdsBuild("512"),
+		resourceQuotaAppstudioCrdsGitops("512", "512", "512", "512", "512"),
+		resourceQuotaAppstudioCrdsIntegration("512", "512", "512"),
+		resourceQuotaAppstudioCrdsRelease("512", "512", "512", "512", "512"),
+		resourceQuotaAppstudioCrdsEnterpriseContract("512"),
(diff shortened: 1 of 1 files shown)


testsupport/tiers/checks.go
```diff
@@ -511,6 +511,13 @@ func (a *appstudioEnvTierChecks) GetNamespaceObjectChecks(_ string) []namespaceO
 		resourceQuotaComputeDeploy("20", "32Gi", "1750m", "32Gi"),
 		zeroResourceQuotaComputeBuild(),
 		resourceQuotaStorage("50Gi", "50Gi", "50Gi", "12"),
+		resourceQuotaAppstudioCrds("512", "512", "512"),
```
alexeykazakov, line 514: Wrong tier. Should be in `func (a *appstudioTierChecks)` Not in `func (a *appstudioEnvTierChecks)`

dirgim: The new resource quotas are supposed to be per-namespace, isn't `appstudioTierChecks` meant for cluster-scoped checks?

alexeykazakov: No. Each tier has its own struct and a set of methods to check different resources in that tier.
You added your checks to the `func (a *appstudioEnvTierChecks) GetNamespaceObjectChecks(_ string)` method. Which returns the checks for namespace scoped objects in the `appstudio-env` tier.
But instead you need to add those checks to the `func (a *appstudioTierChecks) GetNamespaceObjectChecks(_ string)` method which is for the namespace scoped object checks for the `appstudio` tier.

`appstudio` and `appstudio-env`are two different tiers. The former is for appstudio workspaces and the later is for future environments in workspaces (like dev, stage, prod, etc).

dirgim: Ah, I understand now, thanks. I moved the checks to the correct function now.

alexeykazakov (approved):
Looks good now. But could you please avoid pushing with force to PR branches in the future? It would help with follow up review when it's enough to review the follow up commits only instead of re-review the entire PR over again.
Thank you! :)

PR #705, opened 2023-04-14
fix: get Space & SpaceBinding directly in a loop

MatousJobanek:
There are two reasons:
1. the `Wait*` functions don't return "NotFound" error if the poll times out, which results in a failed test if the Space/SpaceBinding wasn't found
2.  we don't have to wait the whole period - it's just there or not, so getting it directly is more suitable in this case

an example of the failed test: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/codeready-toolchain_toolchain-e2e/704/pull-ci-codeready-toolchain-toolchain-e2e-master-e2e/1646501903141441536

```
        	Error Trace:	/go/src/github.com/codeready-toolchain/toolchain-e2e/test/e2e/parallel/space.go:97
        	            				/go/src/github.com/codeready-toolchain/toolchain-e2e/test/e2e/parallel/space_cleanup_test.go:138
        	            				/go/src/github.com/codeready-toolchain/toolchain-e2e/test/e2e/parallel/space_cleanup_test.go:30
        	Test:       	TestSpaceAndSpaceBindingCleanup/for_SpaceBinding/when_space_is_deleted
```

### DIFF
--- a/test/e2e/parallel/space_cleanup_test.go
+++ b/test/e2e/parallel/space_cleanup_test.go
@@ -27,7 +27,7 @@ func TestSpaceAndSpaceBindingCleanup(t *testing.T) {
 	t.Run("for SpaceBinding", func(t *testing.T) {
 		t.Run("when space is deleted", func(t *testing.T) {
 			// given
-			space, _, spaceBinding := setupForSpaceBindingCleanupTest(t, awaitilities, memberAwait, "joe", "redhat")
+			space, _, spaceBinding := setupForSpaceBindingCleanupTest(t, awaitilities, memberAwait, "joe", "for-redhat")
 
 			// when
 			err := hostAwait.Client.Delete(context.TODO(), space)
@@ -40,7 +40,7 @@ func TestSpaceAndSpaceBindingCleanup(t *testing.T) {
 
 		t.Run("when mur is deleted", func(t *testing.T) {
 			// given
-			_, userSignup, spaceBinding := setupForSpaceBindingCleanupTest(t, awaitilities, memberAwait, "lara", "ibm")
+			_, userSignup, spaceBinding := setupForSpaceBindingCleanupTest(t, awaitilities, memberAwait, "lara", "for-ibm")
 
 			// when
 			// deactivate the UserSignup so that the MUR will be deleted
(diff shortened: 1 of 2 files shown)


test/e2e/parallel/space_cleanup_test.go
```diff
@@ -27,7 +27,7 @@ func TestSpaceAndSpaceBindingCleanup(t *testing.T) {
 	t.Run("for SpaceBinding", func(t *testing.T) {
 		t.Run("when space is deleted", func(t *testing.T) {
 			// given
-			space, _, spaceBinding := setupForSpaceBindingCleanupTest(t, awaitilities, memberAwait, "joe", "redhat")
+			space, _, spaceBinding := setupForSpaceBindingCleanupTest(t, awaitilities, memberAwait, "joe", "for-redhat")
```
MatousJobanek, line 30: we should not use `redhat` as the name (or just a prefix) of the namespaces. It shouldn't be a big deal in OCP, but let's try to avoid it also here

testsupport/wait/host.go
```diff
@@ -2002,32 +2002,41 @@ func (a *HostAwaitility) WaitForSubSpace(t *testing.T, spaceRequestName, spaceRe
...
+		toolchainv1alpha1.SpaceBindingMasterUserRecordLabelKey: murName,
+		toolchainv1alpha1.SpaceBindingSpaceLabelKey:            spaceName,
+	}
+
+	spaceBindingList := &toolchainv1alpha1.SpaceBindingList{}
+	if err := a.Client.List(context.TODO(), spaceBindingList, client.MatchingLabels(labels), client.InNamespace(a.Namespace)); err != nil {
+		return nil, err
+	}
+	if len(spaceBindingList.Items) == 0 {
+		return nil, nil
```
mfrancisc, line 2034: should we return an error here, or in the caller check if returned spaceBinding is `nil` and return a `false` to the `wait.Poll` function?

mfrancisc: makes sense, also the caller can decide what he wants todo in case of `nil` SpaceBinding.

Thanks for fixing it.

MatousJobanek:
/retest
infra

rajivnathan (approved):

PR #710, opened 2023-04-20
fix & improve WebConsolePlugin test

MatousJobanek:
a few fixes & improvements of WebConsolePlugin test
* enable web console plugin in second member explicitly 
* dropped TestSuites
* move to parallel
* set the target cluster of the user explicitly
* iterate over both cluster so it's verified for both of them
* dropped useless poll function
* optimized `WaitForSecret` method by dropping the namespace param
* added more comprehensive checks of the created manifests
* dropped unnecessary code

paired PR https://github.com/codeready-toolchain/member-operator/pull/431


test/e2e/parallel/web_console_plugin_test.go
```diff
@@ -0,0 +1,120 @@
...
+		defer healthCheckResponse.Body.Close()
+		require.Equal(t, http.StatusOK, healthCheckResponse.StatusCode, "error polling health check endpoint", route)
+
+		req, err = http.NewRequest("GET", manifestURL, nil)
+		require.NoError(t, err)
+		req.Header.Set("Authorization", signupRequest.GetToken())
+
+		manifestResponse, err := httpClient.Do(req)
+		require.NoError(t, err)
+		defer manifestResponse.Body.Close()
```
mfrancisc, line 113: minor detail - should we check the `manifestResponse.StatusCode` as well ?

testsupport/wait/member.go
```diff
@@ -1259,14 +1259,14 @@ func (a *MemberAwaitility) WaitForConfigMap(t *testing.T, namespace, name string
 	return cm, err
 }
 
-// WaitForSecret waits until a Secret with the given name exists in the given namespace
-func (a *MemberAwaitility) WaitForSecret(t *testing.T, namespace, name string) (*corev1.Secret, error) {
```
mfrancisc, line 1263: IMO keeping the namespace as a parameter would make this function reusable also for other use cases. I'm thinking for example at the spacerequest test which in the future may need to check secrets in other namespaces. But not a big issue.

MatousJobanek: if needed, then we can change it... In the case it should be reusable, then it shouldn't be tied to any awaitility - just a standalone functino

test/e2e/parallel/web_console_plugin_test.go
MatousJobanek: why not: a02cdbe0

testsupport/wait/member.go
mfrancisc: Makes sense!  

So `MemberAwaitility` is mainly designed to interact with the member operator namespaces only. I was thinking at the multi-cluster scenario/environment, where member is a cluster not only a namespace , but we don't support that yet in e2e tests.

mfrancisc (approved):

mfrancisc:
/retest

I'm wondering what is this new infra issue `failed to create hosted zone: TooManyHostedZones: Limits Exceeded: MAX_HOSTED_ZONES_BY_OWNER - Cannot create more hosted zones.` 

I can ask test platform team in case there's a misconfiguration or something ...

test/e2e/parallel/web_console_plugin_test.go
```diff
@@ -0,0 +1,122 @@
...
+		require.Equal(t, http.StatusOK, healthCheckResponse.StatusCode, "error polling console plugin manifests", route)
```
alexeykazakov, line 115:
```suggestion
		require.Equal(t, http.StatusOK, healthCheckResponse.StatusCode, "error calling console plugin manifests", route)
```

alexeykazakov (approved):
Looks good! Just a few minor comments.

test/e2e/parallel/web_console_plugin_test.go
```diff
@@ -0,0 +1,122 @@
...
+			TargetCluster(memberAwait).
+			ManuallyApprove().
+			EnsureMUR().
+			RequireConditions(ConditionSet(Default(), ApprovedByAdmin())...).
+			Execute(t)
+		response := signupRequest.GetSignupResponse(t)
+
+		// Response should contain a ConsoleURL with a value something like:
+		// https://console-openshift-console.apps.99b682869228f7464338-mgmt.ci.hypershift.devcluster.openshift.com/
+		require.Contains(t, response, "consoleURL")
```
rajivnathan, line 36: Is the case correct? `ConsoleURL` in the comment vs `consoleURL` in the assertion

test/e2e/parallel/web_console_plugin_test.go
```diff
@@ -0,0 +1,122 @@
...
+				Name:      "consolepluginroute",
+				Namespace: memberAwait.Namespace,
+				Annotations: map[string]string{
+					"openshift.io/host.generated": "true",
+				},
+			},
+			Spec: v1.RouteSpec{
+				To: v1.RouteTargetReference{
+					Kind: "Service",
+					Name: "member-operator-console-plugin",
```
rajivnathan, line 59: Should we wait for the `member-operator-console-plugin` service to exist before trying to create the route? Would it make the test more stable?

rajivnathan (commented):
Looks good! Just a couple questions

test/e2e/parallel/web_console_plugin_test.go
MatousJobanek: I have no clue why this check was introduced at all TBH 
https://github.com/codeready-toolchain/toolchain-e2e/pull/690#discussion_r1166832887
I would maybe just drop it entirely as it doesn't seem to have any impact on the rest of the test.

test/e2e/parallel/web_console_plugin_test.go
```diff
@@ -0,0 +1,122 @@
...
+		var healthCheckResponse *http.Response
+
+		req, err := http.NewRequest("GET", healthCheckURL, nil)
+		require.NoError(t, err)
+		req.Header.Set("Authorization", signupRequest.GetToken())
+
+		healthCheckResponse, err = httpClient.Do(req) //nolint
+		require.NoError(t, err)
+		defer healthCheckResponse.Body.Close()
+		require.Equal(t, http.StatusOK, healthCheckResponse.StatusCode, "error polling health check endpoint", route)
```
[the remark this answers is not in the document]
MatousJobanek: good catch, fixed in ee0aa34e
thanks

test/e2e/parallel/web_console_plugin_test.go
MatousJobanek: anyway, I removed it seems to be unnecessary and just extra noise in the test d25ccfe1
cc @sbryzak

test/e2e/parallel/web_console_plugin_test.go
MatousJobanek: and the paired PR: https://github.com/codeready-toolchain/member-operator/pull/431

PR #719, opened 2023-05-01
Test invalid proxy request headers

rajivnathan:
tests for https://github.com/codeready-toolchain/registration-service/pull/317

### DIFF
--- a/test/e2e/proxy_test.go
+++ b/test/e2e/proxy_test.go
@@ -326,9 +326,9 @@ func TestProxyFlow(t *testing.T) {
 						defer func() {
 							resp.Body.Close()
 						}()
-						require.Equal(t, 403, resp.StatusCode)
+						require.Equal(t, 403, resp.StatusCode) // should be forbidden
 						r, _ := io.ReadAll(resp.Body)
-						assert.NotContains(t, string(r), "NodeList")
+						assert.NotContains(t, string(r), "NodeList") // request should be forbidden so there should be no NodeList in the response
 					})
 				}
 
--- a/test/e2e/proxy_test.go
+++ b/test/e2e/proxy_test.go
@@ -289,6 +289,50 @@ func TestProxyFlow(t *testing.T) {
 				_, err := hostAwaitWithShorterTimeout.CreateAPIProxyClient(t, user.token, proxyWorkspaceURL)
 				require.EqualError(t, err, `an error on the server ("unable to get target cluster: the requested space is not available") has prevented the request from succeeding`)
 			})
+
+			t.Run("invalid request headers", func(t *testing.T) {
+				// given
+				proxyWorkspaceURL := hostAwait.ProxyURLWithWorkspaceContext(user.compliantUsername)
+				rejectedHeaders := []headerKeyValue{
+					{"Impersonate-Group", "system:cluster-admins"},
+					{"Impersonate-Group", "system:node-admins"},
+				}
+				client := http.Client{
+					Timeout: time.Duration(5 * time.Second), // because sometimes the network connection may be a bit slow
+				}
+				client.Transport = &http.Transport{
+					TLSClientConfig: &tls.Config{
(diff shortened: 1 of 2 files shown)


alexeykazakov (approved):
Nice!

test/e2e/proxy_test.go
```diff
@@ -289,6 +289,50 @@ func TestProxyFlow(t *testing.T) {
 				_, err := hostAwaitWithShorterTimeout.CreateAPIProxyClient(t, user.token, proxyWorkspaceURL)
 				require.EqualError(t, err, `an error on the server ("unable to get target cluster: the requested space is not available") has prevented the request from succeeding`)
 			})
+
+			t.Run("invalid request headers", func(t *testing.T) {
+				// given
+				proxyWorkspaceURL := hostAwait.ProxyURLWithWorkspaceContext(user.compliantUsername)
+				rejectedHeaders := []headerKeyValue{
```
mfrancisc, line 296: Does it make sense to check also the other headers:

- Impersonate-User
- Impersonate-Extra-( extra name )
- Impersonate-Uid

and mixing them , as we did with the unit tests and as documented here https://kubernetes.io/docs/reference/access-authn-authz/authentication/#user-impersonation ?

test/e2e/proxy_test.go
```diff
@@ -289,6 +289,50 @@ func TestProxyFlow(t *testing.T) {
...
+
+						// when
+						resp, err := client.Do(request)
+
+						// then
+						require.NoError(t, err)
+						require.NotNil(t, resp)
+						require.Equal(t, 403, resp.StatusCode)
+						r, _ := io.ReadAll(resp.Body)
+						defer func() {
```
mfrancisc, line 328: is it actually required to defer or we could just close the resp as last thing before ending the current loop?

test/e2e/proxy_test.go
```diff
@@ -289,6 +289,50 @@ func TestProxyFlow(t *testing.T) {
...
+							resp.Body.Close()
+						}()
+						assert.NotContains(t, string(r), "NodeList")
```
mfrancisc, line 331: maybe we could add a comment here explaining why this should not contain that `NodeList` or we could check for an empty response body ? I don't actually recall what is returned as body when the `StatusCode` is `403` ...

mfrancisc (commented):
Looks nice! 👍 

I've added few comments

test/e2e/proxy_test.go
rajivnathan: I tested other headers without the changes in https://github.com/codeready-toolchain/registration-service/pull/317 but couldn't get it to do what I was expecting so I figured it wasn't worth testing in e2e tests. They're covered in unit tests so that's good enough I think.

test/e2e/proxy_test.go
rajivnathan: It's generally a good practice to use defer to ensure the body is closed in case it is not read fully by the end of the loop. But I just noticed that I have it in the wrong place after moving the code around. It should be done earlier actually for it to be effective. Done in 2d7bbc343e3339dee1ea60d0669a7b58666c4abe

test/e2e/proxy_test.go
rajivnathan: The response is 
```
{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"nodes is forbidden: User \"proxymember1\" cannot list resource \"nodes\" in API group \"\" at the cluster scope","reason":"Forbidden","details":{"kind":"nodes"},"code":403}
```
Checking status code is 403 is probably good enough but I changed the assertion to check for the specific error message just to be more thorough. See 6128d0f356b76102a47fa2e9034a1fa269af05a2

mfrancisc: Thanks, this makes it more clear, at least for me!

mfrancisc (approved):
Looks Good! 

Thanks for checking my comments.

PR #494, opened 2022-03-07
Add test for gitopsdeployments to appstudio-user-actions role

jgwest:
This PR:
- Adds a test for the gitopsdeployment resource of the appstudio-user-action

See https://github.com/codeready-toolchain/host-operator/pull/605

PR #721, opened 2023-05-04
Added Workspace Name Label check

bamachrn:
This PR is for checking `appstudio.redhat.com/workspace_name ` label in the namespace

Paired with https://github.com/codeready-toolchain/host-operator/pull/798

### DIFF
--- a/testsupport/space.go
+++ b/testsupport/space.go
@@ -187,6 +187,7 @@ func verifyResourcesProvisionedForSpace(t *testing.T, hostAwait *wait.HostAwaiti
 	space, err := hostAwait.WaitForSpace(t, spaceName,
 		wait.UntilSpaceHasTier(tier.Name),
 		wait.UntilSpaceHasLabelWithValue(fmt.Sprintf("toolchain.dev.openshift.com/%s-tier-hash", tier.Name), hash),
+		wait.UntilSpaceHasLabelWithValue("appstudio.redhat.com/workspace_name", tier.Name),
 		wait.UntilSpaceHasConditions(Provisioned()),
 		wait.UntilSpaceHasStateLabel(toolchainv1alpha1.SpaceStateLabelValueClusterAssigned),
 		wait.UntilSpaceHasStatusTargetCluster(targetCluster.ClusterName))
--- a/testsupport/space.go
+++ b/testsupport/space.go
@@ -219,9 +219,7 @@ func verifyResourcesProvisionedForSpace(t *testing.T, hostAwait *wait.HostAwaiti
 		ns, err := targetCluster.WaitForNamespace(t, space.Name, tier.Spec.Namespaces[0].TemplateRef, space.Spec.TierName, wait.UntilNamespaceIsActive())
 		require.NoError(t, err)
 		require.Contains(t, ns.Labels, toolchainv1alpha1.WorkspaceLabelKey)
-		require.Contains(t, ns.Labels, "appstudio.redhat.com/workspace_name")
 		assert.Equal(t, space.Name, ns.Labels[toolchainv1alpha1.WorkspaceLabelKey])
-		assert.Equal(t, space.Name, ns.Labels["appstudio.redhat.com/workspace_name"])
 	}
 	// Wait for space to have list of provisioned namespaces in Space status.
 	// the expected namespaces for `nsTmplSet.Status.ProvisionedNamespaces` are checked as part of VerifyNSTemplateSet function above.
--- a/testsupport/space.go
+++ b/testsupport/space.go
@@ -187,7 +187,6 @@ func verifyResourcesProvisionedForSpace(t *testing.T, hostAwait *wait.HostAwaiti
 	space, err := hostAwait.WaitForSpace(t, spaceName,
 		wait.UntilSpaceHasTier(tier.Name),
 		wait.UntilSpaceHasLabelWithValue(fmt.Sprintf("toolchain.dev.openshift.com/%s-tier-hash", tier.Name), hash),
-		wait.UntilSpaceHasLabelWithValue("appstudio.redhat.com/workspace_name", tier.Name),
 		wait.UntilSpaceHasConditions(Provisioned()),
 		wait.UntilSpaceHasStateLabel(toolchainv1alpha1.SpaceStateLabelValueClusterAssigned),
 		wait.UntilSpaceHasStatusTargetCluster(targetCluster.ClusterName))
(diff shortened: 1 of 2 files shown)


testsupport/space.go
```diff
@@ -187,6 +187,7 @@ func verifyResourcesProvisionedForSpace(t *testing.T, hostAwait *wait.HostAwaiti
 	space, err := hostAwait.WaitForSpace(t, spaceName,
 		wait.UntilSpaceHasTier(tier.Name),
 		wait.UntilSpaceHasLabelWithValue(fmt.Sprintf("toolchain.dev.openshift.com/%s-tier-hash", tier.Name), hash),
+		wait.UntilSpaceHasLabelWithValue("appstudio.redhat.com/workspace_name", tier.Name),
```
mfrancisc, line 190: I guess you should check this label on the namespace resources not the `Space`. 

If you have to check this label you need to add a check in the block of code starting from line 217:

```
	if tier.Name == "appstudio" {
		// checks that namespace exists and has the expected label(s)
		ns, err := targetCluster.WaitForNamespace(t, space.Name, tier.Spec.Namespaces[0].TemplateRef, space.Spec.TierName, wait.UntilNamespaceIsActive())
		require.NoError(t, err)
		require.Contains(t, ns.Labels, toolchainv1alpha1.WorkspaceLabelKey)
                 require.Contains(t, ns.Labels, "appstudio.redhat.com/workspace_name")
		assert.Equal(t, space.Name, ns.Labels[toolchainv1alpha1.WorkspaceLabelKey])
	}
``` 

plus maybe add a similar check for the `appstudio-env` tier if required. 

Also the value of your label is not `tier.Name` but `space.Name` I guess.

bamachrn: updated the `appstudio` tier check. I could not find anything on `appstudio-env` tier check in space.go. So added checks on the checks.go. should I add a check in the space.go as well for `appstudio-env` tier? how do you think?

testsupport/space.go
```diff
@@ -219,7 +219,9 @@ func verifyResourcesProvisionedForSpace(t *testing.T, hostAwait *wait.HostAwaiti
 		ns, err := targetCluster.WaitForNamespace(t, space.Name, tier.Spec.Namespaces[0].TemplateRef, space.Spec.TierName, wait.UntilNamespaceIsActive())
 		require.NoError(t, err)
 		require.Contains(t, ns.Labels, toolchainv1alpha1.WorkspaceLabelKey)
+		require.Contains(t, ns.Labels, "appstudio.redhat.com/workspace_name")
 		assert.Equal(t, space.Name, ns.Labels[toolchainv1alpha1.WorkspaceLabelKey])
+		assert.Equal(t, space.Name, ns.Labels["appstudio.redhat.com/workspace_name"])
```
MatousJobanek, line 224: I'm not sure if we really need this part here. We already verify the provisioned labels as part of the checks.go, which should be enough

testsupport/tiers/checks.go
```diff
@@ -1395,6 +1397,18 @@ func gitOpsServiceLabel() namespaceObjectsCheck {
 	}
 }
 
+func appstudioWorkSpaceNameLabel() namespaceObjectsCheck {
+	return func(t *testing.T, ns *corev1.Namespace, memberAwait *wait.MemberAwaitility, _ string) {
+		// TODO fix for migration/existing namespaces cases
+		labelWaitCriterion := []wait.LabelWaitCriterion{}
+		if !strings.HasPrefix(ns.Name, "migration-") {
+			labelWaitCriterion = append(labelWaitCriterion, wait.UntilObjectHasLabel("appstudio.redhat.com/workspace_name", ns.Name))
+		}
```
MatousJobanek, line 1406: @rajivnathan correct me if I'm wrong, we don't need this check anymore - we take care of updating the label for the already existing namespaces, right? 
I know that you introduced this check for the other label https://github.com/codeready-toolchain/toolchain-e2e/pull/494/commits/e71b580164777834e1086ace135b1dab4c4c6a8d but it was before we fixed that in the NSTemplateSet controller https://github.com/codeready-toolchain/member-operator/pull/372 so the already existing namespaces should be updated now.

```suggestion
		labelWaitCriterion = append(labelWaitCriterion, wait.UntilObjectHasLabel("appstudio.redhat.com/workspace_name", ns.Name))
```

testsupport/space.go
bamachrn: yes this is a repeat check. let me remove it.

testsupport/tiers/checks.go
bamachrn: ack

testsupport/space.go
mfrancisc: @MatousJobanek I'm wondering if we still need that if statement for the `appstudio` tier ? 
We could probably check the `toolchainv1alpha1.WorkspaceLabelKey` as part of the `func (a *appstudioTierChecks) GetNamespaceObjectChecks` and drop this specific case ?

testsupport/tiers/checks.go
```diff
@@ -1395,6 +1397,17 @@ func gitOpsServiceLabel() namespaceObjectsCheck {
 	}
 }
 
+func appstudioWorkSpaceNameLabel() namespaceObjectsCheck {
+	return func(t *testing.T, ns *corev1.Namespace, memberAwait *wait.MemberAwaitility, _ string) {
+		// TODO fix for migration/existing namespaces cases
```
mfrancisc, line 1402:
```suggestion
```

bamachrn: thanks

mfrancisc (approved):

testsupport/tiers/checks.go
mfrancisc: Actually seems that the `TestAfterMigration` is failing because the namespace `migration-appstudio-provisioned-tenant` doesn't have that expected label. 
So probably that if statement is still needed.

bamachrn: yes, added them back. let's see.

testsupport/space.go
MatousJobanek: no, we don't need that - let's drop it as part of your cleanup story :+1:

testsupport/tiers/checks.go
MatousJobanek: if it failed, then it means that there is a bug in our logic. The already existing namespaces should be updated with the new label.

MatousJobanek: Ok, I found what is wrong. The namespace was updated, but you expect a wrong value.
```
Actual labels:map[appstudio.redhat.com/workspace_name:migration-appstudio-provisioned 
```
vs:
```
appstudio.redhat.com/workspace_name:migration-appstudio-provisioned-tenant
```
@bamachrn you need to fix it in your tests - the label contains the name of the workspace, not of the namespace. And please drop the if-statement again.

testsupport/tiers/checks.go
```diff
@@ -1395,6 +1397,18 @@ func gitOpsServiceLabel() namespaceObjectsCheck {
 	}
 }
 
+func appstudioWorkSpaceNameLabel() namespaceObjectsCheck {
+	return func(t *testing.T, ns *corev1.Namespace, memberAwait *wait.MemberAwaitility, owner string) {
+		// TODO fix for migration/existing namespaces cases
+		labelWaitCriterion := []wait.LabelWaitCriterion{}
+		if !strings.HasPrefix(ns.Name, "migration-") {
+			labelWaitCriterion = append(labelWaitCriterion, wait.UntilObjectHasLabel("appstudio.redhat.com/workspace_name", owner))
+		}
```
MatousJobanek, line 1406: please, drop this if-statement
```suggestion
		labelWaitCriterion = append(labelWaitCriterion, wait.UntilObjectHasLabel("appstudio.redhat.com/workspace_name", owner))
```

MatousJobanek:
/retest
infra

PR #722, opened 2023-05-05
fix: drop workspace label check from space testsupport

mfrancisc:
- drop redundant check on the namespace since `VerifyNSTemplateSet` already executes `memberAwait.WaitForNamespace` , plus this check was assuming only 1 namespace (while the `VerifyNSTemplateSet` loops over the expected namespaces )
- remove workspace label check since the label will be dropped 

This is part of Jira: https://issues.redhat.com/browse/CRT-1776


mfrancisc:
/retest

TestSpaceLister failed, trying to reproduce locally

alexeykazakov (approved):

PR #724, opened 2023-05-08
make migration test usernames shorter

mfrancisc:
This change is needed because now usernames length is <= 20 chars, causing some of the migration test checks to fail.

### DIFF
--- a/test/migration/setup_runner.go
+++ b/test/migration/setup_runner.go
@@ -16,14 +16,14 @@ import (
 )
 
 const (
-	ProvisionedUser             = "migration-provisioned"
-	DeactivatedUser             = "migration-deactivated"
-	BannedUser                  = "migration-banned-provisioned"
-	AppStudioProvisionedUser    = "migration-appstudio-provisioned"
-	SecondMemberProvisionedUser = "migration-second-member-provisioned-user"
+	ProvisionedUser             = "mgr-provisioned"
+	DeactivatedUser             = "mgr-deactivated"
+	BannedUser                  = "mgr-banned"
+	AppStudioProvisionedUser    = "mgr-appstudio"
+	SecondMemberProvisionedUser = "mgr-member2-user"
 
-	ProvisionedAppStudioSpace    = "migration-appstudio-provisioned-space"
-	SecondMemberProvisionedSpace = "migration-second-member-provisioned-space"
+	ProvisionedAppStudioSpace    = "mgr-appstudio-space"
+	SecondMemberProvisionedSpace = "mgr-member2-space"
 )
 
 type SetupMigrationRunner struct {
--- a/test/migration/setup_runner.go
+++ b/test/migration/setup_runner.go
@@ -17,7 +17,7 @@ import (
 
 const (
 	ProvisionedUser             = "mig-prov"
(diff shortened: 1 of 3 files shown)


testsupport/space.go
```diff
@@ -231,7 +231,7 @@ func verifyResourcesProvisionedForSpace(t *testing.T, hostAwait *wait.HostAwaiti
 }
 
 func CreateMurWithAdminSpaceBindingForSpace(t *testing.T, awaitilities wait.Awaitilities, space *toolchainv1alpha1.Space, cleanup bool) (*toolchainv1alpha1.UserSignup, *toolchainv1alpha1.MasterUserRecord, *toolchainv1alpha1.SpaceBinding) {
-	username := "for-space-" + space.Name
+	username := space.Name
```
MatousJobanek, line 234: with this function, we wanted to provision a MUR to have the SpaceBinding set so the Space doesn't get deleted. this means that the User would have access to two spaces:
1. the one that was created as a standalone Space
2. the one that is provisioned for the user by default

That was the reason for introducing the prefix `for-space-` - to differentiate the names of the Spaces. 
However, since the name would be the same, then the UserSignup controller woudn't provision a separate Space CR, but would reuse the already existing one. 

That being said - could we change the prefix to:
```go
username := "sp-" + space.Name
```

test/migration/setup_runner.go
```diff
@@ -16,14 +16,14 @@ import (
...
-	ProvisionedAppStudioSpace    = "migration-appstudio-provisioned-space"
-	SecondMemberProvisionedSpace = "migration-second-member-provisioned-space"
+	ProvisionedUser             = "mgr-provisioned"
+	DeactivatedUser             = "mgr-deactivated"
+	BannedUser                  = "mgr-banned"
+	AppStudioProvisionedUser    = "mgr-appstudio"
+	SecondMemberProvisionedUser = "mgr-member2-user"
+
+	ProvisionedAppStudioSpace    = "mgr-appstudio-space"
+	SecondMemberProvisionedSpace = "mgr-member2-space"
```
MatousJobanek, line 26: mgr is usually used as an abbreviation for different words. 
how about using `mig-` prefix? And we could also shorten the other parts of the name so we can afford adding the `sp-` prefix.
```suggestion
	ProvisionedUser             = "mig-prov"
	DeactivatedUser             = "mig-deact
	BannedUser                  = "mig-banned"
	AppStudioProvisionedUser    = "mig-appst"
	SecondMemberProvisionedUser = "mig-m2-user"

	ProvisionedAppStudioSpace    = "mig-appst-space"
	SecondMemberProvisionedSpace = "mig-m2-space"
```

testsupport/space.go
mfrancisc: But unless I'm missing something, the function triggers a signups request with `NoSpace()` few lines above, it doesn't seem to provision the second space.But it's  also Monday , so probably I just need more coffe ☕ 😄

MatousJobanek: ah, OK. I didn't see the line. Yeah, you are right then :+1:

MatousJobanek (approved):

PR #727, opened 2023-05-11
Test user_signups_approved_with_method_total metric

rajivnathan:
Paired with https://github.com/codeready-toolchain/host-operator/pull/804

### DIFF
--- a/test/metrics/metrics_test.go
+++ b/test/metrics/metrics_test.go
@@ -19,75 +19,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-// TestMetricsWhenUsersAutomaticallyApproved verifies that `UserSignupsApprovedMetric` and `UserSignupsApprovedWithMethodMetric` counters are increased when users are approved
-func TestMetricsWhenUsersAutomaticallyApproved(t *testing.T) {
-	// given
-	awaitilities := WaitForDeployments(t)
-	hostAwait := awaitilities.Host()
-	memberAwait := awaitilities.Member1()
-	memberAwait2 := awaitilities.Member2()
-	hostAwait.UpdateToolchainConfig(t, testconfig.AutomaticApproval().Enabled(true))
-	// host metrics should be available at this point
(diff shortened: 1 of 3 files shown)


test/metrics/metrics_test.go
```diff
@@ -19,6 +19,72 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+// TestMetricsWhenUsersAutomaticallyApproved verifies that `UserSignupsApprovedMetric` and `UserSignupsApprovedWithMethodMetric` counters are increased when users are approved
+func TestMetricsWhenUsersAutomaticallyApproved(t *testing.T) {
```
alexeykazakov, line 23: Shouldn't we also rename `TestMetricsWhenUsersDeactivated()` to `TestMetricsWhenUsersManuallyApproved()`? Those two functions look identical and test both activation and deactivation but the names are a bit confusing...

alexeykazakov (approved):

test/metrics/metrics_test.go
rajivnathan: Ok, I renamed them to `TestMetricsWhenUsersManuallyApprovedAndThenDeactivated` and `TestMetricsWhenUsersAutomaticallyApprovedAndThenDeactivated`

mfrancisc (approved):
Looks fine 👍 

I don't think it's super important but was wondering - should we test `Ban` and `Disable` with a mix of `manual` and `automatic` approved usersignups ? Now I see the `TestMetricsWhenUsersBanned` and `TestMetricsWhenUserDisabled` are being provisioned "only" with manual approved users.

rajivnathan:
>I don't think it's super important but was wondering - should we test Ban and Disable with a mix of manual and automatic approved usersignups ? Now I see the TestMetricsWhenUsersBanned and TestMetricsWhenUserDisabled are being provisioned "only" with manual approved users.

TestMetricsWhenUsersBanned and TestMetricsWhenUserDisabled aren't impacted by the approval type so it's not really necessary.

PR #749, opened 2023-06-21
Perf test update

rajivnathan:
Includes the following changes:
- Disable copy csv feature before test
- In addition to outputting the results to the terminal it now also outputs results to a .csv file to make it easier to copy-paste the data into the spreadsheet for comparison
- Added a `testname` flag to make the .csv file names more purposeful eg. `--testname="8xlarge-baseline"` results in a file named `<timestamp>-8xlarge-baseline.csv`
- Fixed long-standing issue where progress bars duplicated lines and polluted the command output
- Longer timeout (15 mins instead of the default 5 mins) specifically for RHODS operator install because it often takes longer


alexeykazakov (approved):
Looks good! Do you want to sync the operators we install here with what we have in stage/prod in a separate PR?

rajivnathan:
@alexeykazakov removed the operator stuff, will create in a separate PR

PR #756, opened 2023-06-28
Operator Onboarding Setup Tool Improvements

rajivnathan:
Running the perf setup tool that's part of the onboarding process took a long time and required some workarounds like running two instances of the tool to reduce the overall running time. I recently discovered that one of the things causing slowness was client-side throttling. Basically, the template that is applied to create resources for each user leads to the client creating many resources very quickly and causes it to hit the client-side throttling problem:
```
I0620 14:47:11.146533   25957 request.go:601] Waited for 1.097001422s due to client-side throttling, not priority and fairness, request: GET:https://api.<redacted>:6443/apis/rbac.authorization.k8s.io/v1/namespaces/pancake-0002-dev/rolebindings/build-robot-edit-11
```
The 11th rolebinding always triggered the client-side throttling because the default is to throttle requests if 10 requests are made in quick succession but there are client parameters that can be used to increase the defaults and eliminate client-side throttling.

With the improvements from this PR the required time to load up the cluster with 2000 users was reduced from ~2hr to ~45 mins and removes the need for such workarounds. As a result the instructions are simpler too.

Summary of changes:
- Adjusts QPS and Burst configuration on clients to eliminate client-side throttling
- Each task is split into concurrent goroutines
- Each concurrent routine has its own client
- Removed the user batches parameter since the old batching logic is no longer necessary
- Change the wait logic to wait for Space to be completely provisioned instead of only waiting for Namespace to exist, there weren't issues waiting for Namespace but this is generally a more robust way
- Updates to the README with simplified instructions


alexeykazakov (approved):
Great improvment!

setup/configuration/configuration.go
```diff
@@ -96,6 +94,11 @@ func NewClient(term terminal.Terminal, kubeconfigPath string) (client.Client, *r
 		term.Fatalf(err, "cannot create client config")
 	}
 
+	// Set QPS and Burst to higher values to avoid client-side throttling issues
+	// prometheus uses these QPS and Burst values so it shouldn't be an issue, see https://github.com/prometheus-operator/prometheus-operator/blob/9d68ecf289d711c66bef39d2f83429265abc6986/pkg/k8sutil/k8sutil.go#L96-L97
+	clientConfig.QPS = 100
+	clientConfig.Burst = 100
```
MatousJobanek, line 100: Good catch.
Would be worth changing that in e2e test init part as well - I don't think that we set anything explicitly there.
Also, for clients created for ToolchainClusters CRs we set 20/30 https://github.com/codeready-toolchain/toolchain-common/blob/master/pkg/cluster/service.go#L32-L33 but I'm not sure if it can have a negative impact on anything we use the clients for :thinking:

MatousJobanek (approved):
Amazing improvement :+1: :1st_place_medal: :rocket:

mfrancisc (approved):
Great improvement! 🚀 

I don't remember if we actually save those results somewhere , like in a shared google drive folder or some artifact repository. If not maybe we can consider doing that, so we could track performance improvements in time.

setup/configuration/configuration.go
rajivnathan: Thanks! Good point about e2e tests. Do you know if we see the client-side throttling messages there? I can look into it if you're not sure.

>but I'm not sure if it can have a negative impact on anything we use the clients for

I'm not sure but I doubt there are concerns on the client-side except I assume that if the server receives too many requests then it could hit server-side throttling. I think we need to be more careful in production code to avoid too many clients with high QPS/Burst so that we reduce the likelihood of bombarding the api server with too many requests too quickly and cause performance issues. But it doesn't seem like we're doing that as a regular practice so I guess we should be fine. It's likely a lesser concern in our tests since the only api server requests are coming solely from the tests and operators whereas on production there are users & workloads as well that are far more unpredictable.

rajivnathan:
@mfrancisc Thanks!

>I don't remember if we actually save those results somewhere , like in a shared google drive folder or some artifact repository. If not maybe we can consider doing that, so we could track performance improvements in time.

yes, we are currently using spreadsheets to track these metrics but we'd like to store the runs in a DB someday and I have a few other wish list things but it comes down to priorities 🙂

PR #764, opened 2023-07-20
fix: restart deployments after creating resources

mfrancisc:
see https://redhat-internal.slack.com/archives/CHK0J6HT6/p1689863736799519

### DIFF
--- a/make/test.mk
+++ b/make/test.mk
@@ -347,7 +347,7 @@ create-host-resources:
 		echo "{\"spec\":{\"members\":{\"specificPerMemberCluster\":{\"$${TOOLCHAIN_CLUSTER_NAME}\":{\"webhook\":{\"deploy\":false},\"webConsolePlugin\":{\"deploy\":true},\"environment\":\"${ENVIRONMENT}\"}}}}}" > $$PATCH_FILE; \
 		oc patch toolchainconfig config -n ${HOST_NS} --type=merge --patch "$$(cat $$PATCH_FILE)"; \
 	fi;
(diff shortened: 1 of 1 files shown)


make/test.mk
```diff
@@ -346,6 +346,8 @@ create-host-resources:
 		echo "{\"spec\":{\"members\":{\"specificPerMemberCluster\":{\"$${TOOLCHAIN_CLUSTER_NAME}\":{\"webhook\":{\"deploy\":false},\"webConsolePlugin\":{\"deploy\":true}}}}}}" > $$PATCH_FILE; \
 		oc patch toolchainconfig config -n ${HOST_NS} --type=merge --patch "$$(cat $$PATCH_FILE)"; \
 	fi;
+	echo "Restart host operator pods so it can get the ToolchainConfig CRs while it's starting up".
```
rajivnathan, line 349: The output should be updated to indicate we're restarting the host operator so that the SpaceRequest controller configuration in ToolchainConfig can be picked up by host operator main.go and SpaceRequest controller will be started.

mfrancisc: I left it a bit more generic , because the problem was with SpaceRequest specifically but it can happen with any other configuration loaded during startup only.

rajivnathan: 👍  Just a small suggestion:
```suggestion
	echo "Restart host operator pods so it can get the updated ToolchainConfig CRs while it's starting up"
```

MatousJobanek (approved):

PR #765, opened 2023-07-26
ASC-352 Add proxy tests for components

jkopriva:


rajivnathan:
Closing this since it was discussed with the team and we do not want to add dependencies on any AppStudio CRDs which need to stay up to date.

We will stick with only the 2 existing CRDs that have a special purpose:
- Application was introduced to test different RBAC set in our role templates
- Environment was introduced to make sure that the appstudio tier can be deployed in openshift-ci clusters as part of the e2e tests

We'll continue to work with standard k8s resources for the proxy tests otherwise - the proxy doesn't care if it's a Pod, Deployment or a appstudio component - it's just resources

Feel free to reopen if you plan to make further changes. Thanks!

PR #767, opened 2023-08-03
verify that SA secrets refs are not deleted

MatousJobanek:
paired PR: https://github.com/codeready-toolchain/member-operator/pull/469

### DIFF
--- a/test/e2e/parallel/serviceaccount_test.go
+++ b/test/e2e/parallel/serviceaccount_test.go
@@ -20,7 +20,6 @@ func TestDoNotOverrideServiceAccount(t *testing.T) {
 	// given
 	// Skipping the TestDoNotOverrideServiceAccount test instead of deleting it because we will need to create SAs as part
 	// of the environment sub-workspaces so the test & logic will be useful to keep.
-	t.Skip("skipping since the user SA was removed as part of https://github.com/codeready-toolchain/host-operator/pull/719. To be added back with https://issues.redhat.com/browse/ASC-249")
 	t.Parallel()
 	awaitilities := WaitForDeployments(t)
 	member := awaitilities.Member1()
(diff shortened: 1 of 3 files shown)


test/e2e/parallel/serviceaccount_test.go
```diff
@@ -20,7 +20,6 @@ func TestDoNotOverrideServiceAccount(t *testing.T) {
 	// given
 	// Skipping the TestDoNotOverrideServiceAccount test instead of deleting it because we will need to create SAs as part
 	// of the environment sub-workspaces so the test & logic will be useful to keep.
```
mfrancisc, line 22: we could drop those now
```suggestion
```

test/e2e/parallel/serviceaccount_test.go
```diff
@@ -36,35 +35,64 @@ func TestDoNotOverrideServiceAccount(t *testing.T) {
...
+	secrets:
+		for j := 0; j < i; j++ {
+			expName := fmt.Sprintf("dummy-secret-%d", j)
+			for _, secretRef := range sa.Secrets {
+				if secretRef.Name == fmt.Sprintf("dummy-secret-%d", j) {
+					continue secrets
+				}
+			}
+			assert.Fail(t, fmt.Sprintf("secret '%s' not found", expName))
+		}
```
mfrancisc, line 80: just trying to understand - when `i` is 0 this check will be skipped (ok it will be checked once i>0).
But why do we need to start from `secrets` again every time we find a secret?

Could we check that all the expected number of secrets (j) are there , and fail the the first one that is missing ?  

something like:
```
	for j := 0; j <= i; j++ {
		expName := fmt.Sprintf("dummy-secret-%d", j)
		found := false
		for _, secretRef := range sa.Secrets {
			if secretRef.Name == fmt.Sprintf("dummy-secret-%d", j) {
			        found = true
				break
			}
		}
		if !found {
			assert.Fail(t, fmt.Sprintf("secret '%s' not found", expName))
		}
	}
``` 
and remove the `secrets` tag and continue statement ?

mfrancisc (commented):
Nice job 👏 

I've added few comments only.

test/e2e/parallel/serviceaccount_test.go
MatousJobanek: > just trying to understand - when i is 0 this check will be skipped (ok it will be checked once i>0).

right - bad copy-paste from the `for i` line :-) 

as for your version - isn't it the same?
the only benefit would be that I could combine secrets & pullSecrets together

mfrancisc: If I'm not mistaken is not the same because:
1. you can combine them right
2. you are not starting from j=0 every time you find a secret, but you validate them in one single loop, because the current `continue secrets` exists the outer loop as well right ?

test/e2e/parallel/serviceaccount_test.go
MatousJobanek: removed in https://github.com/codeready-toolchain/toolchain-e2e/pull/768

PR #768, opened 2023-08-03
 address comments from PR #767 

MatousJobanek:
this PR addresses comments from https://github.com/codeready-toolchain/toolchain-e2e/pull/767
the only commit to be reviewed in this PR is: 5edc3160

### DIFF
--- a/test/e2e/parallel/serviceaccount_test.go
+++ b/test/e2e/parallel/serviceaccount_test.go
@@ -56,9 +56,12 @@ func TestDoNotOverrideServiceAccount(t *testing.T) {
 		})
 		require.NoError(t, err)
 
-		// drop the SpaceRoles annotation from the namespace to trigger the reconciliation
-		_, err = member.UpdateNSTemplateSet(t, mur.Name, func(nsTmplSet *v1alpha1.NSTemplateSet) {
-			delete(nsTmplSet.Annotations, v1alpha1.LastAppliedSpaceRolesAnnotationKey)
+		// Update the namespace annotations & labels to trigger the reconciliation
+		_, err = member.UpdateNamespace(t, mur.Name, func(ns *corev1.Namespace) {
+			// drop the last-applied-space-roles annotation, so we are sure that the content of the roles are re-applied
+			delete(ns.Annotations, v1alpha1.LastAppliedSpaceRolesAnnotationKey)
+			// change the tier name, so we are sure that the content of the tier template is re-applied
+			ns.Labels[v1alpha1.TierLabelKey] = "appstudio"
 		})
 		require.NoError(t, err)
 
--- a/test/e2e/parallel/serviceaccount_test.go
+++ b/test/e2e/parallel/serviceaccount_test.go
@@ -18,8 +18,6 @@ import (
 
 func TestDoNotOverrideServiceAccount(t *testing.T) {
 	// given
-	// Skipping the TestDoNotOverrideServiceAccount test instead of deleting it because we will need to create SAs as part
-	// of the environment sub-workspaces so the test & logic will be useful to keep.
 	t.Parallel()
 	awaitilities := WaitForDeployments(t)
 	member := awaitilities.Member1()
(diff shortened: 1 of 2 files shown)


mfrancisc (approved):
Great job 🏅 

Thanks for looking into my comments.

mfrancisc:
@MatousJobanek I've noticed only now that there are a couple of issues with the test, if I'm not mistaken:

1. `expectedSecrets` is always nil because we need to pass `fmt.Sprintf("%s-env", mur.Name)` as namespace to `getSASecrets` and not just `mur.Name`
2. the NSTemplateSet for some reason doesn't have annotations and labels apparently, so I'm not totally sure the reconcile loop is being triggered 
```
✗ k get nstemplateset -n toolchain-member-03171406 -o yaml
apiVersion: v1
items:
  kind: NSTemplateSet
  metadata:
    creationTimestamp: "2023-08-03T21:51:04Z"
    finalizers:
    - finalizer.toolchain.dev.openshift.com
    generation: 2
    name: do-not-override-sa
    namespace: toolchain-member-03171406
    resourceVersion: "361459"
    uid: 9a63554f-fbeb-4257-8148-d897b2e582ce
  spec:
    clusterResources:
      templateRef: appstudio-env-clusterresources-e0e1f34-e0e1f34
    namespaces:
    - templateRef: appstudio-env-env-88b275d-88b275d
    spaceRoles:
    - templateRef: appstudio-env-admin-ba5db27-ba5db27
      usernames:
      - do-not-override-sa
    tierName: appstudio-env
  status:
    conditions:
    - lastTransitionTime: "2023-08-03T21:51:27Z"
      reason: Provisioned
      status: "True"
      type: Ready
    provisionedNamespaces:
    - name: do-not-override-sa-env
      type: default
kind: List
metadata:
  resourceVersion: ""
```

3. MINOR thing , we could drop the `nolint:unused` comment in `getSASecrets`

mfrancisc:
also out of curiosity - why are we updating the `sa.Secrets` and `sa.ImagePullSecrets` references but we are not creating the actual secrets ? I guess there is some other controller/automation that does that in RHTAP ?

test/e2e/parallel/serviceaccount_test.go
```diff
@@ -57,47 +59,53 @@ func TestDoNotOverrideServiceAccount(t *testing.T) {
 		})
 		require.NoError(t, err)
 
-		// drop the SpaceRoles annotation from the namespace to trigger the reconciliation
-		_, err = member.UpdateNSTemplateSet(t, mur.Name, func(nsTmplSet *v1alpha1.NSTemplateSet) {
-			delete(nsTmplSet.Annotations, v1alpha1.LastAppliedSpaceRolesAnnotationKey)
+		// Update the namespace annotations & labels to trigger the reconciliation
+		_, err = member.UpdateNamespace(t, mur.Name, func(ns *corev1.Namespace) {
```
mfrancisc, line 63:
```suggestion
		_, err = member.UpdateNamespace(t, nsName, func(ns *corev1.Namespace) {
```

mfrancisc (commented):
Looks great! 👍 

just one missing occurrence!

test/e2e/parallel/serviceaccount_test.go
MatousJobanek: yeah, I know - already fixed that. Still doing & testing some changes locally  ;-)

MatousJobanek: just wait a sec - got stuck in a meeting :smile:

MatousJobanek:
> expectedSecrets is always nil because we need to pass fmt.Sprintf("%s-env", mur.Name) as namespace to getSASecrets and not just mur.Name

good catch. I left the original version that was there before, but didn't notice that the name of the namespace is different.
fixed in 701b39c

> the NSTemplateSet for some reason doesn't have annotations and labels apparently, so I'm not totally sure the reconcile loop is being triggered

yup... I noticed that yesterday as well, funny thing is that I changed the logic to update the namespace, but for some reason I reverted it back :man_facepalming: 
modified in 7015649 and fixed the namespace name again in 813c92fa
btw, I added another change in the SA - removal of the space label. Since we verify that the label is set as part of the tier checks then we will be 100% sure that the reconcile loop was triggered and the SA updated. 454e9c97

> MINOR thing , we could drop the nolint:unused comment in getSASecrets

good point, I missed that - removed in c5ac516

> also out of curiosity - why are we updating the sa.Secrets and sa.ImagePullSecrets references but we are not creating the actual secrets ? I guess there is some other controller/automation that does that in RHTAP ?

There are two types of secrets:
1. secrets that are created by OpenShift for every SA
2. secrets created by a controller (or a human) and connected to the SA

In the first case, the OpenShift controller checks if there is a secret linked in the SA, if not then it creates a new one and links it to the SA. When our controller removed the reference, then the OpenShift controller created a new secret - that's why we had a bunch of them in the namespaces and that's why we have also the verification that the number of secrets won't change - to verify if this OpenShift controller logic was triggered or not.

In the second case, the human or RHTAP controller wanted to link one particular secret. We technically don't care if it exists or not. The refs should just stay there no matter how the secret was linked. This is what I'm testing by adding additional refs.
I added additional comments in c7877e1 
I also modified the check of the secret refs so it verifies the original secret refs as well c5e33352

test/e2e/parallel/serviceaccount_test.go
mfrancisc: Sorry, I got email notifications that you pushed changes so I thought you were done :)

mfrancisc (approved):
Looks great! 👍 

Thanks for the additional improvements and for explaining the reasoning behind the secret refs part! Great job!

MatousJobanek:
/retest
infra

MatousJobanek:
/retest
infra

mfrancisc:
/retest

infra

PR #769, opened 2023-08-04
fix: SpaceRequest SA token generation doesn't keep in account of the member cluster

mfrancisc:
paired with: https://github.com/codeready-toolchain/host-operator/pull/841

It doesn't actually validate that the kubeconfig has access to different member cluster then the one where the spacerequest resource is, it just validates that we can provision subspace on different member cluster then the one where parentSpace and spacerequest resource is.


MatousJobanek (approved):
Nice :+1:

PR #774, opened 2023-08-10
verify replacement of * by verbs in appstudio templates

MatousJobanek:
https://github.com/redhat-appstudio/book/pull/123
paired with https://github.com/codeready-toolchain/host-operator/pull/847

### DIFF
--- a/testsupport/tiers/checks.go
+++ b/testsupport/tiers/checks.go
@@ -1471,17 +1471,17 @@ func appstudioUserActionsRole() spaceRoleObjectsCheck {
 				{
 					APIGroups: []string{"appstudio.redhat.com"},
 					Resources: []string{"applications", "components", "componentdetectionqueries"},
-					Verbs:     []string{"*"},
+					Verbs:     []string{"get", "list", "watch", "create", "update", "patch", "delete"},
 				},
 				{
 					APIGroups: []string{"appstudio.redhat.com"},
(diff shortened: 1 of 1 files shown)


mfrancisc (approved):
Thanks for fixing it 👍 

Just can't remember which verbs are included in `*` other then what you added? Couldn't find this explanation in k8s docs yet ...

MatousJobanek:
Literally any verb. 
It's not obvious, but you can specify any verb, technically anything. You can then execute SAR to check if the a user has the specific verb granted.
Imagine that you would create a role defining verb "somethingcrazy" and execute SAR to check if a user "francisc" has permissions "somethingcrazy" for some specific resource/apigroup.
As as example of something real - RBAC resources support verb "escalate" that let's you escalate permissions over those that were granted to you. Another real verb (for the RBAC) is also "bind".

TBH, I'm not aware of any verb apart from those that I listed in the PR (for these specific resources/apiGroups), but I guess that you got the point. * grants anything, which means also those verbs that don't belong to the most common ones, that's why it's better to list them.

mfrancisc:
@MatousJobanek thanks a lot for explaining that. I now remember the issue with "escalate" verb :) , and I agree that this seems like a best practice (avoiding `*`), even if in this specific case there might not be other verbs available for those resources/api groups. 

Thanks for your details and for the fix!

PR #790, opened 2023-08-31
Fix & mention local targets, and remove quay username prefix

MatousJobanek:
https://redhat-internal.slack.com/archives/CHK0J6HT6/p1693460828997409

### DIFF
--- a/dev_install.adoc
+++ b/dev_install.adoc
@@ -54,13 +54,23 @@ make appstudio-dev-deploy-latest
 If you want to install a local version of any of the Sandbox operators in dev mode then:
 
 . link:quay.adoc[Configure your quay account for dev deployment]
-. Run the following: +
-```
-make dev-deploy-e2e
+. Run any from the following commands:
+```bash
+# To deploy local versions of all repositories:
+make dev-deploy-e2e-local
+
+# To deploy local version only of the host-operator repo:
+make dev-deploy-e2e-host-local
+
+# To deploy local version only of the member-operator repo:
+make dev-deploy-e2e-member-local
+
+# To deploy local version only of the registration-service repo:
+make dev-deploy-e2e-registration-local
 ```
 
 === ToolchainStatus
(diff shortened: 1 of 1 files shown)


dev_install.adoc
```diff
@@ -54,13 +54,23 @@ make appstudio-dev-deploy-latest
 If you want to install a local version of any of the Sandbox operators in dev mode then:
 
 . link:quay.adoc[Configure your quay account for dev deployment]
-. Run the following: +
-```
-make dev-deploy-e2e
```
mfrancisc, line 59: should we document also this option ? 
```
# To deploy latest versions of all repositories:
make dev-deploy-e2e
```

mfrancisc (approved):

dev_install.adoc
MatousJobanek: it's technically the same as `dev-deploy-latest` if no `*_REPO_PATH` is provided. 
`dev-deploy-latest` just sets the requirement for the latest versions explicitly.

mfrancisc: Right , but I don't see that command documented here either. 

Anyway it's just minor thing!

MatousJobanek: there are many targets in our makefiles that are not documented here. Let's just consider `dev-deploy-e2e` as the middle-target that is called by all other targets that are documented here. 
With additional knowledge, you can use it on its own, but it's not recommended for those who are not familiar with the targets at all.

MatousJobanek: or you mean the `dev-deploy-latest` ? it's documented in the previous section: https://github.com/codeready-toolchain/toolchain-e2e/blob/master/dev_install.adoc#latest-greatest-dev-sandbox

mfrancisc: Cool ! Ok so we do have documentation about how to deploy the latest versions , in case I don't want/need to deploy my local version. Thanks !

PR #795, opened 2023-09-13
Add note about idler configuration

robobario:
Hi, as new users the idler feature baffled us for a while. We were trying to deploy workloads for the first time and saw the pods our operator was trying to start get deleted and recreated in a loop and took a while to track down where the delete came from. Maybe having some note about the idler here would help other users who want to onboard operators?

Alternatively, maybe the CLI tool (`go run setup/main.go --help`) could mention what it does. Like:

```
-i, --idler-timeout string     overrides the default idler timeout, pods in the user's namespace are deleted after this duration (default "15s")
```

Thanks :bow:


alexeykazakov (approved):
Good point! Thank you for the PR!

PR #797, opened 2023-09-20
feat: Implement public API that will open SpaceBindings to end users through proxy part 1

mfrancisc:
Update proxy (spacelister) e2e tests to check for the newly added `workspace.Status.AvailableRoles` field. The field should contain the Space roles for the NSTemplateTier. 
For example in the case of `appstudio` tier the list should be: `admin, maintainer, contributor`

Jira story: https://issues.redhat.com/browse/ASC-410

Related PRs:
- https://github.com/codeready-toolchain/toolchain-common/pull/327

Paired with: 
- https://github.com/codeready-toolchain/registration-service/pull/348

### DIFF
--- a/test/e2e/proxy_test.go
+++ b/test/e2e/proxy_test.go
@@ -683,7 +683,7 @@ func TestSpaceLister(t *testing.T) {
 		// then
 		// car should see only car's workspace
 		require.Len(t, workspaces, 1)
-		verifyHasExpectedWorkspace(t, expectedWorkspaceFor(t, awaitilities.Host(), users["car"], true), workspaces...)
+		verifyHasExpectedWorkspace(t, expectedWorkspaceFor(t, awaitilities.Host(), users["car"], commonproxy.WithType("home")), workspaces...)
 	})
 
 	t.Run("car gets workspaces", func(t *testing.T) {
@@ -693,7 +693,7 @@ func TestSpaceLister(t *testing.T) {
 
 			// then
 			require.NoError(t, err)
-			verifyHasExpectedWorkspace(t, expectedWorkspaceFor(t, awaitilities.Host(), users["car"], true, appStudioTierRolesWSOption), *workspace)
+			verifyHasExpectedWorkspace(t, expectedWorkspaceFor(t, awaitilities.Host(), users["car"], commonproxy.WithType("home"), appStudioTierRolesWSOption), *workspace)
 		})
 
 		t.Run("cannot get bus workspace", func(t *testing.T) {
@@ -713,8 +713,8 @@ func TestSpaceLister(t *testing.T) {
 		// then
 		// bus should see both its own and car's workspace
 		require.Len(t, workspaces, 2)
-		verifyHasExpectedWorkspace(t, expectedWorkspaceFor(t, awaitilities.Host(), users["bus"], true), workspaces...)
-		verifyHasExpectedWorkspace(t, expectedWorkspaceFor(t, awaitilities.Host(), users["car"], false), workspaces...)
+		verifyHasExpectedWorkspace(t, expectedWorkspaceFor(t, awaitilities.Host(), users["bus"], commonproxy.WithType("home")), workspaces...)
+		verifyHasExpectedWorkspace(t, expectedWorkspaceFor(t, awaitilities.Host(), users["car"]), workspaces...)
 	})
 
 	t.Run("bus gets workspaces", func(t *testing.T) {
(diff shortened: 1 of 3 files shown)


test/e2e/proxy_test.go
```diff
@@ -665,6 +665,7 @@ func TestSpaceLister(t *testing.T) {
 			identityID:            uuid.Must(uuid.NewV4()),
 		},
 	}
+	appStudioTierRolesWSoption := commonproxy.WithAvailableRoles([]string{"admin", "contributor", "maintainer"})
```
alexeykazakov, line 668: I think that a proper name would be 
```suggestion
	appStudioTierRolesWSOption := commonproxy.WithAvailableRoles([]string{"admin", "contributor", "maintainer"})
```
;)

test/e2e/proxy_test.go
```diff
@@ -692,7 +693,7 @@ func TestSpaceLister(t *testing.T) {
 
 			// then
 			require.NoError(t, err)
-			verifyHasExpectedWorkspace(t, expectedWorkspaceFor(t, awaitilities.Host(), users["car"], true), *workspace)
+			verifyHasExpectedWorkspace(t, expectedWorkspaceFor(t, awaitilities.Host(), users["car"], true, appStudioTierRolesWSoption), *workspace)
```
alexeykazakov, line 696: Is it enough to add the roles to the expected WS? Do we actually check anywhere that the expected roles present?

mfrancisc: From what I can see, by adding the roles to the expected WS the `verifyHasExpectedWorkspaceFor` function will compare the `actualWorkspace.Status` with the `expectedWorkspace.Status` here: https://github.com/codeready-toolchain/toolchain-e2e/blob/6135f39caa4a45a677cc46854e0a7efb0e502d49/test/e2e/proxy_test.go#L1047

so it should be fine, but unfortunately I didn't get to run this test till the end yet because of the other failure I've reported in the slack thread.

test/e2e/proxy_test.go
mfrancisc: yup , that's what I meant. 

done please see: https://github.com/codeready-toolchain/toolchain-e2e/pull/797/commits/1910a963765b2f448930f82b8cb2c4de84b84192

test/e2e/proxy_test.go
alexeykazakov: Yes, you are right. I missed the status check.

alexeykazakov (approved):
I guess the `/workspace/:name` case will be covered in the following PR(s)?

mfrancisc:
> I guess the /workspace/:name case will be covered in the following PR(s)?

@alexeykazakov  TBH I thought that updating the `TestSpaceLister` test and specifically the test cases that are getting a specific (work)space would be enough to cover my changes. But please let me know if I'm missing some tests.

alexeykazakov:
> TBH I thought that updating the TestSpaceLister test and specifically the test cases that are getting a specific (work)space would be enough to cover my changes. But please let me know if I'm missing some tests.

I didn't notice that we already have a test for `/workspace/:name`. I thought it's only for the list: `/workspace`. So, you are right. TBH I thought we do not currently support `/workspace/:name` at all. This is why I wanted to introduce a new handler (and a route) for that instead of modifying the existing handler.

mfrancisc:
/retest

reg-service PR was updated

test/e2e/proxy_test.go
```diff
@@ -773,7 +774,7 @@ func TestSpaceLister(t *testing.T) {
 
 			// then
 			require.NoError(t, err)
-			verifyHasExpectedWorkspace(t, expectedWorkspaceFor(t, awaitilities.Host(), users["bicycle"], true), *bicycleWS)
+			verifyHasExpectedWorkspace(t, expectedWorkspaceFor(t, awaitilities.Host(), users["bicycle"], true, appStudioTierRolesWSOption), *bicycleWS)
```
MatousJobanek, line 777: it's just a detail, but since you introduced the options, then you could update the function to:
```go
func expectedWorkspaceFor(t *testing.T, hostAwait *wait.HostAwaitility, user *proxyUser, additionalWSOptions ...commonproxy.WorkspaceOption) toolchainv1alpha1.Workspace {
```
and then call it:
```suggestion
			verifyHasExpectedWorkspace(t, expectedWorkspaceFor(t, awaitilities.Host(), users["bicycle"], commonproxy.WithType("home"), appStudioTierRolesWSOption), *bicycleWS)
```

MatousJobanek (approved):
Nice :+1:

test/e2e/proxy_test.go
mfrancisc: Good point! Thanks!

I've replaced the parameter with the option. Please see: https://github.com/codeready-toolchain/toolchain-e2e/pull/797/commits/2fb797da4667ef5afac1f6fb601caeef7b75ad42

PR #800, opened 2023-09-27
fix: allow insecure clusters

filariow:


alexeykazakov (approved):

rajivnathan (approved):

alexeykazakov:
OK. It failed :(
```
    init.go:68: 
        	Error Trace:	/go/src/github.com/codeready-toolchain/toolchain-e2e/testsupport/init.go:68
        	            				/usr/local/go/src/sync/once.go:74
        	            				/usr/local/go/src/sync/once.go:65
        	            				/go/src/github.com/codeready-toolchain/toolchain-e2e/testsupport/init.go:45
        	            				/go/src/github.com/codeready-toolchain/toolchain-e2e/test/migration/setup/setup_migration_test.go:12
        	            	specifying a root certificates file with the insecure flag is not allowed
        	Test:       	TestSetupMigration
--- FAIL: TestSetupMigration (0.00s)
```

testsupport/init.go
```diff
@@ -53,7 +54,12 @@ func WaitForDeployments(t *testing.T) wait.Awaitilities {
 
 		apiConfig, err := clientcmd.NewDefaultClientConfigLoadingRules().Load()
 		require.NoError(t, err)
-		kubeconfig, err := clientcmd.NewDefaultClientConfig(*apiConfig, &clientcmd.ConfigOverrides{}).ClientConfig()
+		configOverrides := clientcmd.ConfigOverrides{
+			ClusterDefaults: api.Cluster{
+				InsecureSkipTLSVerify: true,
```
dperaza4dustbit, line 59: Do we want to make this an option we can pass make instead of hardcoding to true? What if we want to test security?

dperaza4dustbit (commented):
looks good but one sugestion

testsupport/init.go
mfrancisc: +1

It seems that disabling the TLSVerify doesn't work on openshift CI at least. If this configuration works on your cluster then maybe we could introduce an env variable to optionally turn it off? 

Something like:
```
make test-e2e DISABLE_KUBE_CLIENT_TLS_VERIFY=true
````
and keep the tls verification on by default, as it is now.

Comment (mfrancisc) [reply testsupport/init.go]: Nice! Extracting the config logic builder into a function is what I was thinking as well! 👍 

Should we add few lines to document this new variable, maybe adding a know issues/workarounds section here: https://github.com/codeready-toolchain/toolchain-e2e/blob/master/CRC.adoc 
reporting the tls failure and the workaround of using this env variable ? 

If it's not specific to CRC maybe we could add this new section in the main README.

WDYT?

**Review by alexeykazakov — APPROVED** [review]
Looks good! But please update the README as @mfrancisc suggested.

File: testsupport/init.go (hunk 1 above)
Comment (filariow) [reply testsupport/init.go]: I like the idea, let me work on this! 👍🏾

Comment (filariow) [reply testsupport/init.go]: @mfrancisc  I've added a note in the README's `Running End-to-End Tests`. WDYT?

**Review by mfrancisc — APPROVED** [review]
Looks good 👍 

**Comment by filariow** [comment]
> Looks good 👍
> 
> Thanks for addressing my comments.

Thank you for your reviews. I pushed a new commit renaming the new function to a more coherent name.

File: README.adoc [hunk 2]
```diff
@@ -68,6 +68,8 @@ It will also create all required CRDs, role and role bindings for the service ac
 
 NOTE: you can override the default namespace names where the end-to-end tests are going to be executed - eg.: `make test-e2e HOST_NS=my-host MEMBER_NS=my-member` file.
 
+NOTE: you can disable HTTPS in tests setting the `DISABLE_KUBE_CLIENT_TLS_VERIFY` variable to `true` - eg.: `make test-e2e DISABLE_KUBE_CLIENT_TLS_VERIFY=true`. This flag helps when you test in clusters using Self-Signed Certificates.
```
baijum, line 71: It's not disabling HTTPS, I would suggest something like this:
```suggestion
NOTE: you can disable SSL/TLS certificate verification in tests setting the `DISABLE_KUBE_CLIENT_TLS_VERIFY` variable to `true` - eg.: `make test-e2e DISABLE_KUBE_CLIENT_TLS_VERIFY=true`. This flag helps when you test in clusters using Self-Signed Certificates.
```

filariow: you're right! good catch, let me fix that

PR #807, opened 2023-10-03
verify that targetClustersfor UserAccount is dynamically loaded

MatousJobanek:
paired PR https://github.com/codeready-toolchain/host-operator/pull/872

### DIFF
--- a/test/e2e/parallel/retarget_test.go
+++ b/test/e2e/parallel/retarget_test.go
@@ -25,7 +25,6 @@ func TestRetargetUserByChangingSpaceTargetCluster(t *testing.T) {
 		TargetCluster(member1Await).
 		EnsureMUR().
 		RequireConditions(ConditionSet(Default(), ApprovedByAdmin())...).
-		DisableCleanup().
 		Execute(t).Resources()
 
 	space, err := hostAwait.WaitForSpace(t, userSignup.Status.CompliantUsername,
--- a/test/e2e/parallel/retarget_test.go
+++ b/test/e2e/parallel/retarget_test.go
@@ -1,0 +1,52 @@
+package parallel
+
+import (
+	"testing"
+
+	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
+	. "github.com/codeready-toolchain/toolchain-e2e/testsupport"
+	. "github.com/codeready-toolchain/toolchain-e2e/testsupport/space"
+	. "github.com/codeready-toolchain/toolchain-e2e/testsupport/wait"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestRetargetUserByChangingSpaceTargetCluster(t *testing.T) {
+	// given
+	t.Parallel()
+
+	awaitilities := WaitForDeployments(t)
+	hostAwait := awaitilities.Host()
+	member1Await := awaitilities.Member1()
+	member2Await := awaitilities.Member2()
+
+	userSignup, mur := NewSignupRequest(awaitilities).
+		ManuallyApprove().
+		TargetCluster(member1Await).
+		EnsureMUR().
+		RequireConditions(ConditionSet(Default(), ApprovedByAdmin())...).
+		DisableCleanup().
+		Execute(t).Resources()
+
+	space, err := hostAwait.WaitForSpace(t, userSignup.Status.CompliantUsername,
+		UntilSpaceHasStatusTargetCluster(member1Await.ClusterName),
+		UntilSpaceHasAnyTierNameSet())
+	require.NoError(t, err)
+
+	// when
+	space, err = hostAwait.UpdateSpace(t, space.Name, func(s *toolchainv1alpha1.Space) {
+		s.Spec.TargetCluster = member2Await.ClusterName
+	})
+	require.NoError(t, err)
+
+	// then
+	// expect NSTemplateSet to be deleted on member-1 cluster
+	err = member1Await.WaitUntilNSTemplateSetDeleted(t, space.Name)
+	require.NoError(t, err)
+
+	// and provisioned on member-2
+	_, err = member2Await.WaitForNSTmplSet(t, space.Name)
+	require.NoError(t, err)
+	VerifyResourcesProvisionedForSpace(t, awaitilities, space.Name, UntilSpaceHasStatusTargetCluster(member2Await.ClusterName))
+	VerifyUserRelatedResources(t, awaitilities, userSignup, mur.Spec.TierName, ExpectUserAccountIn(member2Await))
+}
--- a/test/e2e/parallel/retarget_test.go
+++ b/test/e2e/parallel/retarget_test.go
@@ -1,20 +1,21 @@
 package parallel
 
 import (
+	"os"
 	"testing"
 
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
 	. "github.com/codeready-toolchain/toolchain-e2e/testsupport"
 	. "github.com/codeready-toolchain/toolchain-e2e/testsupport/space"
+	"github.com/codeready-toolchain/toolchain-e2e/testsupport/spacebinding"
 	. "github.com/codeready-toolchain/toolchain-e2e/testsupport/wait"
 
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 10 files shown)


test/e2e/parallel/retarget_test.go
```diff
@@ -0,0 +1,51 @@
...
+func TestRetargetUserByChangingSpaceTargetCluster(t *testing.T) {
+	// given
+	t.Parallel()
+
+	awaitilities := WaitForDeployments(t)
+	hostAwait := awaitilities.Host()
+	member1Await := awaitilities.Member1()
+	member2Await := awaitilities.Member2()
+
+	userSignup, mur := NewSignupRequest(awaitilities).
```
alexeykazakov, line 23: What do you think about also sharing this space with another user? So the test would cover more complicated case with a space with spacebindgings?

test/e2e/usersignup_test.go
```diff
@@ -197,7 +197,7 @@ func (s *userSignupIntegrationTest) TestProvisionToOtherClusterWhenOneIsFull() {
 			Execute(s.T()).Resources()
 
 		// then
-		require.NotEqual(t, mur1.Spec.UserAccounts[0].TargetCluster, mur2.Spec.UserAccounts[0].TargetCluster)
+		require.NotEqual(t, mur1.Status.UserAccounts[0].Cluster.Name, mur2.Status.UserAccounts[0].Cluster.Name)
```
mfrancisc, line 200: I couldn't find where `mur.Status.UserAccounts.Cluster.Name` is being set in the host-operator and member-operator, I see it's being checked but not set. I'm sure I've missed something. 

But should we use `hostAwait.WaitForSpace(t, mur1.Name, wait.UntilSpaceHasAnyTargetClusterSet())` and same for `mur2` here as well ?

mfrancisc (commented):
Nice 👍 

I left only few minor comments/questions.

test/e2e/parallel/retarget_test.go
MatousJobanek: good idea - added another test case ffac8bc0

MatousJobanek: :man_facepalming:  53353e3d

test/e2e/usersignup_test.go
MatousJobanek: > I couldn't find where mur.Status.UserAccounts.Cluster.Name is being set in the host-operator and member-operator, I see it's being checked but not set. I'm sure I've missed something.

I'm not sure if I follow what you mean. Don't forget that this is a status, so it's set by a MUR controller when it syncs the information with UserAccount https://github.com/codeready-toolchain/host-operator/blob/927edad1f670889d02485c1536a19009bec48e18/controllers/masteruserrecord/sync.go#L95

> But should we use hostAwait.WaitForSpace(t, mur1.Name, wait.UntilSpaceHasAnyTargetClusterSet()) and same for mur2 here as well ?

I probably don't follow your message again. Do you mean that we should have an extra check that would wait until the Space has the needed targetCluster set, etc...? This is already done in the "SignupRequest" fluent API logic - we call it with `EnsureMUR()` which ensures both the MUR and Space are fully provisoned. I know that the name is misleading - it's another thing to improve but in separate PR, I don't want to mix it with the functional changes.

testsupport/user_assertions.go
```diff
@@ -50,29 +73,58 @@ func VerifyUserRelatedResources(t *testing.T, awaitilities wait.Awaitilities, si
...
+		userAccountStatusWaitCriterion)
 	require.NoError(t, err)
 
-	memberAwait := GetMurTargetMember(t, awaitilities, mur)
+	// Verify last target cluster annotation is set
+	lastCluster, foundLastCluster := userSignup.Annotations[toolchainv1alpha1.UserSignupLastTargetClusterAnnotationKey]
+	require.True(t, foundLastCluster)
+	// todo - change that as soon as we set the last-target-cluster annotation based on the "home" Space location
+	// todo - and as soon as we drop the embedded UserAccounts from MUR
+	require.Equal(t, mur.Spec.UserAccounts[0].TargetCluster, lastCluster)
```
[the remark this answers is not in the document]
MatousJobanek: not really - this is one of the things to be modified in the next stories/PRs. The annotation is set when we provision MUR - based on the information that we set directly in the MUR, and not based on what is in Space nor MUR.Status part.

testsupport/wait/host.go
```diff
@@ -471,6 +471,46 @@ func UntilMasterUserRecordHasUserAccountStatuses(expected ...toolchainv1alpha1.U
...
+					if ua.Cluster.Name == expectedCluster {
+						found = true
+						break
+					}
+				}
+				if !found {
+					return false
+				}
+			}
+			return true
```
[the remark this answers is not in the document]
MatousJobanek: Content-wise, there is no difference between userAccount1 and userAccount2. They both contain the very same content. The only difference is where they are provisioned.
If a user has access to 100 Spaces and all of them would be provisioned in the same cluster, then there would be only one single UserAccount for the MUR created. If the Spaces would be provisioned to 100 different member clusters, then the MUR would have 100 of UserAccount "replicas", and thus 100 UserAccount items in its status.

The only difference would be if we wanted to propagate the information about the "home" Space to the MUR status as well. In that case, we would need to mark which item in the status corresponds to the cluster where the "home" Space is provisioned. That's all.

test/e2e/usersignup_test.go
mfrancisc: > I'm not sure if I follow what you mean. Don't forget that this is a status, so it's set by a MUR controller when it syncs the information with UserAccount https://github.com/codeready-toolchain/host-operator/blob/927edad1f670889d02485c1536a19009bec48e18/controllers/masteruserrecord/sync.go#L95

Ok I see now that the `mur.Status.UserAccount.Cluster.Name` is being set for the fist time here: I'm not sure if I follow what you mean. Don't forget that this is a status, so it's set by a MUR controller when it syncs the information with UserAccount https://github.com/codeready-toolchain/host-operator/blob/927edad1f670889d02485c1536a19009bec48e18/controllers/masteruserrecord/sync.go#L95

And the UserAccount is being added to the status or updated based on `index` < `0` here:https://github.com/codeready-toolchain/host-operator/blob/927edad1f670889d02485c1536a19009bec48e18/controllers/masteruserrecord/sync.go#L103-L107

Thanks for the pointer!

mfrancisc: > I probably don't follow your message again. Do you mean that we should have an extra check that would wait until the Space has the needed targetCluster set, etc...? 

Basically I was wondering if we should rely on this `MUR.Status.UserAccounts[0].Cluster.Name` field or if we should get the Spaces for both mur1 and mur1 like we do here for example https://github.com/codeready-toolchain/toolchain-e2e/pull/807/files#diff-d5d12b6ea07b0a78c9ad1987e1486c565ecf16839f77a93e91b09475cf6c0e99R148 and use the `Space.Spec.TargetCluster` for the comparison. But I guess this `MUR.Status.UserAccounts[0].Cluster.Name` will always be up-to-date.

testsupport/user_assertions.go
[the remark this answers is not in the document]
mfrancisc: Got it ! Thanks for explaining that!

testsupport/wait/host.go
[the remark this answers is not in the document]
mfrancisc: Makes sense! Thanks for explaining that!

mfrancisc (approved):
Nice Job 🏅 

Thanks for answering my questions 🙏

test/e2e/parallel/retarget_test.go
mfrancisc: > 🤦‍♂️ https://github.com/codeready-toolchain/toolchain-e2e/commit/53353e3d9d02520cf5ad0120bf3be0205776056b

sometimes I push similar commits 😄 , maybe I'll try to create git pre-commit hook for it.

test/e2e/usersignup_test.go
MatousJobanek: yeah, good idea: done here d424d5ae

mfrancisc: Nice! Makes sense to have it in addition to the existing check 👍 

Thanks!

PR #808, opened 2023-10-03
Update documentation for usage with quay.io

sadlerap:
If a user intends to run end-to-end tests using quay as a repository, then the image `quay.io/<quay-username>/member-operator-console-plugin` needs to be publicly available.  Update the documentation for quay to reflect this.


alexeykazakov (approved):
Good catch.

PR #813, opened 2023-10-09
add test to confirm that propagatedclaims are correctly propagated al…

sbryzak:
…l the way from UserSignup to UserAccount

Related PR: https://github.com/codeready-toolchain/host-operator/pull/880


test/e2e/usersignup_test.go
```diff
@@ -284,6 +284,39 @@ func (s *userSignupIntegrationTest) TestUserResourcesCreatedWhenOriginalSubIsSet
...
+
+	// Update the UserSignup
+	userSignup, err := hostAwait.UpdateUserSignup(s.T(), userSignup.Name, func(us *toolchainv1alpha1.UserSignup) {
+		// Modify the user's AccountID
+		us.Spec.IdentityClaims.AccountID = "nnnbbb111234"
+	})
+	require.NoError(s.T(), err)
+
+	// Confirm the AccountID is updated
+	require.Equal(s.T(), "nnnbbb111234", userSignup.Spec.IdentityClaims.AccountID)
```
alexeykazakov, line 314: Hm... Doesn't it just test the `hostAwait.UpdateUserSignup()` function? :)
Shouldn't you check the `UserAccount.Spec.PropogatedClaims` have been updated instead (not the UserSignup!)? And maybe the `Mur.Spec.PropogatedClaims` too if we do want to check all intermediate resources ;)

sbryzak: The `UserAccount.Spec.PropagatedClaims` is checked in `VerifyResourcesProvisionedForSignup` (technically in `VerifyUserRelatedResources` in `user_assertions.go`) on line 317, just below this code.  The MUR also undergoes some verification, but not its PropagatedClaims - however since the `UserAccount` is verified in this very same function, then it's impossible for MUR to have different values as it is the source of the property values propagated to `UserAccount` in the first place.

alexeykazakov: Ok. Sounds good then.

alexeykazakov (approved):

test/e2e/usersignup_test.go
MatousJobanek: for better readability, it would be nice to add a specific wait (before `VerifyResourcesProvisionedForSignup`) that would verify that the value is propagated down to the UserAccount.  I know that `VerifyResourcesProvisionedForSignup` checks that the UserAccount has the same values of the propagatedClaims as is in UserSignup CR, but having it here specified explicitly as well would be better for reading the code in the future

PR #822, opened 2023-10-17
Re-enable webhook checks on migration setup test

rajivnathan:


testsupport/wait/member.go
```diff
@@ -2240,12 +2239,6 @@ func (a *MemberAwaitility) verifySecret(t *testing.T) []byte {
 }
 
 func (a *MemberAwaitility) verifyMutatingWebhookConfig(t *testing.T, ca []byte) {
-	if val := os.Getenv("skip-webhook-checks-on-setup"); val == "true" {
```
alexeykazakov, line 2243: Or we could keep it and just out the setting of this environment variable in the migration tests for future use in the comment (and keep the description of what it is for). But I will keep it up to you to decide :)

alexeykazakov (approved):

mfrancisc (approved):

testsupport/wait/member.go
rajivnathan: Actually, I was thinking we probably don't need to verify webhooks as part of the migration setup tests. I just had a discussion with @MatousJobanek about this and he agreed. So I think we can simplify it by doing some refactoring to skip verification of webhook configuration and maybe some other things too so that we don't need to do this dance repeatedly whenever we make such changes. I'll handle that in a separate PR.

PR #823, opened 2023-10-23
Improve perf tool error msg

rajivnathan:
Improve the error message and README related to logging in to the cluster because a token is needed for gathering metrics


mfrancisc (approved):
Thanks for the improvement 👍

PR #824, opened 2023-10-23
Check for additional CNV namespace policy

rajivnathan:
Updated tests for https://github.com/codeready-toolchain/host-operator/pull/903

### DIFF
--- a/testsupport/tiers/checks.go
+++ b/testsupport/tiers/checks.go
@@ -1035,7 +1035,7 @@ func networkPolicyAllowFromMonitoring() namespaceObjectsCheck {
 }
 
 func networkPolicyAllowFromOlmNamespaces() namespaceObjectsCheck {
-	return networkPolicyIngress("allow-from-olm-namespaces", "openshift.io/scc", "anyuid")
+	return assertNetworkPolicyIngressForNamespaces("allow-from-olm-namespaces", "openshift.io/scc", "anyuid")
 }
 
 func networkPolicyAllowFromConsoleNamespaces() namespaceObjectsCheck {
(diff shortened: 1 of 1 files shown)


alexeykazakov (approved):

testsupport/tiers/checks.go
```diff
@@ -1054,20 +1054,25 @@ func networkPolicyIngressFromPolicyGroup(name, group string) namespaceObjectsChe
...
 		expected := &netv1.NetworkPolicy{
 			Spec: netv1.NetworkPolicySpec{
 				Ingress: []netv1.NetworkPolicyIngressRule{
 					{
-						From: []netv1.NetworkPolicyPeer{
-							{
-								NamespaceSelector: &metav1.LabelSelector{MatchLabels: map[string]string{labelName: labelValue}},
-							},
-						},
+						From: peers,
```
MatousJobanek, line 1075: This doesn't match what is in the template. 
In the template you define two Ingress rules:
```yaml
  spec:
    ingress:
    - from:
      - namespaceSelector:
          matchLabels:
            kubernetes.io/metadata.name: openshift-virtualization-os-images
    - from:
      - namespaceSelector:
          matchLabels:
            kubernetes.io/metadata.name: openshift-cnv
```
while here you expect two label selectors for the same rule. that would correspond to:
```
  spec:
    ingress:
    - from:
      - namespaceSelector:
          matchLabels:
            kubernetes.io/metadata.name: openshift-virtualization-os-images
      - namespaceSelector:
          matchLabels:
            kubernetes.io/metadata.name: openshift-cnv
```
TBH, not sure what is the difference between these two formats functional-wise

MatousJobanek (approved):
please unify the format between the template and the expectations in the tests

testsupport/tiers/checks.go
rajivnathan: ah, good catch!

rajivnathan: Looking at https://kubernetes.io/docs/concepts/services-networking/network-policies/#behavior-of-to-and-from-selectors I think the two Ingress rules is the correct way for our use case since we want ingress traffic from both namespaces.

rajivnathan: Should be fixed with 067813d170f1b18bd48f71c82a9c420ce5d69b06

PR #826, opened 2023-10-24
add migration assertions for space requests

sadlerap:
Make sure that existing spaces don't fall over during migrations.

See also: https://github.com/codeready-toolchain/host-operator/pull/891

### DIFF
--- a/test/migration/setup_runner.go
+++ b/test/migration/setup_runner.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
+	commoncluster "github.com/codeready-toolchain/toolchain-common/pkg/cluster"
 	"github.com/codeready-toolchain/toolchain-common/pkg/states"
 	testspace "github.com/codeready-toolchain/toolchain-common/pkg/test/space"
 	"github.com/codeready-toolchain/toolchain-e2e/testsupport"
@@ -28,6 +29,8 @@ const (
 
 	ProvisionedAppStudioSpace    = "mig-appst-space"
 	SecondMemberProvisionedSpace = "mig-m2-space"
+	ProvisionedSpaceRequest      = "mig-space-request"
+	ProvisionedParentSpace       = "mig-parent-space"
 )
 
 type SetupMigrationRunner struct {
@@ -41,6 +44,7 @@ func (r *SetupMigrationRunner) Run(t *testing.T) {
 	toRun := []func(t *testing.T){
 		r.prepareAppStudioProvisionedSpace,
 		r.prepareSecondMemberProvisionedSpace,
+		r.prepareProvisionedSubspace,
 		r.prepareProvisionedUser,
 		r.prepareSecondMemberProvisionedUser,
 		r.prepareDeactivatedUser,
@@ -89,6 +93,29 @@ func (r *SetupMigrationRunner) createAndWaitForSpace(t *testing.T, name, tierNam
 	if r.WithCleanup {
 		cleanup.AddCleanTasks(t, r.Awaitilities.Host().Client, space)
 	}
+}
+
+func (r *SetupMigrationRunner) prepareProvisionedSubspace(t *testing.T) {
+	memberAwait := r.Awaitilities.Member2()
+	r.createAndWaitForSpace(t, ProvisionedParentSpace, "base", memberAwait)
+
+	srClusterRoles := []string{commoncluster.RoleLabel(commoncluster.Tenant)}
+	t.Logf("creating space request %v for parent space %v", ProvisionedSpaceRequest, ProvisionedParentSpace)
+	spaceRequest := tsspace.CreateSpaceRequestForParentSpace(t,
+		r.Awaitilities,
+		memberAwait.ClusterName,
+		ProvisionedParentSpace,
+		tsspace.WithName(ProvisionedSpaceRequest),
+		tsspace.WithSpecTargetClusterRoles(srClusterRoles),
+		tsspace.WithSpecTierName("base"))
+
+	_, err := r.Awaitilities.Host().WaitForSubSpace(t,
+		spaceRequest.GetName(),
+		spaceRequest.GetNamespace(),
+		ProvisionedParentSpace,
+		wait.UntilSpaceHasConditions(wait.Provisioned()),
+		wait.UntilSpaceHasAnyProvisionedNamespaces())
+	require.NoError(t, err)
 }
 
 func (r *SetupMigrationRunner) prepareProvisionedUser(t *testing.T) {
--- a/test/migration/setup_runner.go
+++ b/test/migration/setup_runner.go
@@ -101,7 +101,7 @@ func (r *SetupMigrationRunner) createAndWaitForSpace(t *testing.T, name, tierNam
 func (r *SetupMigrationRunner) prepareProvisionedSubspace(t *testing.T) {
 	hostAwait := r.Awaitilities.Host()
 	memberAwait := r.Awaitilities.Member2()
-	r.createAndWaitForSpace(t, ProvisionedParentSpace, "base", memberAwait)
+	r.createAndWaitForSpace(t, ProvisionedParentSpace, "appstudio-env", memberAwait)
 	memberCluster, found, err := hostAwait.GetToolchainCluster(t, cluster.Member, memberAwait.Namespace, nil)
 	require.NoError(t, err)
 	require.True(t, found)
@@ -114,7 +114,7 @@ func (r *SetupMigrationRunner) prepareProvisionedSubspace(t *testing.T) {
 		ProvisionedParentSpace,
 		tsspace.WithName(ProvisionedSpaceRequest),
 		tsspace.WithSpecTargetClusterRoles(srClusterRoles),
-		tsspace.WithSpecTierName("base"))
+		tsspace.WithSpecTierName("appstudio-env"))
 
 	subSpace, err := hostAwait.WaitForSubSpace(t,
 		spaceRequest.GetName(),
--- a/test/migration/setup_runner.go
+++ b/test/migration/setup_runner.go
@@ -2,12 +2,15 @@ package migration
 
 import (
 	"context"
-	"github.com/gofrs/uuid"
 	"sync"
 	"testing"
 	"time"
 
+	"github.com/gofrs/uuid"
+	"k8s.io/apimachinery/pkg/types"
+
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
+	"github.com/codeready-toolchain/toolchain-common/pkg/cluster"
 	commoncluster "github.com/codeready-toolchain/toolchain-common/pkg/cluster"
 	"github.com/codeready-toolchain/toolchain-common/pkg/states"
 	testspace "github.com/codeready-toolchain/toolchain-common/pkg/test/space"
@@ -96,8 +99,12 @@ func (r *SetupMigrationRunner) createAndWaitForSpace(t *testing.T, name, tierNam
 }
 
 func (r *SetupMigrationRunner) prepareProvisionedSubspace(t *testing.T) {
+	hostAwait := r.Awaitilities.Host()
 	memberAwait := r.Awaitilities.Member2()
 	r.createAndWaitForSpace(t, ProvisionedParentSpace, "base", memberAwait)
+	memberCluster, found, err := hostAwait.GetToolchainCluster(t, cluster.Member, memberAwait.Namespace, nil)
+	require.NoError(t, err)
+	require.True(t, found)
 
 	srClusterRoles := []string{commoncluster.RoleLabel(commoncluster.Tenant)}
 	t.Logf("creating space request %v for parent space %v", ProvisionedSpaceRequest, ProvisionedParentSpace)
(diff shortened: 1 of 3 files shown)


test/migration/verify/verify_migration_test.go
```diff
@@ -119,6 +122,43 @@ func verifySecondMemberProvisionedSpace(t *testing.T, awaitilities wait.Awaitili
 	cleanup.AddCleanTasks(t, awaitilities.Host().Client, userSignupForSpace)
 }
 
+func verifyProvisionedSubSpace(t *testing.T, awaitilities wait.Awaitilities) {
+	parentSpace, _ := VerifyResourcesProvisionedForSpace(t, awaitilities, migration.ProvisionedParentSpace)
+	userSignupForSpace := checkMURMigratedAndGetSignup(t, awaitilities.Host(), migration.ProvisionedParentSpace)
```
mfrancisc, line 127: unless I've missed something - the parentSpace is created using `createAndWaitForSpace` which doesn't provision a usersignup resource.

test/migration/verify/verify_migration_test.go
```diff
@@ -119,6 +122,43 @@ func verifySecondMemberProvisionedSpace(t *testing.T, awaitilities wait.Awaitili
...
+	targetClusterRoles := []string{cluster.RoleLabel(cluster.Tenant)}
+	hostAwait := awaitilities.Host()
+	memberAwait := awaitilities.Member2()
+	memberCluster, found, err := hostAwait.GetToolchainCluster(t, cluster.Member, memberAwait.Namespace, nil)
+	require.NoError(t, err)
+	require.True(t, found)
+
+	subSpace, err := awaitilities.Host().WaitForSubSpace(t,
+		migration.ProvisionedSpaceRequest,
+		memberAwait.Namespace,
```
mfrancisc, line 137: the spacerequest is not created in the `toolchain-member-operator` namespace, this should be `GetDefaultNamespace(parentSpace.Status.ProvisionedNamespaces)`

test/migration/verify/verify_migration_test.go
```diff
@@ -119,6 +122,43 @@ func verifySecondMemberProvisionedSpace(t *testing.T, awaitilities wait.Awaitili
...
+		parentSpace.GetName(),
+		wait.UntilSpaceHasTargetClusterRoles(targetClusterRoles),
+		wait.UntilSpaceHasTier("base"),
+		wait.UntilSpaceHasAnyProvisionedNamespaces())
+	require.NoError(t, err)
+
+	spaceRequest, err := memberAwait.WaitForSpaceRequest(t,
+		types.NamespacedName{
+			Namespace: memberAwait.Namespace,
```
mfrancisc, line 146: same here - the namespace come from the default namespace in parentSpace

test/migration/verify/verify_migration_test.go
```diff
@@ -119,6 +122,43 @@ func verifySecondMemberProvisionedSpace(t *testing.T, awaitilities wait.Awaitili
...
+		},
+		wait.UntilSpaceRequestHasConditions(wait.Provisioned()),
+		wait.UntilSpaceRequestHasStatusTargetClusterURL(memberCluster.Spec.APIEndpoint),
+		wait.UntilSpaceRequestHasNamespaceAccess(subSpace),
+	)
+	require.NoError(t, err)
+	VerifyNamespaceAccessForSpaceRequest(t, memberAwait.Client, spaceRequest)
+
+	cleanup.AddCleanTasks(t, awaitilities.Host().Client, subSpace)
+	cleanup.AddCleanTasks(t, awaitilities.Member1().Client, spaceRequest)
```
mfrancisc, line 157: minor detail - seems like you are using member2 for the creation of the spacerequest and not member1. Probably the cleanup will still work since it's there is only 1 cluster in e2e mode.

testsupport/space/spacerequest.go
```diff
@@ -59,6 +65,25 @@ func CreateSpaceRequest(t *testing.T, awaitilities wait.Awaitilities, memberName
 	return spaceRequest, parentSpace
 }
 
+func CreateSpaceRequestForParentSpace(t *testing.T, awaitilities wait.Awaitilities, memberName, parent string, opts ...SpaceRequestOption) *toolchainv1alpha1.SpaceRequest {
```
mfrancisc, line 68: minor thing - I was wondering if we could do something regarding the duplicate parts with `CreateSpaceRequest` function. Probably a flag or option to get the parentSpace (which will create it or just pass a name). But feel free to leave it as is.

mfrancisc (commented):
Looks good! Thanks for adding a migration check for SpaceRequests.   
I have few comments.

test/migration/verify/verify_migration_test.go
sadlerap: Yes, it's made with `createAndWaitForSpace`, but the other spaces in migration setup [also call this function][1].

[1]: https://github.com/sadlerap/toolchain-e2e/blob/subspace/test/migration/setup_runner.go#L65-L71

mfrancisc: My bad, I saw now that it actually creates the MUR and UserSignup [here](https://github.com/codeready-toolchain/toolchain-e2e/blob/e2addcbf280e64bc040a281bf986bc0b273d168f/test/migration/setup_runner.go#L75) 🤦‍♂️ . Sorry for the confusion.

test/migration/verify/verify_migration_test.go
sadlerap: yep, I'm intending on using `member2` for this test.

testsupport/space/spacerequest.go
sadlerap: Idea: check if the parent space is `nil`: if it is, create it, otherwise use that's provided.

test/migration/verify/verify_migration_test.go
sadlerap: `createAndWaitForSpace` calls [`tsspace.CreateMurWithAdminSpaceBindingForSpace`][1], which creates both the `UserSignup` and the `MasterUserRecord` [here][2].

[1]: https://github.com/codeready-toolchain/toolchain-e2e/blob/e2addcbf280e64bc040a281bf986bc0b273d168f/test/migration/setup_runner.go#L75C20-L75C67
[2]: https://github.com/codeready-toolchain/toolchain-e2e/blob/e2addcbf280e64bc040a281bf986bc0b273d168f/testsupport/space/space.go#L180

testsupport/space/spacerequest.go
mfrancisc: makes sense, if it's not too much refactoring 👍

sadlerap: I attempted a refactoring, and I wound up with this:

```go
func CreateSpaceRequest(t *testing.T, awaitilities wait.Awaitilities, parentSpaceName, memberName string, opts ...SpaceRequestOption) (*toolchainv1alpha1.SpaceRequest, *toolchainv1alpha1.Space) {
	require.NoError(t, err)
	parentSpace := &toolchainv1alpha1.Space{}

	providedParentSpace := len(parentSpaceName) == 0
	if providedParentSpace {
		// let's first create a parentSpace
		parentSpace, _, _ = CreateSpace(t, awaitilities, testspace.WithTierName("appstudio"), testspace.WithSpecTargetCluster(memberAwait.ClusterName))
		// wait for the namespace to be provisioned since we will be creating the spacerequest into it.
		parentSpace, err = awaitilities.Host().WaitForSpace(t, parentSpace.Name, wait.UntilSpaceHasAnyProvisionedNamespaces())
		require.NoError(t, err)
	} else {
		err = awaitilities.Host().Client.Get(context.TODO(), types.NamespacedName{Namespace: awaitilities.Host().Namespace, Name: parentSpaceName}, parentSpace)
		require.NoError(t, err)
	}

	// create the space request in the "default" namespace provisioned by the parentSpace
	spaceRequest := NewSpaceRequest(t, append(opts, WithNamespace(GetDefaultNamespace(parentSpace.Status.ProvisionedNamespaces)))...)
	require.NotEmpty(t, spaceRequest)

	if providedParentSpace {
		// if we were given a parentSpace, don't clean up the space request on
		// test completion.  Needed for subspace migration tests.
		err = memberAwait.Create(t, spaceRequest)
	} else {
		err = memberAwait.CreateWithCleanup(t, spaceRequest)
	}
	require.NoError(t, err)

	return spaceRequest, parentSpace
}
```

I'm not entirely convinced it's a benefit, since its branches depend solely on whether a string was empty or not.  Considering that `CreateSpaceRequest` is only used in parallel e2e testing, while `CreateSpaceRequestForParentSpace` would be used in migration testing, I for one feel fine leaving them separate.  The creation and the usage of the SpaceRequest are different enough to me to warrant different functions here.

If you're fine with it, I'll leave this for now - we can always revisit this in the future.

test/migration/verify/verify_migration_test.go
sadlerap: Fixed in 1048498.

test/migration/verify/verify_migration_test.go
sadlerap: Fixed in 1048498.

test/migration/verify/verify_migration_test.go
sadlerap: Fixed in 1048498.

sadlerap:
hmm, I'll take a look at those failures tomorrow - not sure what I changed to break things.

sadlerap:
I think I've found the issue causing test failures, but I'm not sure it's an issue with my change or if it's a pre-existing bug that my change uncovers.  It seems that we have a service account `namespace-manager` that's required for spacerequests to successfully resolve, but I can't find any reference to it in e2e tests.  I believe this causes the `SpaceRequest` to not have access to the namespaces of the `Space` it owns.

test/migration/setup_runner.go
```diff
@@ -91,6 +98,44 @@ func (r *SetupMigrationRunner) createAndWaitForSpace(t *testing.T, name, tierNam
...
+
+	_, err = memberAwait.WaitForSpaceRequest(t,
+		types.NamespacedName{
+			Namespace: spaceRequest.Namespace,
+			Name:      spaceRequest.Name,
+		},
+		wait.UntilSpaceRequestHasStatusTargetClusterURL(memberCluster.Spec.APIEndpoint),
+		wait.UntilSpaceRequestHasNamespaceAccess(subSpace),
+		wait.UntilSpaceRequestHasConditions(wait.Provisioned()),
+	)
```
MatousJobanek, line 135: you could significantly simplify it to decrease the dependency on the actual implementation

Let's assume that whatever is in master just works, we don't need to verify all the details, just wait until the SpaceRequest is completed. The actual verification should be executed when the operator is updated with the changed version.

```suggestion
	_, err = memberAwait.WaitForSpaceRequest(t,
		types.NamespacedName{
			Namespace: spaceRequest.Namespace,
			Name:      spaceRequest.Name,
		},
		wait.UntilSpaceRequestHasConditions(wait.Provisioned()),
	)
```

test/migration/verify/verify_migration_test.go
```diff
@@ -119,6 +122,44 @@ func verifySecondMemberProvisionedSpace(t *testing.T, awaitilities wait.Awaitili
...
+	require.NoError(t, err)
+	require.True(t, found)
+	subSpaceNamespace := GetDefaultNamespace(parentSpace.Status.ProvisionedNamespaces)
+
+	subSpace, err := awaitilities.Host().WaitForSubSpace(t,
+		migration.ProvisionedSpaceRequest,
+		subSpaceNamespace,
+		parentSpace.GetName(),
+		wait.UntilSpaceHasTargetClusterRoles(targetClusterRoles),
+		wait.UntilSpaceHasTier("base"),
```
MatousJobanek, line 141:
```suggestion
		wait.UntilSpaceHasTier("appstudio-env"),
```

test/migration/verify/verify_migration_test.go
```diff
@@ -119,6 +122,44 @@ func verifySecondMemberProvisionedSpace(t *testing.T, awaitilities wait.Awaitili
...
+		wait.UntilSpaceRequestHasStatusTargetClusterURL(memberCluster.Spec.APIEndpoint),
+		wait.UntilSpaceRequestHasNamespaceAccess(subSpace),
+		wait.UntilSpaceRequestHasConditions(wait.Provisioned()),
+	)
+	require.NoError(t, err)
+	VerifyNamespaceAccessForSpaceRequest(t, memberAwait.Client, spaceRequest)
+
+	cleanup.AddCleanTasks(t, awaitilities.Host().Client, subSpace)
+	cleanup.AddCleanTasks(t, awaitilities.Member2().Client, spaceRequest)
+	cleanup.AddCleanTasks(t, awaitilities.Host().Client, parentSpace)
```
MatousJobanek, line 159: the subSpace is managed by SpaceRequest, if you try to delete it before deleting SpaceRequest, then the SpaceRequest controller will re-create it.
technically speaking, the only thing you need to do is to delete the parent-space, the rest will be cleaned up automatically by the GC and via finalizers
```suggestion
	cleanup.AddCleanTasks(t, awaitilities.Host().Client, parentSpace)
```

test/migration/setup_runner.go
```diff
@@ -91,6 +98,44 @@ func (r *SetupMigrationRunner) createAndWaitForSpace(t *testing.T, name, tierNam
...
+
+	srClusterRoles := []string{commoncluster.RoleLabel(commoncluster.Tenant)}
+	t.Logf("creating space request %v for parent space %v", ProvisionedSpaceRequest, ProvisionedParentSpace)
+	spaceRequest := tsspace.CreateSpaceRequestForParentSpace(t,
+		r.Awaitilities,
+		memberAwait.ClusterName,
+		ProvisionedParentSpace,
+		tsspace.WithName(ProvisionedSpaceRequest),
+		tsspace.WithSpecTargetClusterRoles(srClusterRoles),
+		tsspace.WithSpecTierName("base"))
```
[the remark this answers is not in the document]
sadlerap: Fixed with 6246d5e.

test/migration/verify/verify_migration_test.go
sadlerap: Fixed with 6246d5e.

test/migration/setup_runner.go
sadlerap: Fixed in 30ff3cf.

test/migration/verify/verify_migration_test.go
sadlerap: Fixed in 30ff3cf.

testsupport/space/spacerequest.go
```diff
@@ -59,6 +66,25 @@ func CreateSpaceRequest(t *testing.T, awaitilities wait.Awaitilities, memberName
...
+	// create the space request in the "default" namespace provisioned by the parentSpace
+	namespace := GetDefaultNamespace(parentSpace.Status.ProvisionedNamespaces)
+	spaceRequest := NewSpaceRequest(t, append(opts, WithNamespace(namespace))...)
+	require.NotEmpty(t, spaceRequest)
+
+	// don't cleanup the new spacerequest, since we need it for migration testing
+	err = memberAwait.Create(t, spaceRequest)
+	require.NoError(t, err)
+	return spaceRequest
+}
```
MatousJobanek, line 86: could you please move this function to the `setup_runner.go` file? it's very specific to the migration use-case, there is no cleanup activated for the CR, so it's better to keep it there to avoid any confusion

testsupport/space/spacerequest.go
mfrancisc: makes sense, thanks for looking into it! As I mentioned it was just a minor thing! 👍

testsupport/space/spacerequest.go
sadlerap: Fixed in bdb46f7.

MatousJobanek (approved):
Awesome :+1: :1st_place_medal: :rocket: 
Thanks for covering this case in the migration test suite :+1:

MatousJobanek:
/retest
infra

PR #829, opened 2023-10-29
Test update UserSignup status with HomeSpace

baijum:
Test for the change in: https://github.com/codeready-toolchain/host-operator/pull/905


test/e2e/usersignup_test.go
```diff
@@ -45,7 +45,25 @@ func (s *userSignupIntegrationTest) TearDownTest() {
 	memberAwait2.Clean(s.T())
 }
 
-func (s *userSignupIntegrationTest) TestAutomaticApproval() {
+func (s *userSignupIntegrationTest) TestHomeSpaceStatusUpdate() {
```
alexeykazakov, line 48: Same here. Instead of adding a brand new test I would rather update the existing test(s) to check the home space status there.

test/e2e/usersignup_test.go
```diff
@@ -45,7 +45,25 @@ func (s *userSignupIntegrationTest) TearDownTest() {
...
+		EnsureMUR().
+		RequireConditions(wait.ConditionSet(wait.Default(), wait.ApprovedByAdmin())...).
+		Execute(s.T()).Resources()
+
+	// then
+	homeSpaceName := userSignup.Status.HomeSpace
+	require.Equal(s.T(), fmt.Sprintf("%s-2", homeSpace.Name), homeSpaceName)
+}
+
+func (s *userSignupIntegrationTest) NoTestAutomaticApproval() {
```
baijum, line 66: My bad. I will fix this line.

baijum: Corrected.

test/e2e/usersignup_test.go
baijum: Made it as a sub-test. Please check again.

alexeykazakov: I was thinking about adding the home space status check here: https://github.com/codeready-toolchain/toolchain-e2e/blob/master/testsupport/user_assertions.go#L229
This function is used in many tests which create spaces. So there would be no need to add more tests if we added the check to that function.

MatousJobanek (approved):
Looks good. Could you please fix the conflicts?

MatousJobanek:
@baijum Could you please fix the conflicts?

PR #836, opened 2023-11-09
Split WaitForDeployments

rajivnathan:
See https://github.com/codeready-toolchain/toolchain-e2e/pull/822/files#r1361427780 for background


alexeykazakov (approved):

test/migration/verify/verify_migration_test.go
```diff
@@ -24,7 +24,7 @@ import (
 
 func TestAfterMigration(t *testing.T) {
 	// given
-	awaitilities := WaitForDeployments(t)
+	awaitilities := WaitForOperators(t)
```
MatousJobanek, line 27: :thinking: 
this should be called rather in the setup suite when we deploy the operators from master, and not in the "verify" one when the operators are already updated with the modified version

MatousJobanek (commented):
you could also extract the metrics related parts, right? We don't have to have them in the migration setup suite.
Technically also the verification of the proxy route.

test/migration/verify/verify_migration_test.go
alexeykazakov: Good catch. Yeah... we shouldn't verifying webhooks and other "optional" things before we update the operator. And do the full verification of the installations of all components including webhooks after we update the operator to the latest version from the paired PRs.

rajivnathan: I actually thought we wanted to skip these verifications for migration altogether since it's covered by the other tests anyways. So I actually missed switching the WaitForDeployments()  to WaitForOperators() in TestSetupMigration. But I think what you're suggesting would work well too. So if I understand correctly 3d7cde9fb79268d455d5394d4090a819001f9b44 is what you're expecting?

rajivnathan:
>you could also extract the metrics related parts, right? We don't have to have them in the migration setup suite.
Technically also the verification of the proxy route.

@MatousJobanek done in 6c7fde716e7d5bcfc3dabb9642b0c1200e4da34d

MatousJobanek (approved):
Nice :+1: Thanks :tada:

PR #838, opened 2023-11-10
Added migration test

sbryzak:
Related PR: https://github.com/codeready-toolchain/host-operator/pull/928

### DIFF
--- a/test/e2e/usersignup_test.go
+++ b/test/e2e/usersignup_test.go
@@ -47,15 +47,7 @@ func (s *userSignupIntegrationTest) TearDownTest() {
 
 func (s *userSignupIntegrationTest) TestUserSignupMigration() {
 	hostAwait := s.Host()
-	memberAwait := s.Member1()
-	// Create user signup
-	hostAwait.UpdateToolchainConfig(s.T(),
-		testconfig.AutomaticApproval().Enabled(true),
-		testconfig.CapacityThresholds().
-			MaxNumberOfSpaces(
-				testconfig.PerMemberCluster(memberAwait.ClusterName, 1000),
-			).
-			ResourceCapacityThreshold(80))
+	hostAwait.UpdateToolchainConfig(s.T(), testconfig.AutomaticApproval().Enabled(true))
 
 	// Create a new UserSignup
 	userSignup := NewUserSignup(hostAwait.Namespace, "migration-test@acme.com", "migration-test@acme.com")
--- a/test/e2e/usersignup_test.go
+++ b/test/e2e/usersignup_test.go
@@ -45,6 +45,55 @@ func (s *userSignupIntegrationTest) TearDownTest() {
 	memberAwait2.Clean(s.T())
 }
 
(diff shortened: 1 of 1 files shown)


test/e2e/usersignup_test.go
```diff
@@ -45,6 +45,55 @@ func (s *userSignupIntegrationTest) TearDownTest() {
...
+	hostAwait := s.Host()
+	memberAwait := s.Member1()
+	// Create user signup
+	hostAwait.UpdateToolchainConfig(s.T(),
+		testconfig.AutomaticApproval().Enabled(true),
+		testconfig.CapacityThresholds().
+			MaxNumberOfSpaces(
+				testconfig.PerMemberCluster(memberAwait.ClusterName, 1000),
+			).
+			ResourceCapacityThreshold(80))
```
mfrancisc, line 58: just curious - if there was any specific reason for setting those values. As defaults would be `MaxNumberOfSpaces: 0` (which means infinity) and `ResourceCapacity` is already 80 by default. So maybe those are not needed, but it's a minor thing.

test/e2e/usersignup_test.go
```diff
@@ -45,6 +45,55 @@ func (s *userSignupIntegrationTest) TearDownTest() {
...
+				actual.Spec.IdentityClaims.OriginalSub == userSignup.Spec.OriginalSub
+
+		},
+		Diff: func(actual *toolchainv1alpha1.UserSignup) string {
+			return "expected UserSignup to have IdentityClaims migrated"
+		},
+	}
+
+	// Check the UserSignup is successfully migrated now
+	userSignup, err = hostAwait.WaitForUserSignup(s.T(), userSignup.Name, identityClaimsSetCriteria)
```
mfrancisc, line 93: maybe redundand - but should we also verify that the usersignup is approved and the provisioned resources?

something like:
```
	// Check the UserSignup is approved now
	userSignup, err = hostAwait.WaitForUserSignup(s.T(), userSignup.Name, wait.UntilUserSignupHasConditions(wait.ConditionSet(wait.Default(), wait.ApprovedAutomatically())...))
	require.NoError(s.T(), err)

	// Confirm the MUR was created and target cluster was set
	VerifyResourcesProvisionedForSignup(s.T(), s.Awaitilities, userSignup, "deactivate30", "base")
```

mfrancisc (commented):
Nice Job 👍 

I left few minor comments.

test/e2e/usersignup_test.go
sbryzak: Good catch, this was a copy paste from another test and the intention wasn't to set this config to anything out of the ordinary, thanks.

test/e2e/usersignup_test.go
sbryzak: Good point but I think it's not necessary in this case.  If the migration condition test passes then the approval of the usersignup and provisioning are not a concern.

alexeykazakov (approved):

PR #841, opened 2023-11-13
test disabled inheritance for spaces

mmulholla:
replace toolchain-common for testing. https://github.com/codeready-toolchain/toolchain-common/pull/337

Also includes a new test for disabled inheritance with spaces.

### DIFF
--- a/test/e2e/parallel/space_test.go
+++ b/test/e2e/parallel/space_test.go
@@ -478,6 +478,29 @@ func TestSubSpaces(t *testing.T) {
 			})
 		})
 	})
+	t.Run("we create a subSpace, with disable inheritance and do not expect roles and usernames from the parent to be inherited in NSTemplateSet", func(t *testing.T) {
+		// when
+		// we have a parentSpace
+		parentSpace, _, parentSpaceBindings := CreateSpace(t, awaitilities, testspace.WithSpecTargetCluster(memberAwait.ClusterName), testspace.WithTierName("appstudio"))
+		// then
+		// wait until MUR and Space have been provisioned
+		VerifyResourcesProvisionedForSpace(t, awaitilities, parentSpace.Name, UntilSpaceHasStatusTargetCluster(memberAwait.ClusterName))
+		_, err := hostAwait.WaitForMasterUserRecord(t, parentSpaceBindings.Spec.MasterUserRecord)
+		require.NoError(t, err)
+
+		// when
+		// we also have a subSpace with same tier but with disbale inheritance
+		subSpace := CreateSubSpace(t, awaitilities, testspace.WithSpecParentSpace(parentSpace.Name), testspace.WithTierName("appstudio"), testspace.WithSpecTargetCluster(memberAwait.ClusterName), testspace.WithDisableInheritance(true))
+
+		// wait until subSpace has been provisioned
+		_, subSpaceNSTemplateSet := VerifyResourcesProvisionedForSpace(t, awaitilities, subSpace.Name)
+		// check that username and role from parentSpace was not inherited in the subSpace NSTemplateSet
+		_, err = memberAwait.WaitForNSTmplSet(t, subSpaceNSTemplateSet.Name,
+			UntilNSTemplateSetHasConditions(Provisioned()),
+			UntilNSTemplateSetHasNoSpaceRoles(),
+		)
+		require.NoError(t, err)
+	})
 }
 
 func ProvisioningFailed(msg string) toolchainv1alpha1.Condition {
--- a/test/e2e/parallel/space_test.go
+++ b/test/e2e/parallel/space_test.go
@@ -489,7 +489,7 @@ func TestSubSpaces(t *testing.T) {
 		require.NoError(t, err)
 
 		// when
-		// we also have a subSpace with same tier but with disbale inheritance
+		// we also have a subSpace with same tier but with disable inheritance
 		subSpace := CreateSubSpace(t, awaitilities, testspace.WithSpecParentSpace(parentSpace.Name), testspace.WithTierName("appstudio"), testspace.WithSpecTargetCluster(memberAwait.ClusterName), testspace.WithDisableInheritance(true))
 
 		// wait until subSpace has been provisioned
--- a/test/e2e/parallel/space_test.go
+++ b/test/e2e/parallel/space_test.go
@@ -478,6 +478,16 @@ func TestSubSpaces(t *testing.T) {
 			})
 		})
 	})
+}
+
+func TestSubSpaceInheritance(t *testing.T) {
+	// given
+	t.Parallel()
+	// make sure everything is ready before running the actual tests
+	awaitilities := WaitForDeployments(t)
+	hostAwait := awaitilities.Host()
+	memberAwait := awaitilities.Member1()
+
 	t.Run("we create a subSpace, with disable inheritance and do not expect roles and usernames from the parent to be inherited in NSTemplateSet", func(t *testing.T) {
 		// when
 		// we have a parentSpace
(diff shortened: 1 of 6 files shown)


test/e2e/parallel/space_test.go
```diff
@@ -478,6 +478,29 @@ func TestSubSpaces(t *testing.T) {
...
+		VerifyResourcesProvisionedForSpace(t, awaitilities, parentSpace.Name, UntilSpaceHasStatusTargetCluster(memberAwait.ClusterName))
+		_, err := hostAwait.WaitForMasterUserRecord(t, parentSpaceBindings.Spec.MasterUserRecord)
+		require.NoError(t, err)
+
+		// when
+		// we also have a subSpace with same tier but with disable inheritance
+		subSpace := CreateSubSpace(t, awaitilities, testspace.WithSpecParentSpace(parentSpace.Name), testspace.WithTierName("appstudio"), testspace.WithSpecTargetCluster(memberAwait.ClusterName), testspace.WithDisableInheritance(true))
+
+		// wait until subSpace has been provisioned
+		_, subSpaceNSTemplateSet := VerifyResourcesProvisionedForSpace(t, awaitilities, subSpace.Name)
```
mfrancisc, line 496: this function expects some space roles to be present, while the subSpace has none if I'm not wrong. I think this is also the cause of the current test failure.

mfrancisc (commented):
Looks good, nice job!

I have few general questions:
- should we create subSpace with no inheritance using SpaceRequests, so that we verify if the disableInheritance parameter is being propagated correctly from SpaceRequest to Space objects. Or this will be a separate test? 

- should we create bigger hierarchy ( 3 spaces at least )

WDYT?

test/e2e/parallel/space_test.go
```diff
@@ -480,6 +480,39 @@ func TestSubSpaces(t *testing.T) {
...
+		parentSpace, _, parentSpaceBindings := CreateSpace(t, awaitilities, testspace.WithSpecTargetCluster(memberAwait.ClusterName), testspace.WithTierName("appstudio"))
+		// then
+		// wait until MUR and Space have been provisioned
+		VerifyResourcesProvisionedForSpace(t, awaitilities, parentSpace.Name, UntilSpaceHasStatusTargetCluster(memberAwait.ClusterName))
+		mur, err := hostAwait.WaitForMasterUserRecord(t, parentSpaceBindings.Spec.MasterUserRecord)
+		require.NoError(t, err)
+
+		// when
+		// we also have a subSpace with same tier but with disable inheritance
+		subSpace := CreateSubSpace(t, awaitilities, testspace.WithSpecParentSpace(parentSpace.Name), testspace.WithTierName("appstudio"), testspace.WithSpecTargetCluster(memberAwait.ClusterName), testspace.WithDisableInheritance(true))
```
MatousJobanek, line 503: it's just a detail, but this format is easier to read
```suggestion
		subSpace := CreateSubSpace(t, awaitilities, 
			testspace.WithSpecParentSpace(parentSpace.Name), 
			testspace.WithTierName("appstudio"), 
			testspace.WithSpecTargetCluster(memberAwait.ClusterName), 
			testspace.WithDisableInheritance(true))
```

test/e2e/parallel/space_test.go
```diff
@@ -480,6 +480,39 @@ func TestSubSpaces(t *testing.T) {
...
+		testsupportsb.CreateSpaceBinding(t, awaitilities.Host(), mur, subSpace, "viewer")
```
MatousJobanek, line 504: This technically creates a new SpaceBinding for the sub-space and for the same user that has access to the parent space. In other words, the test will pass no matter if the inheritance is disabled or not.

What you could do instead is to provision a brand new user and share the sub-space with this user
```suggestion
		_, secondMur := NewSignupRequest(awaitilities).
			ManuallyApprove().
			EnsureMUR().
			RequireConditions(ConditionSet(Default(), ApprovedByAdmin())...).
			Execute(t).Resources()
		testsupportsb.CreateSpaceBinding(t, awaitilities.Host(), secondMur, subSpace, "contributor")
```

apart from that, there is no "viewer" role in the appstudio NSTemplateTier - let's use "contributor" for example

test/e2e/parallel/space_test.go
```diff
@@ -480,6 +480,39 @@ func TestSubSpaces(t *testing.T) {
...
+
+		// wait until subSpace has been provisioned
+		VerifyResourcesProvisionedForSpace(t, awaitilities, subSpace.Name)
+		// check that username and role from parentSpace was not inherited in the subSpace NSTemplateSet
+		spaceBindings, err := hostAwait.ListSpaceBindings(subSpace.Name)
+		require.NoError(t, err)
+		require.Equal(t,len(spaceBindings),1)
+		require.Equal(t,spaceBindings[0].Spec.SpaceRole, "viewer")
```
MatousJobanek, line 512: the SpaceBinding inheritance is done "virtually" - we don't replicate the SpaceBindings from the parents. What we do is that Space controller collects all SpaceBindings and updates the NSTemplateSet accordingly. That means that the number of the SpaceBinding CRs will be always the same no matter if the inheritance is disabled or not - what is different is the content of the NSTemplateSet.
```suggestion
		appstudioTier, err := hostAwait.WaitForNSTemplateTier(t, "appstudio")
		require.NoError(t, err)
		memberAwait.WaitForNSTmplSet(t, subSpace.Name, UntilNSTemplateSetHasSpaceRoles(
			SpaceRole(appstudioTier.Spec.SpaceRoles["contributor"].TemplateRef, secondMur.Name)))
```

testsupport/wait/member.go
```diff
@@ -608,6 +608,19 @@ func UntilNSTemplateSetHasSpaceRoles(expected ...toolchainv1alpha1.NSTemplateSet
 	}
 }
 
+// UntilNSTemplateSetHasNoSpaceRoles returns a `NSTemplateSetWaitCriterion` which checks that the given
+// NSTemlateSet has no space roles.
+func UntilNSTemplateSetHasNoSpaceRoles() NSTemplateSetWaitCriterion {
```
MatousJobanek, line 613: this doesn't seem to be used

test/e2e/parallel/space_test.go
mmulholla: I have modified the test and pushed my changes - please ignore the extra logs, will remove when all working. I seem to get the same roles in the NStemplate regardless of whether disable inheritance is set or not and not sure how to confirm I am using it. this in the NS template:
```
        spec:
          clusterResources:
            templateRef: appstudio-clusterresources-f08f06a-f08f06a
          namespaces:
          - templateRef: appstudio-tenant-76b9d37-76b9d37
          spaceRoles:
          - templateRef: appstudio-admin-85ef82d-85ef82d
            usernames:
            - ee86b9fd-e565-4128-a
          - templateRef: appstudio-contributor-004f85a-004f85a
            usernames:
            - 4a5388d8-3a96-4d7c-8
          tierName: appstudio
```
was thinking with disable inhertance I would just get contributor.  Also test code assumes only one role so if I add a second subspace I will need to allow for 2.

mmulholla:
> Looks good, nice job!
> 
> I have few general questions:
> 
> * should we create subSpace with no inheritance using SpaceRequests, so that we verify if the disableInheritance parameter is being propagated correctly from SpaceRequest to Space objects. Or this will be a separate test?
> * should we create bigger hierarchy ( 3 spaces at least )
> 
> WDYT?

Space requests test will be later, when I add that support for host operator. Code is done just need to get the sapce code merged first.

I will create a bigger hierarchy once I get tte one subspace case working.

testsupport/wait/member.go
mmulholla: Will delete if not needed but may need it as test evolves as one check is we don't get a parent space role if the sub space has disabled inheritance.

test/e2e/parallel/space_test.go
mmulholla: done

test/e2e/parallel/space_test.go
mmulholla: yes, adding a role for  the sub sapce.

test/e2e/parallel/space_test.go
MatousJobanek: the code seems to be OK, no idea why it keeps inheriting the SpaceBinding from the parent space

MatousJobanek: It looks like that the tests are running fine, so it was maybe using some old version of the toolchain-common dependency.

MatousJobanek (approved):
Looks good :rocket: 
could you please drop the unused function `UntilNSTemplateSetHasNoSpaceRoles` and fix the linter issue? 
Otherwise, feel free to merge the toolchain-common repo and update the dependencies in these PRs :+1: :1st_place_medal:

testsupport/wait/member.go
mmulholla: Removed it

mmulholla:
> Looks good 🚀 could you please drop the unused function `UntilNSTemplateSetHasNoSpaceRoles` and fix the linter issue? Otherwise, feel free to merge the toolchain-common repo and update the dependencies in these PRs 👍 🥇

Have removed the function.

mmulholla:
Updated to use latest toolchain-common from master, so ready for final review.

alexeykazakov (approved):

PR #845, opened 2023-11-16
parallelize nstemplatetier test

MatousJobanek:
* extract one test that requires changes of the global config
* move the rest of the nstemplatetier_test.go into parallel test suite and add `t.Parallel()` to the tests it in


xcoulon (approved):

mfrancisc (approved):
Nice optimization 👍

PR #862, opened 2023-12-06
Refactor MasterUserRecordController to only use propagated claims

sbryzak:
e2e tests for related PR: https://github.com/codeready-toolchain/host-operator/pull/943


rajivnathan (approved):

rajivnathan:
@sbryzak looks like your e2e tests are failing because you have branches named SANDBOX-482 in your host-operator and registration-service so it's causing issues with the branch pairing mechanism for e2e tests. You probably need to rename the registration service one and delete the SANDBOX-482 one.

testsupport/wait/member.go
```diff
@@ -160,13 +160,13 @@ func UntilUserAccountHasAnnotation(key, value string) UserAccountWaitCriterion {
...
 // UntilUserAccountHasSpec returns a `UserAccountWaitCriterion` which checks that the given
-// USerAccount has the expected spec
+// UserAccount has the expected spec
 func UntilUserAccountHasSpec(expected toolchainv1alpha1.UserAccountSpec) UserAccountWaitCriterion {
 	return UserAccountWaitCriterion{
 		Match: func(actual *toolchainv1alpha1.UserAccount) bool {
 			userAccount := actual.DeepCopy()
 			expectedSpec := expected.DeepCopy()
-			return reflect.DeepEqual(userAccount.Spec, *expectedSpec)
+			return reflect.DeepEqual(userAccount.Spec.PropagatedClaims, expectedSpec.PropagatedClaims) && userAccount.Spec.Disabled == expectedSpec.Disabled
```
MatousJobanek, line 169: why couldn't we just compare the UserAccount.Spec? Is it because the fields are still present in CRDs? If yes, wouldn't it be better to drop the fields from the UserAccount CRD since they are not used in useraccount anymore?

test/migration/verify/verify_migration_test.go
```diff
@@ -33,6 +33,12 @@ func TestAfterMigration(t *testing.T) {
 
 	// run all the verify functions concurrently
 	// to ensure that the objects provisioned with the "old" operator versions are not altered in an unexpected way by the new operator version (the changes are backward compatible)
+
+	// TODO These two lines will break the e2e tests! Remove these immediately after they have been merged with master,
+	// along with the associated code in user_assertions.go
+	//SetInMigrationTests(true)
+	//defer SetInMigrationTests(false)
+
```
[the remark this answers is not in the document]
sbryzak: I missed removing this comment sorry, it was part of a hack to try and get the migration tests passing.

testsupport/wait/member.go
sbryzak: This is a tricky one to solve and what I've been stuck working on all through the weekend.  The issue is that the migration tests are comparing the resources produced by the master branch with those in the PR fork.  I have not found a way to successfully compare the specs for `UserAccount` when master is producing values that have both the `UserID` and `OriginalSub` values populated, and the branch has refactored usage of these properties out.  

I settled on simply comparing the `PropagatedClaims` and `Disabled` properties as those are the only relevant properties going forward, and after merging this PR we can raise another PR to change it back to comparing just the spec again.

If I were to drop the properties from the UserAccount CRD, would the `DeepEqual` then pass even if the properties were missing in the object on one side of the comparison?

MatousJobanek: Yeah, I understood the problem you were trying to solve here. If you dropped the fields from the UserAccount CRDs, then the DeepEqual should just pass in the "verify-migration" step as the new CRD (promoted to the cluster) wouldn't contain the fields anymore. The same applies to the API type that use used to get the UserAccount - the type wouldn't contain the fields, so the equal should pass without any problems.

MatousJobanek: Anyway, it should "just work" even with what you wrote here. The proper solution would be dropping the fields from the CRD so you don't need to change the logic of the comparison at all (as we did in the previous cases of the migrations).
That being said, I'm fine with the change if you revert it back to the deepequal as soon as you drop the fields from the CRDs. I didn't see any story for dropping the fields in the epic, so please add the story there and you can also comment to make this comparison logic generic again.

alexeykazakov (approved):

testsupport/wait/member.go
sbryzak: PR that fixed the UserAccount comparison after this PR was merged: https://github.com/codeready-toolchain/toolchain-e2e/pull/865

PR #872, opened 2023-12-31
Updated API to remove all arbitrary SSO claim properties

sbryzak:
Related PRs:

https://github.com/codeready-toolchain/api/pull/393
https://github.com/codeready-toolchain/host-operator/pull/951

(Yes, there is also a member-operator PR with the same branch name.  This will be addressed before this set of PRs is ready for merge).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -122,4 +122,4 @@ require (
 
 go 1.20
 
-replace github.com/codeready-toolchain/api => github.com/sbryzak/api v0.0.0-20231229111412-eada577c9874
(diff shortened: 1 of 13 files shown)


MatousJobanek:
> (Yes, there is also a member-operator PR with the same branch name. This will be addressed before this set of PRs is ready for merge).

I'm sorry, but this cannot be ready for merge in the current state. We have the limitation when we can pair toolchain-e2e PR with only one PR for a good reason. This means, the e2e tests won't be executed for this PR at all which means it won't meet all required checks to be ready for merge.

In other words, this change needs to be properly propagated in each operator in an isolated manner.

PR #873, opened 2024-01-03
use base1ns as default SpaceTier for e2e tests

MatousJobanek:


alexeykazakov (commented):
I wonder if we want to use the default tier instead of hardcoding `base1ns` in the tests...
But let'd do whatever is easier for now.

MatousJobanek:
Yeah, good point. 
For now I just wanted to give it a try, see what it breaks and if it has a positive impact on the time execution, and then discuss it with the team if this is something we want to do.

alexeykazakov:
I think we do want to do it :)
I didn't realize we still using the old tier in most places.

testsupport/user_assertions.go
```diff
@@ -20,11 +21,18 @@ import (
...
 	for _, signup := range signups {
-		VerifyResourcesProvisionedForSignup(t, awaitilities, signup, "deactivate30", "base")
+		VerifyResourcesProvisionedForSignup(t, awaitilities, signup)
 	}
 }
 
-func VerifyResourcesProvisionedForSignup(t *testing.T, awaitilities wait.Awaitilities, signup *toolchainv1alpha1.UserSignup, userTierName, spaceTierName string) {
+func VerifyResourcesProvisionedForSignup(t *testing.T, awaitilities wait.Awaitilities, signup *toolchainv1alpha1.UserSignup) {
+	toolchainConfig := awaitilities.Host().GetToolchainConfig(t)
+	spaceTier := configuration.GetString(toolchainConfig.Spec.Host.Tiers.DefaultSpaceTier, "base")
```
mfrancisc, line 30: should this be ?
```suggestion
	spaceTier := configuration.GetString(toolchainConfig.Spec.Host.Tiers.DefaultSpaceTier, "base1ns")
```

mfrancisc (approved):
Nice 👍 

I have left only few minor comments.

alexeykazakov (approved):

testsupport/user_assertions.go
MatousJobanek: If nothing is set, then the default value is `base`. https://github.com/codeready-toolchain/host-operator/blob/47e819dd07bd3e4888ce6501f254ce062e32e9ec/controllers/toolchainconfig/configuration.go#L312-L314
We should match this behavior in e2e tests as well

mfrancisc: In case nothing is set and default becomes `base` , wouldn't this cause the migration tests to fail now since the `prepareSecondMemberProvisionedSpace` sets `base1ns` as tier for the migration test while the verify migration tests would look for space with `base` tier?

MatousJobanek: It won't have any effect on migration tests - there are two reasons:
* We create the ToolchainConfig (that contains `base1ns`) at the very beginning, before the migration tests, so the Spaces will be provisioned with the `base1ns` tier by default.
* We use the same toolchain-e2e code for both phases of the migration suite, which means it will use this logic when verifying the Spaces (expecting `base1ns` by default)

> while the verify migration tests would look for space with base tier

where do you see that?

mfrancisc: Assuming we don't set the default in the toolchainconfig, what will happen if I'm not mistaken is:

- the setup provisions space using `base1ns` as per this change https://github.com/codeready-toolchain/toolchain-e2e/pull/873/files#diff-666ce05456880cbdaf2945fec4eec4b93c4c0ca759c3f15f12153bf7fba90c5bR72
- the verify migration checks the second member space and uses `base` tier ( assuming again the default was not set to `base1ns` in toolchainconfig) https://github.com/MatousJobanek/toolchain-e2e/blob/93c01493346181fbe638f9344fe627caaf28d275/test/migration/verify/verify_migration_test.go#L168-L172

but maybe I'm not seeing this correctly.

MatousJobanek: but we do set the defaultSpaceTier in ToolchainConfig CR https://github.com/codeready-toolchain/toolchain-e2e/pull/873/files#diff-ce602f9f3fe1f0171275a3497188048716c830039497f7dc5792a88ce271612bR32 so there shouldn't be any problem with it.

Or do you want to change the setup migration step so it provisions the Space with the default tier instead of the hard-coded one?

MatousJobanek: We rely on the specific content of the ToolchainConfig at other places too, so if we make a mistake in configuring the ToolchainConfig manifest then it's rather a bug that should be fixed.

but anyway, I like the idea of having it less hard-coded, so I created a function to get the default tier and use it at the relevant places - see [148d3d8](https://github.com/codeready-toolchain/toolchain-e2e/pull/873/commits/148d3d815a3cc0a44bd2abaee3e451fb21e5706f) and [e8d9ec3](https://github.com/codeready-toolchain/toolchain-e2e/pull/873/commits/e8d9ec330cbfc581d072d31213faf9b4b3ce9267)

PR #874, opened 2024-01-04
Updating the E2e Test "TestIdlerAndPriorityClass"

fbm3307:
Updating the test "TestIdlerAndPriorityClass" to fix some edge case failures.

- To fix limit error for docker "busybox" image pull 
- Make logger a less confusing
- Update the 'notification' variable to not use previous values while testing.

### DIFF
--- a/test/e2e/parallel/user_workloads_test.go
+++ b/test/e2e/parallel/user_workloads_test.go
@@ -71,11 +71,11 @@ func TestIdlerAndPriorityClass(t *testing.T) {
 	require.NoError(t, err)
 	_, err = memberAwait.WaitForPods(t, "workloads-noise", len(externalNsPodsNoise), wait.PodRunning(), wait.WithPodLabel("idler", "idler"), wait.WithOriginalPriorityClass())
 	require.NoError(t, err)
-	_, err = hostAwait.WithRetryOptions(wait.TimeoutOption(10*time.Second)).WaitForNotificationWithName(t, "test-idler-stage-idled", toolchainv1alpha1.NotificationTypeIdled, wait.UntilNotificationHasConditions(wait.Sent()))
-	require.True(t, errors.IsNotFound(err))
 
 	// Check if notification has been deleted before creating another pod
 	err = hostAwait.WaitUntilNotificationWithNameDeleted(t, "test-idler-dev-idled")
+	require.NoError(t, err)
+	err = hostAwait.WaitUntilNotificationWithNameDeleted(t, "test-idler-stage-idled")
 	require.NoError(t, err)
 
 	// Create another pod and make sure it's deleted.
@@ -247,7 +247,7 @@ func podSpec() corev1.PodSpec {
 		TerminationGracePeriodSeconds: &zero,
 		Containers: []corev1.Container{{
 			Name:    "sleep",
-			Image:   "busybox",
+			Image:   "quay.io/prometheus/busybox:latest",
 			Command: []string{"sleep", "36000"}, // 10 hours
 			Resources: corev1.ResourceRequirements{
 				Requests: corev1.ResourceList{
--- a/test/e2e/parallel/user_workloads_test.go
+++ b/test/e2e/parallel/user_workloads_test.go
@@ -14,7 +14,6 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -71,11 +70,11 @@ func TestIdlerAndPriorityClass(t *testing.T) {
 	require.NoError(t, err)
 	_, err = memberAwait.WaitForPods(t, "workloads-noise", len(externalNsPodsNoise), wait.PodRunning(), wait.WithPodLabel("idler", "idler"), wait.WithOriginalPriorityClass())
 	require.NoError(t, err)
-	err = hostAwait.WithRetryOptions(wait.TimeoutOption(10*time.Second)).WaitUntilNotificationWithNameDeletedOrNotCreated(t, "test-idler-stage-idled")
-	require.True(t, errors.IsNotFound(err))
+	err = hostAwait.WithRetryOptions(wait.TimeoutOption(10*time.Second)).WaitForNotificationToBeNotCreated(t, "test-idler-stage-idled")
+	require.NoError(t, err)
 
 	// Check if notification has been deleted before creating another pod
-	err = hostAwait.WaitUntilNotificationWithNameDeletedOrNotCreated(t, "test-idler-dev-idled")
+	err = hostAwait.WaitUntilNotificationWithNameDeleted(t, "test-idler-dev-idled")
 	require.NoError(t, err)
 
 	// Create another pod and make sure it's deleted.
@@ -88,8 +87,8 @@ func TestIdlerAndPriorityClass(t *testing.T) {
 	time.Sleep(time.Duration(2*idler.Spec.TimeoutSeconds) * time.Second)
 	err = memberAwait.WaitUntilPodDeleted(t, pod.Namespace, pod.Name)
 	require.NoError(t, err)
-	err = hostAwait.WithRetryOptions(wait.TimeoutOption(10*time.Second)).WaitUntilNotificationWithNameDeletedOrNotCreated(t, "test-idler-dev-idled")
-	require.True(t, errors.IsNotFound(err))
+	err = hostAwait.WithRetryOptions(wait.TimeoutOption(10*time.Second)).WaitForNotificationToBeNotCreated(t, "test-idler-dev-idled")
+	require.NoError(t, err)
 
 	// There should not be any pods left in the namespace
 	err = memberAwait.WaitUntilPodsDeleted(t, idler.Name, wait.WithPodLabel("idler", "idler"))
--- a/test/e2e/parallel/user_workloads_test.go
+++ b/test/e2e/parallel/user_workloads_test.go
@@ -14,7 +14,6 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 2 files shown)


test/e2e/parallel/user_workloads_test.go
```diff
@@ -71,12 +71,12 @@ func TestIdlerAndPriorityClass(t *testing.T) {
 	require.NoError(t, err)
 	_, err = memberAwait.WaitForPods(t, "workloads-noise", len(externalNsPodsNoise), wait.PodRunning(), wait.WithPodLabel("idler", "idler"), wait.WithOriginalPriorityClass())
 	require.NoError(t, err)
-	_, err = hostAwait.WithRetryOptions(wait.TimeoutOption(10*time.Second)).WaitForNotificationWithName(t, "test-idler-stage-idled", toolchainv1alpha1.NotificationTypeIdled, wait.UntilNotificationHasConditions(wait.Sent()))
-	require.True(t, errors.IsNotFound(err))
 
 	// Check if notification has been deleted before creating another pod
 	err = hostAwait.WaitUntilNotificationWithNameDeleted(t, "test-idler-dev-idled")
 	require.NoError(t, err)
+	err = hostAwait.WaitUntilNotificationWithNameDeleted(t, "test-idler-stage-idled")
```
alexeykazakov, line 78: Are you checking that the notification for the stage idler was deleted or never created in the first place? The old code which you are fixing here is confusing indeed but what is actually supposed to happen here?

fbm3307: It is to check that old notification for the stage idler was deleted

alexeykazakov: But it looks like the stage idler is there just as noise and no notification for it is supposed to be created. If it's not the case then the comments above which say so are misleading.
For example this comment:
```
// make sure that "noise" pods are still there, and notification is not created for stage namespace
```

fbm3307: I agree about the confusion, the question here would be, if its for noise, and expected to Not create any notification.., then why are we waiting for notification to be sent in the test case?
 
 ```
_, err = hostAwait.WithRetryOptions(wait.TimeoutOption(10*time.Second)).WaitForNotificationWithName(t, "test-idler-stage-idled", toolchainv1alpha1.NotificationTypeIdled, wait.UntilNotificationHasConditions(wait.Sent()))
	require.True(t, errors.IsNotFound(err))
```

This Test case will fail, because it will return nil (wont find the notification), as the notification is available just for 5 seconds and then deleted.

Comment (MatousJobanek) [reply test/e2e/parallel/user_workloads_test.go:78]: ok, I missed that "noise" part during our call. 
@alexeykazakov is right - these lines were there for the very same reason as there are the lines:
https://github.com/codeready-toolchain/toolchain-e2e/blob/4a55451961e21fd546ba453a89fab787f07c880b/test/e2e/parallel/user_workloads_test.go#L91-L92

In other words, this is to check that the notification is not created at all - not deleted, but even not created. That's why we waited for 10 seconds to check that the notification is not really created.

Comment (MatousJobanek) [reply test/e2e/parallel/user_workloads_test.go:78]: @fbm3307 what we could do is to improve the the code - there are two ways:
1. create a new function `WaitForNotificationToBeNotCreated` that will wait for 10 seconds and check that the Notification is not created - it will return an error if it is. 
2. Add additional `RetryOption` which could suppress the printed message:
```
hostAwait.WithRetryOptions(wait.TimeoutOption(10*time.Second), wait.SuppressDiff()).WaitForNotificationWithName(t, "test-idler-stage-idled", toolchainv1alpha1.NotificationTypeIdled, wait.UntilNotificationHasConditions(wait.Sent()))
```
so the `printNotificationWaitCriterionDiffs` method doesn't print anything.

TBH, I like the first option more since the function name says what we can expect to be executted

Comment (fbm3307) [reply test/e2e/parallel/user_workloads_test.go:78]: how about i rename/update this function [https://github.com/codeready-toolchain/toolchain-e2e/pull/874/files#diff-2379a187b4778a2330f68b2b1a930fb26d3b8e1b567efe2966e4cb6d8aafe989R1214-R1226](url) instead of writing a new similar function (so that there is no redundancy)
My understanding is that the functionality of `WaitForNotificationToBeNotCreated `would be similar to `WaitUntilNotificationWithNameDeleted`, any thoughts around this..?

Comment (fbm3307) [reply test/e2e/parallel/user_workloads_test.go:78]: and use that function to check that there are no notification created after 10 seconds

Comment (fbm3307) [reply test/e2e/parallel/user_workloads_test.go]: i think writing a new func would be a better option

Comment (alexeykazakov) [reply test/e2e/parallel/user_workloads_test.go]: The difference between waiting for something to be deleted and not created in the first place is the timeout. We would need a longer timeout for something which exist/existed but now we are waiting to be deleted. But we don't want to use the default (long) timeout for waiting for something which is not supposed to exist at all. This is why we usually wait no longer than 10 seconds in such case. So, we have to use different functions for those cases.

Comment (alexeykazakov) [reply test/e2e/parallel/user_workloads_test.go]: If we always wait for the default timeout for something which doesn't exist then every successful test would wait for that long timeout. This is why a shorter timeout (10 seconds) is preferable.

Comment (fbm3307) [reply test/e2e/parallel/user_workloads_test.go]: Thanks for explanation. I got a better Understanding now .

File: testsupport/wait/host.go [hunk 2]
```diff
@@ -1186,13 +1187,28 @@ func (a *HostAwaitility) WaitForNotificationWithName(t *testing.T, notificationN
...
+
+// WaitForNotificationToBeNotCreated waits and checks that notification is NOT created.
+func (a *HostAwaitility) WaitForNotificationToBeNotCreated(t *testing.T, notificationName string) error {
+	t.Logf("waiting to check notification with name '%s' is NOT created", notificationName)
+	return wait.Poll(a.RetryInterval, 10*time.Second, func() (done bool, err error) {
+		notification := &toolchainv1alpha1.Notification{}
+		if err := a.Client.Get(context.TODO(), types.NamespacedName{Name: notificationName, Namespace: a.Namespace}, notification); err != nil {
+			if errors.IsNotFound(err) {
+				return true, nil
+			}
```
MatousJobanek, line 1207: I'm afraid that this won't wait for 10s, but will exit at the first run when it doesn't find anything.
The expected behavior is that the poll will be running for the whole 10s and only if it finds the notification that it should immediately exit. In other words, you need to swap the returned values and the expected outcome (if you want to stay with wait.Pool).

fbm3307: Yes right, Thanks for pointing that. i updated it now

testsupport/wait/host.go
```diff
@@ -1175,9 +1175,10 @@ func (a *HostAwaitility) WaitForNotifications(t *testing.T, username, notificati
 // WaitForNotificationWithName waits until there is an expected Notifications available with the provided name and with the notification type and which match the conditions (if provided).
 func (a *HostAwaitility) WaitForNotificationWithName(t *testing.T, notificationName, notificationType string, criteria ...NotificationWaitCriterion) (toolchainv1alpha1.Notification, error) {
 	t.Logf("waiting for notification with name '%s'", notificationName)
-	var notification toolchainv1alpha1.Notification
+	notification := &toolchainv1alpha1.Notification{}
 	err := wait.Poll(a.RetryInterval, a.Timeout, func() (done bool, err error) {
-		if err := a.Client.Get(context.TODO(), types.NamespacedName{Name: notificationName, Namespace: a.Namespace}, &notification); err != nil {
+		notification = &toolchainv1alpha1.Notification{}
+		if err := a.Client.Get(context.TODO(), types.NamespacedName{Name: notificationName, Namespace: a.Namespace}, notification); err != nil {
 			return false, err
```
MatousJobanek, line 1182: I think that there is one more thing we missed in this case
```suggestion
			if errors.IsNotFound(err) {
				return false, nil
			}
			return false, err
```

testsupport/wait/host.go
```diff
@@ -1186,13 +1187,26 @@ func (a *HostAwaitility) WaitForNotificationWithName(t *testing.T, notificationN
...
+func (a *HostAwaitility) WaitForNotificationToNotBeCreated(t *testing.T, notificationName string) error {
+	t.Logf("waiting to check notification with name '%s' is NOT created", notificationName)
+	err := wait.Poll(a.RetryInterval, 10*time.Second, func() (done bool, err error) {
+		notification := &toolchainv1alpha1.Notification{}
+		if err := a.Client.Get(context.TODO(), types.NamespacedName{Name: notificationName, Namespace: a.Namespace}, notification); err != nil {
+			return false, err
+		}
+		return true, err
+	})
+	return err
```
MatousJobanek, line 1209: this won't keep waiting for 10s again - it will technically return immediately. 
I had something like this in mind:
```suggestion
	t.Logf("waiting to check notification with name '%s' is NOT created", notificationName)
	notification := &toolchainv1alpha1.Notification{}
	err := wait.Poll(a.RetryInterval, 10*time.Second, func() (done bool, err error) {
		notification = &toolchainv1alpha1.Notification{}
		if err := a.Client.Get(context.TODO(), types.NamespacedName{Name: notificationName, Namespace: a.Namespace}, notification); err != nil {
			if errors.IsNotFound(err) {
				return false, nil
			}
			return false, err
		}
		return true, nil
	})
	if err == nil {
		return fmt.Errorf("notification '%s' was found, but it was expected to not be created/present: \n %v", notificationName, notification)
	}
	if err == wait.ErrWaitTimeout {
		return nil
	}
	return err
```

and then you can call it:
```
err = hostAwait.WaitForNotificationToNotBeCreated(t, "test-idler-stage-idled")
require.NoError(t, err)
```

so it's technically very similar to the other function `WaitForNotificationWithName`, we just change the expectations

fbm3307: yes right i had tried this before but somehow ended up doing the other way round, updated it now

test/e2e/parallel/user_workloads_test.go
```diff
@@ -88,7 +88,7 @@ func TestIdlerAndPriorityClass(t *testing.T) {
 	time.Sleep(time.Duration(2*idler.Spec.TimeoutSeconds) * time.Second)
 	err = memberAwait.WaitUntilPodDeleted(t, pod.Namespace, pod.Name)
 	require.NoError(t, err)
-	_, err = hostAwait.WithRetryOptions(wait.TimeoutOption(10*time.Second)).WaitForNotificationWithName(t, "test-idler-dev-idled", toolchainv1alpha1.NotificationTypeIdled, wait.UntilNotificationHasConditions(wait.Sent()))
+	err = hostAwait.WithRetryOptions(wait.TimeoutOption(10*time.Second)).WaitForNotificationToNotBeCreated(t, "test-idler-dev-idled")
```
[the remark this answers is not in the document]
fbm3307: yes , updated it now

MatousJobanek (approved):
Looks good :+1: 
Thanks for the additional changes :1st_place_medal:

PR #891, opened 2024-02-01
KSPACE-37: Create the SpaceProvisionerConfig objects for all member clusters in the e2e tests

metlos:
Create the SpaceProvisionerConfig objects for all member clusters in the e2e tests.

### DIFF
--- a/make/resources/default-spaceprovisionerconfig.yaml
+++ b/make/resources/default-spaceprovisionerconfig.yaml
@@ -1,0 +1,17 @@
+apiVersion: template.openshift.io/v1
+kind: Template
+metadata:
+  name: default-network-policies
+objects:
+  - apiVersion: toolchain.dev.openshift.com/v1alpha1
+    kind: SpaceProvisionerConfig
+    metadata:
+      name: ${SPACEPROVISIONERCONFIG_NAME}
+      namespace: ${SPACEPROVISIONERCONFIG_NS}
+    spec:
+      toolchainCluster: ${TOOLCHAINCLUSTER_NAME}
+      enabled: true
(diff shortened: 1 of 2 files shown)


make/resources/default-spaceprovisionerconfig.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: template.openshift.io/v1
+kind: Template
+metadata:
+  name: default-network-policies
```
alexeykazakov, line 4: Copy-paste I guess.

alexeykazakov (approved):
Looks good. Just need to fix the template name :)

make/test.mk
```diff
@@ -350,6 +350,12 @@ ifneq ($(E2E_TEST_EXECUTION),true)
 	oc delete pods --namespace ${HOST_NS} -l name=registration-service || true
 endif
 
+.PHONY: create-spaceprovisionerconfigs-for-members
+create-spaceprovisionerconfigs-for-members:
+	for member_name in `oc get toolchaincluster -l type=member -n ${HOST_NS} --no-headers -o custom-columns=":metadata.name"`; do \
```
MatousJobanek, line 355: it's just a detail, but we want to drop the type label, so you could do it here right away.
```suggestion
	for member_name in `oc get toolchaincluster -l -n ${HOST_NS} --no-headers -o custom-columns=":metadata.name"`; do \
```
@fbm3307 please don't forget to check also the makefile targets for any occurrences of the usage of the `type` label

MatousJobanek (approved):
Cool :+1:

PR #895, opened 2024-02-20
Copy the SA token of sub-spaces only for specific tiers

mfrancisc:
Verify that we can provision SpaceRequest with tier that has no secret associated with the namespaces.

Paired with: https://github.com/codeready-toolchain/host-operator/pull/972

Jira: https://issues.redhat.com/browse/ASC-402


dperaza4dustbit (changes requested):
/lgtm

testsupport/wait/member.go
```diff
@@ -376,6 +376,25 @@ func UntilSpaceRequestHasNamespaceAccess(subSpace *toolchainv1alpha1.Space) Spac
 	}
 }
 
+// UntilSpaceRequestHasNamespaceAccessWithoutSecretRef returns a `SpaceRequestWaitCriterion` which checks that the given
+// SpaceRequest has `status.NamespaceAccess[*].SecretRef` empty, since generation of a secret is not expected for the given SpaceRequest.
+func UntilSpaceRequestHasNamespaceAccessWithoutSecretRef() SpaceRequestWaitCriterion {
+	return SpaceRequestWaitCriterion{
+		Match: func(actual *toolchainv1alpha1.SpaceRequest) bool {
+			for _, nsAccess := range actual.Status.NamespaceAccess {
```
rajivnathan, line 384: Should we check the length of `actual.Status.NamespaceAccess` to make sure it's not empty? Otherwise, it could lead to a flaky test in case NamespaceAccess is empty?

mfrancisc: We can add a check here, but it's already checked by https://github.com/mfrancisc/toolchain-e2e/blob/5ac2da4d54d6dcdef17f17e24be9357c678ed88f/test/e2e/parallel/spacerequest_test.go#L277 which ensures that SpaceRequest and SubSpace have the same NamespaceAccess list. Before that we check that the SubSpace NamespaceAccess list is not empty https://github.com/mfrancisc/toolchain-e2e/blob/5ac2da4d54d6dcdef17f17e24be9357c678ed88f/test/e2e/parallel/spacerequest_test.go#L270.

mfrancisc: I have added a check for that in https://github.com/codeready-toolchain/toolchain-e2e/pull/895/commits/539273f479cd819b19228f2f5fbfd56ea173567e.

Please let me know if this looks better now.

MatousJobanek (approved):
Thanks

testsupport/wait/member.go
rajivnathan: Ah, I missed that initially. 👍

rajivnathan (approved):

PR #902, opened 2024-02-27
Updates to setup tool README

rajivnathan:


alexeykazakov (approved):

ranakan19 (approved):

setup/README.adoc
```diff
@@ -154,10 +154,10 @@ make clean-e2e-resources
 
 1. Install operators
 ```
-go run setup/main.go --users 1 --default 1 --custom 0 --username setup
+go run setup/main.go --users 1 --default 1 --custom 0 --username baseline --testname=baseline
 ```
 
 2. Run setup for 2000 users
 ```
-go run setup/main.go --users 2000 --default 2000 --custom 0 --username pizza
```
ranakan19, line 162: no more pizza :(

setup/README.adoc
```diff
@@ -36,7 +36,7 @@ networking:
     hostPrefix: 23
   machineNetwork:
   - cidr: 10.0.0.0/16
-  networkType: OpenShiftSDN
+  networkType: OVNKubernetes
```
mfrancisc, line 39: just curious - why this change ? I'm still using `OpenShiftSDN` for my "local" development. Did something changed in ocp 4.15 ?

mfrancisc (approved):

setup/README.adoc
mfrancisc: Ha got it! Thanks for the heads up. I still need to switch to 4.15 locally so that's why I haven't faced this issue.

mfrancisc: Also last times I've used ClusterBot to generate Rosa cluster, which turned out to be quicker to provision and more "stable" in kind of installation process.

MatousJobanek: :thinking: but our clusters are still running OpenShiftSDN, right? Or did they started migrating the already existing clusters to OVNKubernetes? IIRC, the migration was no-go for the already existing OSD clusters some time ago, so I'm curious if this has changed as well or not.

rajivnathan: Yeah, my understanding is that migration to OVNKubernetes is still not supported for OSD clusters. At least according to the 4.15 docs: https://docs.openshift.com/container-platform/4.15/networking/ovn_kubernetes_network_provider/migrate-from-openshift-sdn.html

We should check with the OSD team though. I mentioned it to @alexeykazakov already.

PR #906, opened 2024-03-01
print setting KUBECONFIG env var for local debug

MatousJobanek:
so it's easy to debug tests in IDE


make/test.mk
```diff
@@ -166,6 +166,7 @@ print-local-debug-info:
 	@echo "oc logs -l name=registration-service --namespace ${REGISTRATION_SERVICE_NS} --all-containers=true --prefix=true --tail=-1"
 	@echo ""
 	@echo "Add the following lines at the very beginning of the test/suite that you want to run/debug from your IDE:"
+	@echo 'os.Setenv("KUBECONFIG","$(or ${KUBECONFIG}, ${HOME}/.kube/config)")'
```
rajivnathan, line 169: Did you mean `"$(${KUBECONFIG} or ${HOME}/.kube/config)"`  or am I missing some command line magic?

MatousJobanek: it's a makefile magic :magic_wand: 
https://www.gnu.org/software/make/manual/html_node/Conditional-Functions.html#index-or
I didn't know about that before either :-)

alexeykazakov (approved):

make/test.mk
rajivnathan: Ah cool, learned something new. Thanks

rajivnathan (approved):

PR #916, opened 2024-03-12
SANDBOX-555: clean up packages being imported more than once

rsoaresd:
# Description

While taking a look at `nstemplatetier_test.go`, I noticed that the wait package was being imported more than once. So, I ran `staticcheck ./...` locally to check if it occurs in other files. This PR cleans up the packages being imported more than once. In the future, maybe we can wonder if it makes sense to add this check to golangci lint config.

[SANDBOX-555](https://issues.redhat.com/browse/SANDBOX-555)


MatousJobanek (approved):
Cool, thanks :+1: :1st_place_medal: :rocket:

rajivnathan (approved):
Nice, thanks for cleaning this up 🙂

PR #917, opened 2024-03-12
SANDBOX-555: move TestToolchainClusterE2E to the parallel tests folder

rsoaresd:
# Description

Move `TestToolchainClusterE2E` to the parallel tests folder since this test can be run in parallel.

[SANDBOX-555](https://issues.redhat.com/browse/SANDBOX-555)


alexeykazakov (approved):
On the first glance it seems to be safe to move it to the parallel folder :+1: 

Thanks!

rsoaresd:
[alexeykazakov](https://github.com/alexeykazakov), the security hotspot is related to the hardcoded IP in this [line](https://github.com/codeready-toolchain/toolchain-e2e/pull/917/files#diff-201af516f70afb21a0742fab4c780b094807010d8b82e0ea40e1aa9e871cfcd9R81). How can be marked as safe or ignored? I am guessing that I do not have permission for that. Can you help me, pls?

MatousJobanek (changes requested):
This cannot be moved to the parallel test suite, at least not now. 
In the first sub-test we create a new ToolchainCluster CR which is supposed to be ready, so it could be technically available for provisioning new accounts, but this shouldn't happen.

MatousJobanek:
there is actually one more thing - the tests are called not only for member type, but also for host. Having an extra member would cause provisioning to a wrong member cluster, having an extra host could cause that the member operator would talk to a wrong host. 
So there could be more cases when it could fail. 
That being said, let's not move it to the parallel test suite, it could cause weird flakiness in the other tests.

rsoaresd:
> there is actually one more thing - the tests are called not only for member type, but also for host. Having an extra member would cause provisioning to a wrong member cluster, having an extra host could cause that the member operator would talk to a wrong host. So there could be more cases when it could fail. That being said, let's not move it to the parallel test suite, it could cause weird flakiness in the other tests.

So sorry! Thank you so much, @MatousJobanek! I will close this PR.

MatousJobanek:
@rsoaresd there is nothing to be sorry about, really. This is pretty hidden side-effect that is not apparent for the first glance. 
We really appreciate your contribution :pray:

PR #924, opened 2024-03-14
KSPACE-46: Use SpaceProvisionerConfig for space assignment

metlos:
This just changes the existing tests to configure the `SpaceProvisionerConfig`s instead of the `ToolchainConfig` for assignment of spaces to clusters.

Related PRs:
* host-operator: https://github.com/codeready-toolchain/host-operator/pull/991
* api: https://github.com/codeready-toolchain/api/pull/407
* toolchain-common: https://github.com/codeready-toolchain/toolchain-common/pull/374


mfrancisc (approved):
Nice Job 👍

testsupport/spaceprovisionerconfig/spaceprovisionerconfig.go
```diff
@@ -20,3 +23,59 @@ func CreateSpaceProvisionerConfig(t *testing.T, await *wait.Awaitility, opts ...
...
+	t.Helper()
+	spcs, err := getAllSpcs(t, await)
+	require.NoError(t, err)
+
+	idx := findIndexOfFirstSpaceProvisionerConfigReferencingCluster(spcs, referencedClusterName)
+	require.GreaterOrEqual(t, idx, 0, "could not find SpaceProvisionerConfig referencing the required cluster: %s", referencedClusterName)
+
+	spc := &spcs[idx]
+
+	originalSpc := spc.DeepCopy()
```
MatousJobanek, line 37: if you rename and modify the function `findIndexOfFirstSpaceProvisionerConfigReferencingCluster` to:
```go
func findSPCFor(spcs []toolchainv1alpha1.SpaceProvisionerConfig, clusterName string) (*toolchainv1alpha1.SpaceProvisionerConfig, error) {
	for _, spc := range spcs {
		if spc.Spec.ToolchainCluster == clusterName {
			return &spc, nil
		}
	}
	return nil, fmt.Errorf("could not find SpaceProvisionerConfig referencing the required cluster: %s", clusterName)
}
```
then you could call it here: 
```suggestion
	spc, err := findSPCFor(spcs, referencedClusterName)
	require.NoError(t, err)

	originalSpc := spc.DeepCopy()
```
which seems to be a bit easier to read (at least for me)

test/e2e/space_autocompletion_test.go
```diff
@@ -148,16 +134,13 @@ func TestAutomaticClusterAssignment(t *testing.T) {
...
-			var memberLimits []testconfig.PerMemberClusterOptionInt
 			for _, m := range toolchainStatus.Status.Members {
 				if memberAwait1.ClusterName == m.ClusterName {
-					memberLimits = append(memberLimits, testconfig.PerMemberCluster(memberAwait1.ClusterName, m.SpaceCount))
+					spaceprovisionerconfig.UpdateForCluster(t, hostAwait.Awaitility, memberAwait1.ClusterName, testSpc.MaxNumberOfSpaces(uint(m.SpaceCount)), testSpc.Enabled(true))
 				} else if memberAwait2.ClusterName == m.ClusterName {
-					memberLimits = append(memberLimits, testconfig.PerMemberCluster(memberAwait2.ClusterName, m.SpaceCount))
+					spaceprovisionerconfig.UpdateForCluster(t, hostAwait.Awaitility, memberAwait2.ClusterName, testSpc.MaxNumberOfSpaces(uint(m.SpaceCount)), testSpc.Enabled(true))
 				}
 			}
```
MatousJobanek, line 143: this could be probably simplified to 
```
for _, m := range toolchainStatus.Status.Members {
	spaceprovisionerconfig.UpdateForCluster(t, hostAwait.Awaitility, m.ClusterName, testSpc.MaxNumberOfSpaces(uint(m.SpaceCount)), testSpc.Enabled(true))
}
```

test/e2e/space_autocompletion_test.go
```diff
@@ -124,21 +110,21 @@ func TestAutomaticClusterAssignment(t *testing.T) {
...
 			} else if memberAwait2.ClusterName == m.ClusterName {
-				memberLimits = append(memberLimits, testconfig.PerMemberCluster(memberAwait2.ClusterName, m.SpaceCount+1))
+				spaceprovisionerconfig.UpdateForCluster(t, hostAwait.Awaitility, memberAwait2.ClusterName, testSpc.MaxNumberOfSpaces(uint(m.SpaceCount+1)), testSpc.Enabled(true))
+				configured += 1
 			}
 		}
-		require.Len(t, memberLimits, 2)
-
-		hostAwait.UpdateToolchainConfig(t, testconfig.CapacityThresholds().MaxNumberOfSpaces(memberLimits...))
+		require.Equal(t, 2, configured)
```
MatousJobanek, line 127: I don't think that we need this - not sure why it was there before

test/e2e/space_autocompletion_test.go
```diff
@@ -100,9 +86,8 @@ func TestAutomaticClusterAssignment(t *testing.T) {
 
 	t.Run("set low capacity threshold and expect that space will have default tier, but won't have target cluster so it won't be provisioned", func(t *testing.T) {
 		// given
-		hostAwait.UpdateToolchainConfig(t,
-			testconfig.CapacityThresholds().ResourceCapacityThreshold(1),
-		)
+		spaceprovisionerconfig.UpdateForCluster(t, hostAwait.Awaitility, memberAwait1.ClusterName, testSpc.MaxMemoryUtilizationPercent(1), testSpc.Enabled(true))
+		spaceprovisionerconfig.UpdateForCluster(t, hostAwait.Awaitility, memberAwait2.ClusterName, testSpc.MaxMemoryUtilizationPercent(1), testSpc.Enabled(true))
```
MatousJobanek, line 90: one thing that I noticed is that you explicitly define  `testSpc.Enabled(true)` everywhere in `space_autocompletion_test.go` even though it should be already set in https://github.com/codeready-toolchain/toolchain-e2e/blob/66d668ac8863788c0d240bfe1dd8de43013ca77d/deploy/host-operator/default-spaceprovisionerconfig.yaml#L13

For example the  `usersignup_test.go` doesn't have it - you don't explicitly define it there

test/metrics/metrics_test.go
```diff
@@ -187,6 +189,8 @@ func TestMetricsWhenUsersAutomaticallyApprovedAndThenDeactivated(t *testing.T) {
 	hostAwait := awaitilities.Host()
 	memberAwait := awaitilities.Member1()
 	memberAwait2 := awaitilities.Member2()
+	spaceprovisionerconfig.UpdateForCluster(t, hostAwait.Awaitility, memberAwait.ClusterName, testSpc.Enabled(true), testSpc.MaxNumberOfSpaces(1))
+	spaceprovisionerconfig.UpdateForCluster(t, hostAwait.Awaitility, memberAwait2.ClusterName, testSpc.Enabled(true), testSpc.MaxNumberOfSpaces(1))
```
MatousJobanek, line 193: why do we need it here? We don't rely on the limits in this test, do we? :upside_down_face:

MatousJobanek (commented):
Looks great :1st_place_medal: 
Similarly as in the host-operator PR, only a few cosmetic suggestions

test/e2e/space_autocompletion_test.go
metlos: it's because the tests reuse the same client and disable/enable the SPCs back and forth. So I'm being explicit about the enabled states so that there is no confusion about the state of the SPCs.

metlos: I know it should not be needed now that we have proper cleanup of the SPCs but I still find it provides for better clarity.

MatousJobanek (approved):
Awesome, great job with the e2e tests :clap:

PR #933, opened 2024-03-25
SANDBOX-554 tests for setting ScheduledDeactivationTime on UserSignup

sbryzak:
Tests for https://github.com/codeready-toolchain/host-operator/pull/988


testsupport/wait/host.go
```diff
@@ -580,6 +580,52 @@ func UntilUserSignupHasStateLabel(expected string) UserSignupWaitCriterion {
...
+				found := false
+				for _, s := range actual.Spec.States {
+					if s == requiredState {
+						found = true
+						break
+					}
+				}
+				if !found {
+					return false
+				}
```
rsoaresd, line 596:
```suggestion
                                if !slices.Contains(actual.Spec.States, requiredState) {
                                    return false
                                }
```

rsoaresd (commented):
Hello @sbryzak 👋 Just a minor suggestion:

alexeykazakov (approved):
Looks good. Apart from @rsoaresd's suggestion.

testsupport/wait/host.go
sbryzak: Nice suggestion, thanks!

test/e2e/user_management_test.go
```diff
@@ -256,6 +256,9 @@ func (s *userManagementTestSuite) TestUserDeactivation() {
 
 		// The non-excluded user should be deactivated
 		err = hostAwait.WaitUntilMasterUserRecordAndSpaceBindingsDeleted(t, murMember1.Name)
+		bytes, err2 := wait.StringifyObject(userSignupMember1)
+		require.NoError(t, err2)
+		t.Logf("### DEBUG usersignup: %s", string(bytes))
```
MatousJobanek, line 261: most likely some left-over from troubleshooting?

testsupport/wait/host.go
```diff
@@ -580,6 +580,52 @@ func UntilUserSignupHasStateLabel(expected string) UserSignupWaitCriterion {
...
+			return fmt.Sprintf("expected UserSignup '%s' to contain states '%s', actual states: '%s'", actual.Name,
+				states, actual.Spec.States)
+		},
+	}
+}
+
+func UntilUserSignupHasScheduledDeactivationTime() UserSignupWaitCriterion {
+	return UserSignupWaitCriterion{
+		Match: func(actual *toolchainv1alpha1.UserSignup) bool {
+			return actual.Status.ScheduledDeactivationTimestamp != nil
```
MatousJobanek, line 610: Could we verify also the content of the timestamp, so we know that it's properly calculated and set?

MatousJobanek (commented):
looks good, but it would be nice to update also the generic assertion function expecting the value to be always set for all provisioned UserSignup CRs.
https://github.com/codeready-toolchain/toolchain-e2e/blob/4485a0ff1b7490fb3a0a6a2ba3a55a71c3c08c17/testsupport/user_assertions.go#L64-L106

test/e2e/user_management_test.go
sbryzak: not quite left over, i'm actively trying to diagnose the cause of the current test failures.

sbryzak:
@MatousJobanek i think that updating the generic `VerifyUserRelatedResources` function might be a bad idea, not all usersignups will have the scheduled deactivation time set, e.g. signups that are on an excluded email domain.

testsupport/wait/host.go
sbryzak: Currently I'm only using this function to ensure the state of the UserSignup, i.e. the deactivation controller has run and set a value for the scheduled deactivation time property.  In the test where I'm calling it I'm already checking the actual value is as expected.

MatousJobanek: I don't see that for all cases. Where do you verify the timestamp for this particular case? 
https://github.com/sbryzak/toolchain-e2e/blob/1257d0941845405bbe0c584c05184d0795684640/test/e2e/user_management_test.go#L352-L355

PR #934, opened 2024-03-26
test: use more fine-grained network policies for host and member operators

xcoulon:
based on https://github.com/codeready-toolchain/sandbox-sre/blob/master/resources/setup/default-network-policies.yaml

### DIFF
--- a/testsupport/wait/awaitility.go
+++ b/testsupport/wait/awaitility.go
@@ -41,7 +41,7 @@ import (
 
 const (
 	DefaultRetryInterval             = time.Millisecond * 100 // make it short because a "retry interval" is waited before the first test
-	DefaultTimeout                   = time.Second * 120
+	DefaultTimeout                   = time.Second * 240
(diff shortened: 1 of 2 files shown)


testsupport/wait/awaitility.go
```diff
@@ -41,7 +41,7 @@ import (
 
 const (
 	DefaultRetryInterval             = time.Millisecond * 100 // make it short because a "retry interval" is waited before the first test
-	DefaultTimeout                   = time.Second * 120
+	DefaultTimeout                   = time.Second * 240
```
alexeykazakov, line 44: Any particular reason for increasing the default timeout?

xcoulon: that was not intended to remain in the PR. 
I got some weird issues where a test would timeout while waiting for a space to be provisioned, and when I checked the space a few minutes later, it would be actually provisioned 🤷‍♂️ 

Anyways, I will revert this change.

PR #939, opened 2024-03-29
test: use more fine-grained network policies for host operator

xcoulon:
based on https://github.com/codeready-toolchain/sandbox-sre/blob/master/resources/setup/default-network-policies.yaml
so that e2e tests run with network policies more similar to staging and prod.

Done by dropping the `create-project` goal and moving its contents back into the
existing ` create-host-project`, `create-member1-project` and `create-member2-project`
goals. Only a couple of lines are duplicated, but having yet another goal seems a bit overkill.

Also, rename the `create-member1` and `create-member2` goals to
`create-member1-project` and `create-member2-project` for consistency
with the sibbling `create-host-project` goal.


mfrancisc (commented):
Looks good 👍 

I have one question related to `default-network-policies.yaml` , is this still used/needed at this point ?

xcoulon:
> Looks good 👍
> 
> I have one question related to `default-network-policies.yaml` , is this still used/needed at this point ?

ah, good point, I forgot to delete the file when working on the PR :/

xcoulon:
> > Looks good 👍
> > I have one question related to `default-network-policies.yaml` , is this still used/needed at this point ?
> 
> ah, good point, I forgot to delete the file when working on the PR :/

wait, `deploy/default-network-policies.yaml` was renamed to `deploy/member-operator-network-policies.yaml`, so now I'm confused: which `default-network-policies.yaml` were you referring to?

mfrancisc:
> deploy/default-network-policies.yaml was renamed to deploy/member-operator-network-policies.yaml

my bad, I haven't noticed the renaming change 🤦‍♂️ . So that's only valid for member namespace now, I see. Thanks for pointing that out and sorry for the confusion.

mfrancisc (approved):

xcoulon:
> > deploy/default-network-policies.yaml was renamed to deploy/member-operator-network-policies.yaml
> 
> my bad, I haven't noticed the renaming change 🤦‍♂️ . So that's only valid for member namespace now, I see. Thanks for pointing that out and sorry for the confusion.

No worries @mfrancisc! I also forgot about this renaming when reading your first comment 😄  
I guess we were both a bit tired 😂

make/test.mk
```diff
@@ -302,20 +302,26 @@ endif
...
-.PHONY: create-member1
-create-member1:
+.PHONY: create-member1-project
+create-member1-project:
 	@echo "Preparing namespace for member operator: $(MEMBER_NS)..."
-	$(MAKE) create-project PROJECT_NAME=${MEMBER_NS}
+	@-oc new-project ${MEMBER_NS} 1>/dev/null
+	@-oc project ${MEMBER_NS}
+	@echo "adding network policies in $(MEMBER_NS) namespace"
+	@-oc process -p NAMESPACE=$(MEMBER_NS) -f deploy/member-operator-network-policies.yaml | oc apply -f -
```
MatousJobanek, line 313: it's just a detail - this network policy allows all ingress, so technically dropping the network policy would have the same effect.
This network policy was introduced because you wanted to combine toolchain-e2e setup with the old legacy sandbox-sre setup via cli (which created some network policies at that time). This use-case is not valid anymore, so I guess that we could drop it, right?

MatousJobanek: after all, we don't have anything like that in sandbox-sre, so dropping `member-operator-network-policies.yaml` would make it closer to production usage.

xcoulon: > it's just a detail - this network policy allows all ingress, so technically dropping the network policy would have the same effect.

right, we probably don't need it on OpenShift CI on OCP, indeed

> after all, we don't have anything like that in sandbox-sre, so dropping `member-operator-network-policies.yaml` would make it closer to production usage.

I'm confused, we have https://github.com/codeready-toolchain/sandbox-sre/blob/master/resources/setup/default-network-policies.yaml today and it's applied on prod, isn't it?

MatousJobanek: yes, but only in the host namespace, we don't create any network policies in member namespaces at all.

MatousJobanek: I'm talking about dropping the `member-operator-network-policies.yaml` only, not `host-operator-network-policies.yaml`

xcoulon: I wonder: if we need to set this `allow-all-ingresses` policy on staging and prod, then is there any benefit in setting it here too, so we have something similar/consistent? or is it overkill and we should just get rid of them? (I'm fine with both options, to be honest)

rajivnathan (approved):
As @MatousJobanek mentioned, I think deploy/member-operator-network-policies.yaml can be deleted too. Otherwise, looks good 👍

make/test.mk
xcoulon: done in 4aed699

make/test.mk
```diff
@@ -302,20 +302,24 @@ endif
...
 
-.PHONY: create-member1
-create-member1:
+.PHONY: create-member1-project
+create-member1-project:
 	@echo "Preparing namespace for member operator: $(MEMBER_NS)..."
-	$(MAKE) create-project PROJECT_NAME=${MEMBER_NS}
+	@-oc new-project ${MEMBER_NS} 1>/dev/null
+	@-oc project ${MEMBER_NS}
+	@echo "adding network policies in $(MEMBER_NS) namespace"
```
MatousJobanek, line 312:
```suggestion
```

make/test.mk
```diff
@@ -302,20 +302,24 @@ endif
...
-.PHONY: create-member2
-create-member2:
+.PHONY: create-member2-project
+create-member2-project:
 ifeq ($(SECOND_MEMBER_MODE),true)
 	@echo "Preparing namespace for second member operator: ${MEMBER_NS_2}..."
-	$(MAKE) create-project PROJECT_NAME=${MEMBER_NS_2}
+	@-oc new-project ${MEMBER_NS_2} 1>/dev/null
+	@-oc project ${MEMBER_NS_2}
+	@echo "adding network policies in $(MEMBER_NS_2) namespace"
```
MatousJobanek, line 322:
```suggestion
```

make/test.mk
xcoulon: ah damn'it 🤦‍♂️

make/test.mk
MatousJobanek: cool :+1: 
btw, are you going to try to figure out network policies for members too - I mean as a long-term plan?

xcoulon: > cool 👍 btw, are you going to try to figure out network policies for members too - I mean as a long-term plan?

yes, let's track it in https://issues.redhat.com/browse/SANDBOX-585

PR #958, opened 2024-05-02
Remove old RHODS install from setup tool

rajivnathan:


alexeykazakov (approved):
What about adding the new one? I guess it will be a separate PR?

PR #962, opened 2024-05-03
Core: Generate xUnit Reports for e2e tests 

sawood14012:
a Wip PR which enables generation of junit/xunit reports to connect report-portal and Quality dashboard


sawood14012:
> Thanks for taking a look at this 🙏
> 
> I'm not familiar with `go-junit-report` but don't you need to pipe it to the `go test ./...` command as [explained here](https://github.com/jstemmer/go-junit-report?tab=readme-ov-file#usage):
> 
> ```
> go test -v 2>&1 ./... | go-junit-report -set-exit-code > report.xml
> ```

it gets piped if you see my generate-report command and execute-tests cmd

make/test.mk
```diff
@@ -175,6 +179,25 @@ else
...
+	@command -v go-junit-report >/dev/null 2>&1 || { echo "go-junit-report is not installed. Installing..."; go install github.com/jstemmer/go-junit-report/v2@latest; }
+	@echo "go-junit-report version:" && go-junit-report -version
+
+.PHONY: generate-report
+generate-report:
+	@echo "Generating report"
+ifneq ($(OPENSHIFT_BUILD_NAMESPACE),) 
+	mkdir -p  ${ARTIFACT_DIR}/${REPORT_PORTAL_DIR}
+	PATH=${PATH}:$(go env GOPATH)/bin
+	go-junit-report > ${ARTIFACT_DIR}/${REPORT_PORTAL_DIR}/${REPORT_NAME}
```
mfrancisc, line 195: my bad, maybe this picks up the updated `PATH`
```suggestion
	PATH=${PATH}:$(go env GOPATH)/bin go-junit-report > ${ARTIFACT_DIR}/${REPORT_PORTAL_DIR}/${REPORT_NAME}
```

make/test.mk
```diff
@@ -175,6 +177,25 @@ else
...
+# 	@command -v go-junit-report >/dev/null 2>&1 || { echo "go-junit-report is not installed. Installing..."; go install github.com/jstemmer/go-junit-report/v2@latest; }
+# 	@echo "go-junit-report version:" && go-junit-report -version
+
+.PHONY: generate-report
+generate-report:
+	@echo "Generating report"
+ifneq ($(OPENSHIFT_BUILD_NAMESPACE),) 
+	mkdir -p  ${ARTIFACT_DIR}/${REPORT_PORTAL_DIR}
+	go install github.com/jstemmer/go-junit-report/v2@latest
+	PATH=${PATH}:$(go env GOPATH)/bin go-junit-report > ${ARTIFACT_DIR}/${REPORT_PORTAL_DIR}/${REPORT_NAME}
```
mfrancisc, line 193:
```suggestion
	GOBIN=$(PWD) go install github.com/jstemmer/go-junit-report/v2@latest
	./go-junit-report > ${ARTIFACT_DIR}/${REPORT_PORTAL_DIR}/${REPORT_NAME}
```

make/test.mk
```diff
@@ -163,7 +165,8 @@ execute-tests:
 	@echo "Status of ToolchainStatus"
 	-oc get ToolchainStatus -n ${HOST_NS} -o yaml
 	@echo "Starting test $(shell date)"
-	MEMBER_NS=${MEMBER_NS} MEMBER_NS_2=${MEMBER_NS_2} HOST_NS=${HOST_NS} REGISTRATION_SERVICE_NS=${REGISTRATION_SERVICE_NS} go test ${TESTS_TO_EXECUTE} -run ${TESTS_RUN_FILTER_REGEXP} -p 1 -parallel ${E2E_PARALLELISM} -v -timeout=90m -failfast || \
+	$(MAKE) check-go-junit-report
```
MatousJobanek, line 168: this should be executed only for tests running in openshift-ci

make/test.mk
```diff
@@ -181,6 +184,23 @@ else
 	$(MAKE) print-local-debug-info  HOST_NS=${HOST_NS} MEMBER_NS=${MEMBER_NS} MEMBER_NS_2=${MEMBER_NS_2} REGISTRATION_SERVICE_NS=${REGISTRATION_SERVICE_NS}
 endif
 
+.PHONY: check-go-junit-report
+
+check-go-junit-report:
+	@command -v ./go-junit-report >/dev/null 2>&1 || { echo "go-junit-report is not installed. Installing..."; GOBIN=$(PWD) go install github.com/jstemmer/go-junit-report/v2@latest; }
+	@echo "go-junit-report version:" && ./go-junit-report -version
```
MatousJobanek, line 191: I would rather avoid installing the binary in the folder of the repository - let's install it either in /tmp folder or somewhere else.

make/test.mk
```diff
@@ -163,7 +165,8 @@ execute-tests:
 	@echo "Status of ToolchainStatus"
 	-oc get ToolchainStatus -n ${HOST_NS} -o yaml
 	@echo "Starting test $(shell date)"
-	MEMBER_NS=${MEMBER_NS} MEMBER_NS_2=${MEMBER_NS_2} HOST_NS=${HOST_NS} REGISTRATION_SERVICE_NS=${REGISTRATION_SERVICE_NS} go test ${TESTS_TO_EXECUTE} -run ${TESTS_RUN_FILTER_REGEXP} -p 1 -parallel ${E2E_PARALLELISM} -v -timeout=90m -failfast || \
+	$(MAKE) check-go-junit-report
+	MEMBER_NS=${MEMBER_NS} MEMBER_NS_2=${MEMBER_NS_2} HOST_NS=${HOST_NS} REGISTRATION_SERVICE_NS=${REGISTRATION_SERVICE_NS} go test ${TESTS_TO_EXECUTE} -run ${TESTS_RUN_FILTER_REGEXP} -p 1 -parallel ${E2E_PARALLELISM} -v -timeout=90m -failfast 2>&1 | $(MAKE) generate-report REPORT_NAME=${REPORT_NAME}  || \
```
MatousJobanek, line 169: you cannot use the pipe for the whole output; otherwise, we will lose the content/output of the tests when running it locally as well as in openshift-ci

sawood14012:
@MatousJobanek i have resolved your comments and updated my approach 

1. This approach ensures that the full output of the tests is retained and can be used both locally and in OpenShift CI. 
2. Also changed the dir of binary to /tmp
3. the binary doesn't execute or generate a report locally.

make/test.mk
```diff
@@ -163,7 +166,14 @@ execute-tests:
...
-	MEMBER_NS=${MEMBER_NS} MEMBER_NS_2=${MEMBER_NS_2} HOST_NS=${HOST_NS} REGISTRATION_SERVICE_NS=${REGISTRATION_SERVICE_NS} go test ${TESTS_TO_EXECUTE} -run ${TESTS_RUN_FILTER_REGEXP} -p 1 -parallel ${E2E_PARALLELISM} -v -timeout=90m -failfast || \
+ifneq ($(OPENSHIFT_BUILD_NAMESPACE),) 
+	$(MAKE) check-go-junit-report
+else 
+	@echo "Skipping Go Junit report check and install"
+endif
+	@echo "Running tests"
+	MEMBER_NS=${MEMBER_NS} MEMBER_NS_2=${MEMBER_NS_2} HOST_NS=${HOST_NS} REGISTRATION_SERVICE_NS=${REGISTRATION_SERVICE_NS} go test ${TESTS_TO_EXECUTE} -run ${TESTS_RUN_FILTER_REGEXP} -p 1 -parallel ${E2E_PARALLELISM} -v -timeout=90m -failfast 2>&1 | tee test_output.log
+	$(MAKE) generate-report REPORT_NAME=${REPORT_NAME} || \
 	($(MAKE) print-logs HOST_NS=${HOST_NS} MEMBER_NS=${MEMBER_NS} MEMBER_NS_2=${MEMBER_NS_2} REGISTRATION_SERVICE_NS=${REGISTRATION_SERVICE_NS} && exit 1)
```
MatousJobanek, line 177: this is executed as a separate command, which means that printing logs will depend on the result of the `generate-report` target, not on the result of the tests. 
The original way it was written printed out the logs only when the test failed and also returned the exit code `1`, so execution of the `print-logs` target depended on the results of the tests - this needs to stay the same.

sawood14012: so i am now running the generate report cmd after print logs hope this works

make/test.mk
```diff
@@ -163,8 +166,15 @@ execute-tests:
...
-	($(MAKE) print-logs HOST_NS=${HOST_NS} MEMBER_NS=${MEMBER_NS} MEMBER_NS_2=${MEMBER_NS_2} REGISTRATION_SERVICE_NS=${REGISTRATION_SERVICE_NS} && exit 1)
+ifneq ($(OPENSHIFT_BUILD_NAMESPACE),) 
+	$(MAKE) check-go-junit-report
+else 
+	@echo "Skipping Go Junit report check and install"
+endif
+	@echo "Running tests"
+	MEMBER_NS=${MEMBER_NS} MEMBER_NS_2=${MEMBER_NS_2} HOST_NS=${HOST_NS} REGISTRATION_SERVICE_NS=${REGISTRATION_SERVICE_NS} \
+	go test ${TESTS_TO_EXECUTE} -run ${TESTS_RUN_FILTER_REGEXP} -p 1 -parallel ${E2E_PARALLELISM} -v -timeout=90m -failfast 2>&1 | tee /tmp/test_output.log || ( $(MAKE) print-logs && exit 1 )
+	$(MAKE) generate-report REPORT_NAME=${REPORT_NAME}
```
MatousJobanek, line 177: if the test fails, then the previous line exits with exit code `1`, so the report won't be generated - see 
```
|| ( $(MAKE) print-logs && exit 1 )
```

MatousJobanek: I guess that you need to do something like this:
```suggestion
	MEMBER_NS=${MEMBER_NS} MEMBER_NS_2=${MEMBER_NS_2} HOST_NS=${HOST_NS} REGISTRATION_SERVICE_NS=${REGISTRATION_SERVICE_NS} \
	(go test ${TESTS_TO_EXECUTE} -run ${TESTS_RUN_FILTER_REGEXP} -p 1 -parallel ${E2E_PARALLELISM} -v -timeout=90m -failfast 2>&1 | tee /tmp/test_output.log && $(MAKE) generate-report REPORT_NAME=${REPORT_NAME}) || ( $(MAKE) print-logs && $(MAKE) generate-report REPORT_NAME=${REPORT_NAME} && exit 1 )
```
but I haven't tested it. 
Not sure if there is some simpler way to deal with it

make/test.mk
```diff
@@ -163,9 +166,17 @@ execute-tests:
...
+	$(MAKE) check-go-junit-report
+else 
+	@echo "Skipping Go Junit report check and install"
+endif
+	@echo "Running tests"
+	MEMBER_NS=${MEMBER_NS} MEMBER_NS_2=${MEMBER_NS_2} HOST_NS=${HOST_NS} REGISTRATION_SERVICE_NS=${REGISTRATION_SERVICE_NS} \
+	go test ${TESTS_TO_EXECUTE} -run ${TESTS_RUN_FILTER_REGEXP} -p 1 -parallel ${E2E_PARALLELISM} -v -timeout=90m -failfast 2>&1 | tee /tmp/test_output.log; \
+	STATUS=$$?; \
+	$(MAKE) generate-report REPORT_NAME=${REPORT_NAME}; \
+	if [ $$STATUS -ne 0 ]; then $(MAKE) print-logs && exit 1; fi
```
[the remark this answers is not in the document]
MatousJobanek: my guess is that `STATUS=$$?` takes the exit value from the `tee`  command that is executed "after" the test as part of the pipe

make/test.mk
MatousJobanek: this won't work because of the pipe and the returned value from `tee` either :-/

make/test.mk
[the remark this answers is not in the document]
rsoaresd: I was thinking of changing to `STATUS=${PIPESTATUS[0]}; \` to capture the exit status of `go test`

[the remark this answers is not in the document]
MatousJobanek: that could work, I didn't know about `PIPESTATUS` :+1:

MatousJobanek:
closing as it's not needed anymore

PR #970, opened 2024-05-10
KUBESAW-16: Label token secrets

metlos:
This adds a simple test to check that all toolchain clusters are labeled using the new `toolchain.dev.openshift.com/toolchain-cluster` label.

Associated PRs:
- toolchain-common: https://github.com/codeready-toolchain/toolchain-common/pull/396
- api: https://github.com/codeready-toolchain/api/pull/423
- host-operator: https://github.com/codeready-toolchain/host-operator/pull/1030
- toolchain-cicd: https://github.com/codeready-toolchain/toolchain-cicd/pull/109


mfrancisc (approved):
Nice job 👍 

An alternative approach could be to list all secrets with the `ToolchainClusterLabel` label and check that there is only 1 secret.

test/e2e/toolchaincluster_test.go
```diff
@@ -24,17 +24,36 @@ func TestToolchainClusterE2E(t *testing.T) {
...
+func verifyToolchainCluster(t *testing.T, await *wait.Awaitility, otherAwait *wait.Awaitility, shouldHaveReferencedSecret bool) {
 	// given
 	current, ok, err := await.GetToolchainCluster(t, otherAwait.Namespace, nil)
 	require.NoError(t, err)
 	require.True(t, ok, "ToolchainCluster should exist")
+
+	// NOTE: this needs to run first, before the sub-tests below, because they reuse the secret for the new
+	// toolchain clusters. This is technically incorrect but sufficient for those tests.
+	// Note that we are going to be changing the workflow such that the label on the secret will actually be the driver
+	// for ToolchainCluster creation and so re-using the secret for different TCs will become impossible in the future.
```
rajivnathan, line 45: I don't quite understand. If reusing the secret for different TCs will be impossible in the future then are the other tests still going to be valid? If they're going to be invalid then we can just remove them now?

rajivnathan (approved):

test/e2e/toolchaincluster_test.go
rajivnathan: Ah, is it supposed to mean the subtests will be updated in a following PR to create the secret with label instead of the ToolchainCluster directly?

metlos: yep. I didn't want to rewrite the tests just yet, because they are still working.

PR #977, opened 2024-05-16
add tests to check that SpaceProvisionerConfig is only ready when the underlying ToolchainCluster is ready

metlos:
Add tests to check that SpaceProvisionerConfig is only ready when the underlying ToolchainCluster is ready.

Related PRs:
- host-operator: https://github.com/codeready-toolchain/host-operator/pull/1033
- api: https://github.com/codeready-toolchain/api/pull/424

I just want to put this out for others to take a look. I'm not sure if the approach to testing the capacity manager is correct and therefore keeping it as a draft for the time being.

### DIFF
--- a/test/e2e/parallel/spaceprovisionerconfig_test.go
+++ b/test/e2e/parallel/spaceprovisionerconfig_test.go
@@ -145,6 +145,7 @@ func TestSpaceProvisionerConfig(t *testing.T) {
 		require.NoError(t, err)
 
 		// when
+		require.NoError(t, host.Client.Get(context.TODO(), client.ObjectKeyFromObject(cluster), cluster))
 		cluster.Spec.SecretRef.Name = ""
 		require.NoError(t, host.Client.Update(context.TODO(), cluster))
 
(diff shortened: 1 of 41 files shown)


test/e2e/parallel/spaceprovisionerconfig_test.go
```diff
@@ -110,4 +102,73 @@ func TestSpaceProvisionerConfig(t *testing.T) {
...
+	// note that we can't use the CopyWithCleanup function because we also
+	// need to modify the SecretRef of the TC.
+	newCluster := cluster.DeepCopy()
+	newCluster.ResourceVersion = ""
+	newCluster.UID = ""
+	newCluster.Name = newName
+	newCluster.Spec.SecretRef.Name = secret.Name
+	require.NoError(t, a.CreateWithCleanup(t, newCluster))
+
+	return newCluster
```
MatousJobanek, line 173: this looks very similar to what we have in test/e2e/toolchaincluster_test.go file - do you think that we could unify the code in the testsupport/ package (or somewhere there)?

metlos: Good point. Most of this could be changed to use `wait.CopyWithCleanup()`...

test/e2e/parallel/spaceprovisionerconfig_test.go
```diff
@@ -51,52 +54,60 @@ func TestSpaceProvisionerConfig(t *testing.T) {
...
+				return
+			}
+			tc.Spec.SecretRef.Name = newSecretName
+			err = host.Client.Update(context.TODO(), tc)
+			if err != nil {
+				return
+			}
+			done = true
+			return
+		}))
```
MatousJobanek, line 90: you could use `host.UpdateToolchainCluster()` instead, right?

MatousJobanek (approved):
Just one comment; otherwise, it looks great :+1: :1st_place_medal:

PR #980, opened 2024-05-17
fix flaky: copy ToolchainCluster secret

MatousJobanek:
We need to copy the secret that is referenced by the ToolchainCluster CR. 
What happens now is that the migration code in ToolchainCluster controller updates the secret with the expected label `toolchain.dev.openshift.com/toolchain-cluster` set to the name of the ToolchainCluster CR, but since the same secret is used by two ToolchainClusters then there is a huge number of immediate updates of the same secret in a row when the controller is changing the label from one value to another. This increase of updates can cause instability of the controller/operator that we can see in openshift-ci around the time when this particular test is running.

**EDIT**: Taking a second look at the ToolchainCluster controller, the flow described above is probably not valid. The ToolchainCluster controller doesn't watch the secrets, so the updates of the label to a different value shouldn't trigger new reconcile. But let's try this, copying the secret (and not reusing the existing one) is a good thing anyway.

**EDIT 2**: Ok, it wasn't the cause because I just faced the same error in this PR as well https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/codeready-toolchain_toolchain-e2e/980/pull-ci-codeready-toolchain-toolchain-e2e-master-e2e/1791354153084129280

**EDIT 3**: As discussed in this thread https://redhat-internal.slack.com/archives/CHK0J6HT6/p1715939245467529 I found a root cause of the problem. The test is adding a prefix to the already existing names of the ToolchainClusters, but in openshift-ci the name is already long, so with the prefix it exceeds the max number of chars (63) we can set at labels. Fix: [c57651a](https://github.com/codeready-toolchain/toolchain-e2e/pull/980/commits/c57651a9ba48f52d6519b3463967aad1929ad32a)

### DIFF
--- a/test/e2e/toolchaincluster_test.go
+++ b/test/e2e/toolchaincluster_test.go
@@ -8,11 +8,11 @@ import (
 	"time"
 
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
+	"github.com/codeready-toolchain/toolchain-common/pkg/test"
 	. "github.com/codeready-toolchain/toolchain-e2e/testsupport"
 	"github.com/codeready-toolchain/toolchain-e2e/testsupport/wait"
 	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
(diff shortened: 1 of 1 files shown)


test/e2e/toolchaincluster_test.go
```diff
@@ -150,6 +144,31 @@ func verifyToolchainCluster(t *testing.T, await *wait.Awaitility, otherAwait *wa
 	})
 }
 
+func generateNewName(prefix, baseName string) string {
+	if len(prefix+baseName) >= 63 {
+		return prefix + baseName[len(prefix):]
```
mfrancisc, line 149: shouldn't this be something like 
```suggestion
		return prefix + baseName[len(prefix):63]
```

mfrancisc: I mean, removing the length of the prefix doesn't ensure that the name is < 63 chars right.

MatousJobanek: ok, my assumption was that the name was already within the boundary

MatousJobanek: I wanted to drop the prefix of the original name only, to keep the uniqueness of the names. But it's maybe not so important since we add some specific prefix anyway. Let me simplify it

MatousJobanek: simplified b7c4c23e

mfrancisc (approved):
Nice catch 🏅

PR #981, opened 2024-05-17
fix flaky: create Space & SpaceBinding always in a loop

MatousJobanek:
https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/codeready-toolchain_member-operator/562/pull-ci-codeready-toolchain-member-operator-master-e2e/1790274999433564160

Space auto-completion test was creating Space & SpaceBinding CRs using a function that wasn't executed in a loop. The test is not running in a parallel test suite (so it shouldn't be theoretically a big deal), but there was apparently some reason for not having the SpaceBinding present and thus the NSTemplateSet didn't have the SpaceRoles set and was being terminated.
So, to avoid any unexpected scenarios, I'm changing it so it uses the loop that ensures that Space & SpaceBinding is always created together.
As part of this investigation I noticed that one parallel test case was using the same function - I fixed that too.

### DIFF
--- a/test/e2e/parallel/space_test.go
+++ b/test/e2e/parallel/space_test.go
@@ -121,38 +121,26 @@ func TestSpaceRoles(t *testing.T) {
 	appstudioTier, err := hostAwait.WaitForNSTemplateTier(t, "appstudio")
 	require.NoError(t, err)
 
-	// given a user (with her own space, but we'll ignore it in this test)
-	_, ownerMUR := NewSignupRequest(awaitilities).
-		Username("spaceowner").
-		Email("spaceowner@redhat.com").
-		ManuallyApprove().
-		TargetCluster(awaitilities.Member1()).
-		EnsureMUR().
-		RequireConditions(ConditionSet(Default(), ApprovedByAdmin())...).
-		NoSpace().
-		Execute(t).
-		Resources()
-
-	// when a space owned by the user above is created (and user is an admin of this space)
-	s, ownerBinding := CreateSpaceWithBinding(t, awaitilities, ownerMUR,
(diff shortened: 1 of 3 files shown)


rajivnathan (approved):

MatousJobanek:
@rsoaresd correct me if I'm wrong, but both test-cases I'm trying to fix in this PR are covered in your other PRs  https://github.com/codeready-toolchain/toolchain-e2e/pull/985 https://github.com/codeready-toolchain/toolchain-e2e/pull/990 so we can technically close mine since your PRs are fixing them more properly

rsoaresd:
> @rsoaresd correct me if I'm wrong, but both test-cases I'm trying to fix in this PR are covered in your other PRs #985 #990 so we can technically close mine since your PRs are fixing them more properly

@MatousJobanek , the test `TestAutomaticClusterAssignment` is not covered on my PRs since it requires space specifications like specifying the space name (which is not possible through `NewSignupRequest`), as discussed on the slack thread. I think we can keep on your PR but delete changes on file `test/e2e/parallel/space_test.go`. WDYT?

MatousJobanek:
@rsoaresd gotcha, I'll wait for your PRs and then solve the conflicts on my side

MatousJobanek:
@rsoaresd I merged the changes and resolved the conflicts. Could you please take a look when you have time? :-)

rsoaresd:
> @rsoaresd I merged the changes and resolved the conflicts. Could you please take a look when you have time? :-)

Sure! Goinggg

rsoaresd (approved):

mfrancisc (approved):
Nice improvement 👍

PR #983, opened 2024-05-27
test: SANDBOX-629: refactor TestPromoteSpace

rsoaresd:
# Description
When we create standalone spaces, we sporadically face flakiness since we currently do not make sure that the SpaceBinding is created along with Space. So, when we create SpaceBinding in a loop with space, we sporadically face flakiness because the space binding controller does not find the space at time - because the space is deleted by the space controller since it did not find the SpaceBinding at time- and for that reason it will delete the SpaceBinding and it will be not found.

To avoid it, we can remove unnecessary standalone space creation, which is the case of the `TestPromoteSpace` test.

[Job example ](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/codeready-toolchain_host-operator/1034/pull-ci-codeready-toolchain-host-operator-master-e2e/1792943569225388032#1:build-log.txt%3A14260):
```
    host.go:2284: failed to find SpaceBinding
        
    space.go:56: 
        	Error Trace:	/tmp/toolchain-e2e/testsupport/space/space.go:56
        	            				/tmp/toolchain-e2e/testsupport/space/space.go:25
        	            				/tmp/toolchain-e2e/test/e2e/parallel/space_test.go:256
        	Error:      	Received unexpected error:
        	            	timed out waiting for the condition
        	Test:       	TestPromoteSpace

```
## Issue ticket number and link
[SANDBOX-629](https://issues.redhat.com/browse/SANDBOX-629)

PR #985, opened 2024-05-27
test: SANDBOX-629: refactor TestSpaceRoles and TestSubSpaces

rsoaresd:
# Description
Remove unnecessary standalone space creation. For more info, check PR #983 or [SANDBOX-629](https://issues.redhat.com/browse/SANDBOX-629).

## Issue ticket number and link
[SANDBOX-629](https://issues.redhat.com/browse/SANDBOX-629)

PR #984, opened 2024-05-27
test: SANDBOX-629: refactor TestCreateSpaceBindingRequest and TestUpdateSpaceBindingRequest

rsoaresd:
# Description
Remove unnecessary standalone space creation. For more info, check PR #983 or [SANDBOX-629](https://issues.redhat.com/browse/SANDBOX-629).

## Issue ticket number and link
[SANDBOX-629](https://issues.redhat.com/browse/SANDBOX-629)

PR #986, opened 2024-05-28
test: SANDBOX-629: refactor CreateSpaceRequest

rsoaresd:
# Description
Remove unnecessary standalone space creation. For more info, check PR #983 or [SANDBOX-629](https://issues.redhat.com/browse/SANDBOX-629).

[SANDBOX-629](https://issues.redhat.com/browse/SANDBOX-629)

### DIFF
--- a/testsupport/signup_request.go
+++ b/testsupport/signup_request.go
@@ -367,10 +367,11 @@ func Close(t *testing.T, resp *http.Response) {
 }
 
 // CreateUserSignupWithSpaceTier creates a UserSignup with space tier specified
-func CreateUserSignupWithSpaceTier(t *testing.T, awaitilities wait.Awaitilities, tier string) (*toolchainv1alpha1.UserSignup, *toolchainv1alpha1.MasterUserRecord) {
+func CreateUserSignupWithSpaceTier(t *testing.T, awaitilities wait.Awaitilities, targetCluster *wait.MemberAwaitility, tier string) (*toolchainv1alpha1.UserSignup, *toolchainv1alpha1.MasterUserRecord) {
 	userSignup, mur := NewSignupRequest(awaitilities).
 		ManuallyApprove().
 		RequireConditions(wait.ConditionSet(wait.Default(), wait.ApprovedByAdmin())...).
+		TargetCluster(targetCluster).
 		SpaceTier(tier).
 		EnsureMUR().
 		Execute(t).
--- a/testsupport/signup_request.go
+++ b/testsupport/signup_request.go
@@ -18,7 +18,6 @@ import (
 	"github.com/codeready-toolchain/toolchain-e2e/testsupport/tiers"
 	"github.com/codeready-toolchain/toolchain-e2e/testsupport/wait"
 
-	u "github.com/gofrs/uuid"
 	"github.com/google/uuid"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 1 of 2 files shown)


MatousJobanek (approved):
Nice :+1:

testsupport/space/spacerequest.go
```diff
@@ -55,10 +55,14 @@ func WithName(name string) SpaceRequestOption {
 }
 
 func CreateSpaceRequest(t *testing.T, awaitilities wait.Awaitilities, memberName string, opts ...SpaceRequestOption) (*toolchainv1alpha1.SpaceRequest, *toolchainv1alpha1.Space) {
+	hostAwait := awaitilities.Host()
 	memberAwait, err := awaitilities.Member(memberName)
 	require.NoError(t, err)
 	// let's first create a parentSpace
-	parentSpace, _, _ := CreateSpace(t, awaitilities, testspace.WithTierName("appstudio"), testspace.WithSpecTargetCluster(memberAwait.ClusterName))
+	_, mur := testsupport.CreateUserSignupWithSpaceTier(t, awaitilities, "appstudio")
```
MatousJobanek, line 62: one thing I just realized - the original code provisioned the space in the given member cluster, the rest of the tests relies on this fact, but the new code would pick any of the member clusters that has enough capacity.
This should be changed so the space is always provisioned to the given member cluster.

MatousJobanek (changes requested):

testsupport/signup_request.go
```diff
@@ -353,3 +365,16 @@ func Close(t *testing.T, resp *http.Response) {
...
+func CreateUserSignupWithSpaceTier(t *testing.T, awaitilities wait.Awaitilities, tier string) (*toolchainv1alpha1.UserSignup, *toolchainv1alpha1.MasterUserRecord) {
+	userSignup, mur := NewSignupRequest(awaitilities).
+		ManuallyApprove().
+		RequireConditions(wait.ConditionSet(wait.Default(), wait.ApprovedByAdmin())...).
+		SpaceTier(tier).
+		EnsureMUR().
+		Execute(t).
+		Resources()
+
+	return userSignup, mur
```
MatousJobanek, line 379: let's also wait for the space and return it from the function so there is some added value in the function and it doesn't only abstract the fluent-api.
```go
	space, err := awaitilities.Host().WaitForSpace(t, mur.Name)
	require.NoError(t, err)
	return userSignup, mur, space
```

MatousJobanek: or you could update the SignupRequest API so the `Resources()` function returns also the Space object (if provisioned)

MatousJobanek: Updating returned values of the `Resources()` function of the SignupRequest API would cause many changes across the whole e2e project, so if you decide to do that, then let's do it in a separate PR.

rsoaresd: I am more inclined to the first option. WDYT?

testsupport/space/spacerequest.go
rsoaresd: Good catch!! My bad, sorry! I will address it!

testsupport/signup_request.go
rsoaresd: I think I will go with option 2, after all 😄 

Regarding CreateUserSignupWithSpaceTier only abstract the fluent-api, I create this function to avoid duplicated code, since we will need it in other tests like PR #984, PR #985, and PR #990

MatousJobanek: yeah, option 2 is a bit more work, but it will help us with cleaning up some of the duplicated code (retrieval of the Spaces). So I'm for it :+1: 

> Regarding CreateUserSignupWithSpaceTier only abstract the fluent-api, I create this function to avoid duplicated code

I understand that there can be some duplication, but this is something (with the number of test-cases) we just cannot avoid. Also, I have a feeling that the number of helper functions is growing and it's sometimes hard to follow which functions does what and why, so if it was there only to abstract the fluent API, then I would rather drop it and go with the small duplication of the same. In addition to that, it's sometimes better to call the fluent API with the exact values so it's clear what is set and used - like the target member cluster the user is gonna be provisioned to.

alexeykazakov (approved):
Thanks!

mfrancisc (approved):
Nice job 👍

MatousJobanek (approved):

PR #989, opened 2024-05-29
fix flakiness: reset invalid SpaceRole to valid one in SBR

MatousJobanek:
[SANDBOX-580](https://issues.redhat.com/browse/SANDBOX-580)
https://redhat-internal.slack.com/archives/CHK0J6HT6/p1716912003579899

* reset the SpaceRole to a valid one so it can be removed quickly
* drop the test case with an invalid MUR - we cannot update MUR (webhook blocks that) and it could cause flakiness in the future.

### DIFF
--- a/test/e2e/parallel/spacebindingrequest_test.go
+++ b/test/e2e/parallel/spacebindingrequest_test.go
@@ -10,6 +10,7 @@ import (
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
 	"github.com/gofrs/uuid"
 	"github.com/stretchr/testify/assert"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	spacebindingrequesttestcommon "github.com/codeready-toolchain/toolchain-common/pkg/test/spacebindingrequest"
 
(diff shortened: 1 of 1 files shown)


test/e2e/parallel/spacebindingrequest_test.go
```diff
@@ -123,36 +122,26 @@ func TestCreateSpaceBindingRequest(t *testing.T) {
...
+			_, err = memberAwait.WaitForSpaceBindingRequest(t, client.ObjectKeyFromObject(spaceBindingRequest),
 				UntilSpaceBindingRequestHasConditions(spacebindingrequesttestcommon.UnableToCreateSpaceBinding(fmt.Sprintf("invalid role 'invalid' for space '%s'", space.Name))),
 			)
 			require.NoError(t, err)
 			bindings, err := hostAwait.ListSpaceBindings(space.Name)
 			require.NoError(t, err)
 			assert.Len(t, bindings, 1)
-		})
 
-		t.Run("unable create space binding request with invalid MasterUserRecord", func(t *testing.T) {
```
mfrancisc, line 135: I'm not 100% sure I understand why dropping this test case. This is not updating the MUR but creating an SBR with an invalid MUR, which is not covered by the WebHook.

mfrancisc (commented):
Thanks for looking into this 🙏 

I'm just not sure we should delete the test case.

mfrancisc (approved):
Thanks for the fix!

test/e2e/parallel/spacebindingrequest_test.go
MatousJobanek: already discussed in https://redhat-internal.slack.com/archives/CHK0J6HT6/p1716974045574919?thread_ts=1716912003.579899&cid=CHK0J6HT6
We cannot reset the MUR back to a valid value (webhook blocks that) so we would end up with an invalid SBR so we could face the same flakiness again. Let's rely on unit-tests in this particular situation

PR #990, opened 2024-05-29
test: SANDBOX-629: refactor TestSubSpaceInheritance

rsoaresd:
# Description
Remove unnecessary standalone space creation. For more info, check PR #983 or [SANDBOX-629](https://issues.redhat.com/browse/SANDBOX-629).

[SANDBOX-629](https://issues.redhat.com/browse/SANDBOX-629)

### DIFF
--- a/test/e2e/parallel/space_test.go
+++ b/test/e2e/parallel/space_test.go
@@ -513,7 +513,15 @@ func TestSubSpaceInheritance(t *testing.T) {
 			SpaceTier("appstudio").
 			EnsureMUR().
 			Execute(t)
-		mur := parentUser.MUR
+
+		user := NewSignupRequest(awaitilities).
+			ManuallyApprove().
+			TargetCluster(memberAwait).
+			RequireConditions(ConditionSet(Default(), ApprovedByAdmin())...).
+			SpaceTier("appstudio").
+			EnsureMUR().
+			Execute(t)
+		mur := user.MUR
 
 		// when
 		// we also have a subSpace with same tier but with disable inheritance
--- a/test/e2e/parallel/space_test.go
+++ b/test/e2e/parallel/space_test.go
@@ -509,15 +509,7 @@ func TestSubSpaceInheritance(t *testing.T) {
 		// when
 		// we also have a subSpace with same tier but with disable inheritance
 		t.Logf("Create sub space with role: contributor")
-		subUser := NewSignupRequest(awaitilities).
-			ManuallyApprove().
-			RequireConditions(ConditionSet(Default(), ApprovedByAdmin())...).
-			NoSpace().
-			WaitForMUR().
-			Execute(t)
-		subSpaceMur := subUser.MUR
-
-		subSpace, subSpaceBindings := CreateSpaceWithBinding(t, awaitilities, subSpaceMur, "contributor",
+		subSpace, _, subSpaceBindings := CreateSpaceWithRole(t, awaitilities, "contributor",
 			testspace.WithSpecParentSpace(parentUser.Space.Name),
 			testspace.WithTierName("appstudio"),
 			testspace.WithSpecTargetCluster(memberAwait.ClusterName),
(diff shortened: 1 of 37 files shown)


test/e2e/parallel/space_test.go
```diff
@@ -498,19 +498,27 @@ func TestSubSpaceInheritance(t *testing.T) {
...
 		// we also have a subSpace with same tier but with disable inheritance
 		t.Logf("Create sub space with role: contributor")
-		subSpace, _, subSpaceBindings := CreateSpaceWithRole(t, awaitilities, "contributor",
-			testspace.WithSpecParentSpace(parentSpace.Name),
+		subUser := NewSignupRequest(awaitilities).
+			ManuallyApprove().
+			RequireConditions(ConditionSet(Default(), ApprovedByAdmin())...).
+			NoSpace().
+			WaitForMUR().
+			Execute(t)
```
MatousJobanek, line 517: correct me if I'm wrong, but we should not replace the creation of the sub-space with provisioning via UserSignup. In this case, we should really create directly the Space CR and we don't have to worry about the cleanup since the automatic cleanup is disabled for sub-spaces

rsoaresd: So sorry!! My bad! Thanks for noticing it!

rsoaresd: @MatousJobanek, should I rollback also the change in `CreateSpaceWithBinding` since we do not need to specify the role, or we can leave it this way for future proposes?

rsoaresd: @MatousJobanek, I was checking now and `CreateSpaceWithRole` creates via UserSignup

MatousJobanek: > should I rollback also the change in CreateSpaceWithBinding since we do not need to specify the role, or we can leave it this way for future proposes?

keep it as it is - it looks fine.

> I was checking now and CreateSpaceWithRole creates via UserSignup

good catch. 
The test verifies that we can create a sub-space with disabled inheritance (the SpaceBindings from the parent space are not inherited) and share it with a user. As a result, only the RoleBindings for user the sub-space was shared with will be created, and not for users that has access to the parent space. Technically speaking, provisioning the Space via a UserSignup would do the same, but it's more about how the test is written.

rsoaresd: Thank you so much! I addressed the creation of the space directly

test/e2e/parallel/space_test.go
```diff
@@ -506,19 +506,20 @@ func TestSubSpaceInheritance(t *testing.T) {
...
+			EnsureMUR().
+			Execute(t)
+		mur := parentUser.MUR
 
 		// when
 		// we also have a subSpace with same tier but with disable inheritance
 		t.Logf("Create sub space with role: contributor")
-		subSpace, _, subSpaceBindings := CreateSpaceWithRole(t, awaitilities, "contributor",
-			testspace.WithSpecParentSpace(parentSpace.Name),
+		subSpace, subSpaceBindings := CreateSpaceWithBinding(t, awaitilities, mur, "contributor",
```
MatousJobanek, line 521: this should be shared with someone else, not with the owner of the parent Space. We need to test that the SpaceBinding from the parent space won't be inherited anyhow - that the owner of the parent space "parentUser" doesn't have access to the sub-space with the disabled inheritance at all.

test/e2e/parallel/space_test.go
```diff
@@ -506,19 +506,28 @@ func TestSubSpaceInheritance(t *testing.T) {
...
+			SpaceTier("appstudio").
+			EnsureMUR().
+			Execute(t)
+
+		user := NewSignupRequest(awaitilities).
+			ManuallyApprove().
+			TargetCluster(memberAwait).
+			RequireConditions(ConditionSet(Default(), ApprovedByAdmin())...).
+			SpaceTier("appstudio").
+			EnsureMUR().
```
MatousJobanek, line 522: to speed up the logic - we don't need to verify all the resources provisioned for MUR and Space
```suggestion
			WaitForMUR().
```

MatousJobanek (approved):
Looks great :rocket: Thanks a lot for fixing the flakiness and improving the tests :+1: :1st_place_medal:

PR #993, opened 2024-06-07
Fix flaky e2e test

sbryzak:
Add a wait condition to ensure that the scheduled deactivation time is set before continuing.

### DIFF
--- a/test/e2e/user_management_test.go
+++ b/test/e2e/user_management_test.go
@@ -377,8 +377,9 @@ func (s *userManagementTestSuite) TestUserDeactivation() {
 				mur.Status.ProvisionedTime.String())
 
 			// The user should be set to deactivating, but not deactivated
-			userSignup, err = hostAwait.WaitForUserSignup(t, userSignup.Name, wait.UntilUserSignupHasConditions(
-				wait.ConditionSet(wait.Default(), wait.ApprovedAutomatically(), wait.Deactivating())...))
+			userSignup, err = hostAwait.WaitForUserSignup(t, userSignup.Name, wait.UntilUserSignupHasScheduledDeactivationTime(),
+				wait.UntilUserSignupHasConditions(
+					wait.ConditionSet(wait.Default(), wait.ApprovedAutomatically(), wait.Deactivating())...))
 			require.NoError(t, err)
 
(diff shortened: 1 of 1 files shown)


test/e2e/user_management_test.go
```diff
@@ -377,8 +377,9 @@ func (s *userManagementTestSuite) TestUserDeactivation() {
 				mur.Status.ProvisionedTime.String())
 
 			// The user should be set to deactivating, but not deactivated
-			userSignup, err = hostAwait.WaitForUserSignup(t, userSignup.Name, wait.UntilUserSignupHasConditions(
-				wait.ConditionSet(wait.Default(), wait.ApprovedAutomatically(), wait.Deactivating())...))
+			userSignup, err = hostAwait.WaitForUserSignup(t, userSignup.Name, wait.UntilUserSignupHasScheduledDeactivationTime(),
+				wait.UntilUserSignupHasConditions(
+					wait.ConditionSet(wait.Default(), wait.ApprovedAutomatically(), wait.Deactivating())...))
```
MatousJobanek, line 382: you combine `wait.Default()` together with `wait.Deactivating()`
the default has 
```go
{
	Type:   toolchainv1alpha1.UserSignupUserDeactivatedNotificationCreated,
	Status: corev1.ConditionFalse,
	Reason: "UserIsActive",
},
{
	Type:   toolchainv1alpha1.UserSignupUserDeactivatingNotificationCreated,
	Status: corev1.ConditionFalse,
	Reason: "UserNotInPreDeactivation",
},
```
and the deactivating:
```go
{
	Type:   toolchainv1alpha1.UserSignupUserDeactivatingNotificationCreated,
	Status: corev1.ConditionTrue,
	Reason: "NotificationCRCreated",
},
```
TBH, this is a bit confusing

MatousJobanek: ah, I missed that, sorry. 
Anyway, this is pretty confusing - why do we combine the `wait.Default()` together with `wait.Deactivating()` when they expect different values for the same condition

MatousJobanek (approved):

test/e2e/user_management_test.go
sbryzak: I'm not sure, but we can discuss it outside of this PR.

PR #995, opened 2024-06-14
Added e2e tests for registration service - GetSignup endpoint enhancements

sbryzak:
tests for **https://github.com/codeready-toolchain/registration-service/pull/432**

### DIFF
--- a/test/e2e/parallel/registration_service_test.go
+++ b/test/e2e/parallel/registration_service_test.go
@@ -851,6 +851,34 @@ func signup(t *testing.T, hostAwait *wait.HostAwaitility) (*toolchainv1alpha1.Us
 	return userSignup, token
 }
 
+type GetSignupEndpointAssertions = func(t *testing.T, await wait.Awaitilities, responseBody map[string]interface{})
+
+func getSignupIsProvisioned(t *testing.T, await wait.Awaitilities, username string, responseBody map[string]interface{}) {
+	hostAwait := await.Host()
+	memberAwait := await.Member1()
+	memberCluster, found, err := hostAwait.GetToolchainCluster(t, memberAwait.Namespace, toolchainv1alpha1.ConditionReady)
+	transformedUsername := commonsignup.TransformUsername(username, []string{"openshift", "kube", "default", "redhat", "sandbox"}, []string{"admin"})
+	require.NoError(t, err)
+	require.True(t, found)
+	assert.Equal(t, memberCluster.Spec.APIEndpoint, responseBody["apiEndpoint"])
+	assert.Equal(t, hostAwait.APIProxyURL, responseBody["proxyURL"])
+	assert.Equal(t, fmt.Sprintf("%s-dev", transformedUsername), responseBody["defaultUserNamespace"])
+	assertRHODSClusterURL(t, memberAwait, responseBody)
+
+}
+
+func assertGetSignupStatus(t *testing.T, await wait.Awaitilities, username, bearerToken string, assertions ...GetSignupEndpointAssertions) {
+	hostAwait := await.Host()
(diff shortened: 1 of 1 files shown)


MatousJobanek (commented):
I'm sorry, but where does it verify that the new Signup fields are correctly populated?

sbryzak:
> I'm sorry, but where does it verify that the new Signup fields are correctly populated?

In the new `signupHasExpectedDates` function, called [here](https://github.com/codeready-toolchain/toolchain-e2e/pull/995/files#diff-b016de217b3d9271da1bcf2a5cb1538894d8ec78b38b00a28491f1445787cfbcR367).

test/e2e/parallel/registration_service_test.go
```diff
@@ -363,7 +363,9 @@ func TestSignupOK(t *testing.T) {
 		VerifyResourcesProvisionedForSignup(t, await, userSignup, "deactivate30", "base")
 
 		// Call signup endpoint with same valid token to check if status changed to Provisioned now
-		assertGetSignupStatusProvisioned(t, await, identity.Username, token)
+		now := time.Now()
+		NewGetSignupClient(t, await, identity.Username, token).Invoke(signupIsProvisioned,
+			signupHasExpectedDates(now, now.Add(time.Hour*24*30), 30))
```
MatousJobanek, line 368: TBH, it's a bit weird to have it at different place than the other verification of the `UserSignup.Status.ScheduledDeactivationTimestamp`, let's move it to one place. 
Also, we should verify ideally also the other cases:
* when the time is changed, which can be done as part of https://github.com/codeready-toolchain/toolchain-e2e/blob/6147bdb9c181d279bd35711be7d600c3009fa7f0/test/e2e/user_management_test.go#L389-L397
* and when there is no deactivation time scheduled - corresponds to https://github.com/codeready-toolchain/toolchain-e2e/blob/6147bdb9c181d279bd35711be7d600c3009fa7f0/test/e2e/user_management_test.go#L462-L464

alexeykazakov (approved):
Looks good to me.

test/e2e/parallel/registration_service_test.go
sbryzak: It seems there's some confusion about this - the `NewGetSignupClient` returns a client for interacting with the `GetSignup` REST service endpoint in registration service, and not for the Kube API.  As such it definitely in my opinion belongs in the registration service tests, as this is the only place where we are invoking the `/signup` endpoint.

MatousJobanek: We are actually invoking the `/signup` endpoint in every single test when we provision users via the SignupRequest API, which means almost everywhere:
https://github.com/codeready-toolchain/toolchain-e2e/blob/f456fc114c0033083e15989642eec16157e9e652/testsupport/signup_request.go#L244-L245
Apart from that, the user_management_test.go already contains a test that calls the endpoint explicitly: 
https://github.com/codeready-toolchain/toolchain-e2e/blob/f456fc114c0033083e15989642eec16157e9e652/test/e2e/user_management_test.go#L620-L628 

The verification you added there is related not only to the rest endpoint on this own, but also to the deactivation states the account can be in. As I mentioned above, it would be nice to verify the other cases:
* when the time is changed
* when there is no deactivation time scheduled

Both these cases are tested in the user_management_test.go you added there before, so adding the verification of the Signup response would be for free and it would test the whole e2e flow.

Anyway, I won't insist on this and I won't block this PR because of that. Looking at the registration_service_test.go, there are already some duplicated things that can be moved to other tests or some generic functions and verify them for most of the provisioned UserSignups automatically, so this change won't make it much worse.

MatousJobanek (approved):

PR #996, opened 2024-06-14
feat: add keycloak make targets and docs

mfrancisc:
Add support for setting up keycloak in dev environment.

Running `make dev-deploy-latest DEV_SSO=true` for example will deploy sandbox with a preconfigured keycloak instance.

Depends on: https://github.com/codeready-toolchain/toolchain-cicd/pull/113

Jira: https://issues.redhat.com/browse/KUBESAW-122


make/run-cicd-script.mk
```diff
@@ -1,5 +1,5 @@
 
-OWNER_AND_BRANCH_LOCATION=codeready-toolchain/toolchain-cicd/master
+OWNER_AND_BRANCH_LOCATION=mfrancisc/toolchain-cicd/devkeycloak
```
mfrancisc, line 2: REMINDER

make/clean.mk
```diff
@@ -31,6 +31,12 @@ clean-cluster-wide-config:
 clean-toolchain-namespaces-in-e2e:
 	$(Q)-oc get projects --output=name | grep -E "toolchain-(member|host)(\-operator)?(\-[0-9]+)?" | xargs oc delete
 
+.PHONY: clean-toolchain-dev-sso-resources
+## Delete resources in the dev sso namespace
+clean-toolchain-dev-sso-resources:
+	$(Q)-oc delete keycloak -n ${DEV_SSO_NS} --all || true
```
filariow, line 37: we may add the `--wait` flag here to avoid problems with finalizers
```suggestion
	$(Q)-oc delete keycloak -n ${DEV_SSO_NS} --all --wait || true
```

make/clean.mk
```diff
@@ -43,13 +49,14 @@ clean-e2e-resources: clean-users clean-toolchain-namespaces-in-e2e clean-cluster
 clean-toolchain-namespaces-in-dev:
 	$(Q)oc delete namespace ${DEV_HOST_NS} || true
 	$(Q)oc delete namespace ${DEV_MEMBER_NS} || true
+	$(Q)oc delete namespace ${DEV_SSO_NS} || true
```
filariow, line 52: to avoid problems with finalizers, would it be a good idea to add `clean-toolchain-dev-sso-resources` as a dependency here?

```suggestion
clean-toolchain-namespaces-in-dev: clean-toolchain-dev-sso-resources
	$(Q)oc delete namespace ${DEV_HOST_NS} || true
	$(Q)oc delete namespace ${DEV_MEMBER_NS} || true
	$(Q)oc delete namespace ${DEV_SSO_NS} || true
```

filariow (approved):
well done! just some optional suggestions for reducing the risk of being annoyed by finalizers

make/clean.mk
alexeykazakov: > also, (just curiosity) shouldn't we crash if deletion don't complete successfully?

Will it complain if there is nothing to delete? If not then I guess we can fail with an error if it failed to delete but nothing to delete is ignored.

filariow: I tested locally with kind, kubectl, and oc. it seems not to fail if there is nothing to delete

make/clean.mk
mfrancisc: Good point, but TBH I'm not 100% sure. I added `clean-toolchain-dev-sso-resources` here https://github.com/codeready-toolchain/toolchain-e2e/pull/996/files#diff-4940ce933c6483bbf87f37d0041ad7587d86cac23b65179f97f1e7a70b6e6d9bR59 , before invoking `clean-toolchain-namespaces-in-dev`. That should do, since it's the only place where we use it.

make/clean.mk
mfrancisc: good point - addressed in https://github.com/codeready-toolchain/toolchain-e2e/pull/996/commits/316f41a00775dfcde57182ad796ab2480121a903  👍

make/clean.mk
filariow: Having both of them as `clean-dev-resources` dependency, will make it work fine, i.e. it won't hang on keycloak resources deletion; but if we invoke directly `clean-toolchain-namespaces-in-dev` it would likely hang. Introducing the `clean-toolchain-dev-sso-resources` dependency to `clean-toolchain-namespaces-in-dev` will avoid this behavior.

As it is possible for the user to invoke directly `clean-toolchain-namespaces-in-dev`, it makes sense to me to add it.

mfrancisc: Ok I see, it could hang if the env was deployed with `DEV_SSO=true` indeed, but users should use `clean-dev-resources` and not `clean-toolchain-namespaces-in-dev` I think. Anyways moving it solves this problem , done in https://github.com/codeready-toolchain/toolchain-e2e/pull/996/commits/e797e4d357c535f55480be116b8ee8f42fda5a07 . Thanks!

make/dev.mk
```diff
@@ -25,10 +26,15 @@ dev-deploy-e2e-two-members: deploy-e2e-to-dev-namespaces-two-members print-reg-s
 .PHONY: deploy-e2e-to-dev-namespaces
 deploy-e2e-to-dev-namespaces:
 	$(MAKE) deploy-e2e MEMBER_NS=${DEV_MEMBER_NS} SECOND_MEMBER_MODE=false HOST_NS=${DEV_HOST_NS} REGISTRATION_SERVICE_NS=${DEV_REGISTRATION_SERVICE_NS} ENVIRONMENT=${DEV_ENVIRONMENT} E2E_TEST_EXECUTION=false IS_OSD=${IS_OSD} DEPLOY_LATEST=${DEPLOY_LATEST}
+	if [[ ${DEV_SSO} == true ]]; then \
+		$(MAKE) download-assets ASSETS_FOLDER=scripts/ci/dev-sso; \
+		$(MAKE) run-cicd-script SCRIPT_PATH=scripts/ci/setup-dev-sso.sh  SCRIPT_PARAMS="--sso-ns $(DEV_SSO_NS)"; \
+	fi
```
MatousJobanek, line 32: looking at this, I'm wondering, wouldn't it be easier to move the assets and the script into toolchain-e2e and execute only the `wait-until-is-installed.sh` from toolchain-cicd repo? I guess that this is the only reason for putting it into the toolchain-cicd repo, right?

mfrancisc: this is one reason indeed, the other reason is that we have all the bash scripts in toolchain-cicd and they are consumed by toolchain-e2e, so I've just followed the convention here. 

I'm fine with moving them but I don't think it will simplify much. Plus the download assets methods seems to be working fine.

fbm3307 (approved):

make/dev.mk
MatousJobanek: Yeah, I understand that you wanted to follow the same format.
Technically speaking, we could move the whole `ci/` folder into toolchain-e2e repo, I don't see any significant reason for keeping them there. IIRC, they are there for the historical reasons - some of them were called from multiple repos.

Anyway, if you don't think that it would simplify anything, then feel free to keep it as it is.

make/dev.mk
```diff
@@ -25,10 +26,15 @@ dev-deploy-e2e-two-members: deploy-e2e-to-dev-namespaces-two-members print-reg-s
...
 
 .PHONY: deploy-e2e-to-dev-namespaces-two-members
 deploy-e2e-to-dev-namespaces-two-members:
 	$(MAKE) deploy-e2e MEMBER_NS=${DEV_MEMBER_NS} MEMBER_NS_2=${DEV_MEMBER_NS_2} HOST_NS=${DEV_HOST_NS} REGISTRATION_SERVICE_NS=${DEV_REGISTRATION_SERVICE_NS} ENVIRONMENT=${DEV_ENVIRONMENT} E2E_TEST_EXECUTION=false IS_OSD=${IS_OSD}
+	if [[ ${DEV_SSO} == true ]]; then $(MAKE) run-cicd-script SCRIPT_PATH=scripts/ci/setup-dev-sso.sh  SCRIPT_PARAMS="--sso-ns $(DEV_SSO_NS)"; fi
```
MatousJobanek, line 37: don't you need to download the assets for this call and the other ones as well?

fbm3307: +1 to Matous, but just for my understanding, what would be the main difference, if we don't download the assets Would we have any other way to use it ?

mfrancisc: Good point, I forgot to update the other make targets when I've moved the manifests to the cicd repo 🤦‍♂️ . 

Done in https://github.com/codeready-toolchain/toolchain-e2e/pull/996/commits/e4d36d5c134a7e564a83ce1f5c3e6c53d8c5bc31 Thanks!

mfrancisc: > but just for my understanding, what would be the main difference, if we don't download the assets Would we have any other way to use it ?

well in this case those targets will not work ( if they are executed before the one with the download assets call ) since the script will not have the manifests to apply , thus will error.

make/dev.mk
mfrancisc: It makes sense to me to move them all, if this doesn't break anything ( I'm thinking about konflux testing for example ) . I think moving them all would simplify, moving just this one not that much IMO. So if it's ok for you let's do it in a refactoring follow up PR and simplify the make targets here as well.

make/dev.mk
```diff
@@ -34,7 +34,10 @@ deploy-e2e-to-dev-namespaces:
 .PHONY: deploy-e2e-to-dev-namespaces-two-members
 deploy-e2e-to-dev-namespaces-two-members:
 	$(MAKE) deploy-e2e MEMBER_NS=${DEV_MEMBER_NS} MEMBER_NS_2=${DEV_MEMBER_NS_2} HOST_NS=${DEV_HOST_NS} REGISTRATION_SERVICE_NS=${DEV_REGISTRATION_SERVICE_NS} ENVIRONMENT=${DEV_ENVIRONMENT} E2E_TEST_EXECUTION=false IS_OSD=${IS_OSD}
-	if [[ ${DEV_SSO} == true ]]; then $(MAKE) run-cicd-script SCRIPT_PATH=scripts/ci/setup-dev-sso.sh  SCRIPT_PARAMS="--sso-ns $(DEV_SSO_NS)"; fi
+	if [[ ${DEV_SSO} == true ]]; then \
+		$(MAKE) download-assets ASSETS_FOLDER=scripts/ci/dev-sso; \
+		$(MAKE) run-cicd-script SCRIPT_PATH=scripts/ci/setup-dev-sso.sh  SCRIPT_PARAMS="--sso-ns $(DEV_SSO_NS)"; \
+	fi
```
filariow, line 40: if `download-assets` fails, we should not `run-cicd-script`

```suggestion
	if [[ ${DEV_SSO} == true ]]; then \
		$(MAKE) download-assets ASSETS_FOLDER=scripts/ci/dev-sso && \
		$(MAKE) run-cicd-script SCRIPT_PATH=scripts/ci/setup-dev-sso.sh  SCRIPT_PARAMS="--sso-ns $(DEV_SSO_NS)"; \
	fi
```

same applies at L52 and L59

mfrancisc: good catch!

make/dev.mk
```diff
@@ -25,10 +26,18 @@ dev-deploy-e2e-two-members: deploy-e2e-to-dev-namespaces-two-members print-reg-s
...
 
 .PHONY: deploy-e2e-to-dev-namespaces-two-members
 deploy-e2e-to-dev-namespaces-two-members:
 	$(MAKE) deploy-e2e MEMBER_NS=${DEV_MEMBER_NS} MEMBER_NS_2=${DEV_MEMBER_NS_2} HOST_NS=${DEV_HOST_NS} REGISTRATION_SERVICE_NS=${DEV_REGISTRATION_SERVICE_NS} ENVIRONMENT=${DEV_ENVIRONMENT} E2E_TEST_EXECUTION=false IS_OSD=${IS_OSD}
+	if [[ ${DEV_SSO} == true ]]; then \
+		$(MAKE) download-assets ASSETS_FOLDER=scripts/ci/dev-sso && \
+		$(MAKE) run-cicd-script SCRIPT_PATH=scripts/ci/setup-dev-sso.sh  SCRIPT_PARAMS="--sso-ns $(DEV_SSO_NS)"; \
+	fi
```
MatousJobanek, line 40: you could extract it to a separate target to limit copy-paste, right?
```suggestion
	$(MAKE) setup-dev-sso DEV_SSO=${DEV_SSO}

.PHONY: deploy-e2e-to-dev-namespaces-two-members
deploy-e2e-to-dev-namespaces-two-members:
	$(MAKE) deploy-e2e MEMBER_NS=${DEV_MEMBER_NS} MEMBER_NS_2=${DEV_MEMBER_NS_2} HOST_NS=${DEV_HOST_NS} REGISTRATION_SERVICE_NS=${DEV_REGISTRATION_SERVICE_NS} ENVIRONMENT=${DEV_ENVIRONMENT} E2E_TEST_EXECUTION=false IS_OSD=${IS_OSD}
	$(MAKE) setup-dev-sso DEV_SSO=${DEV_SSO}

setup-dev-sso:
	if [[ ${DEV_SSO} == true ]]; then \
		$(MAKE) download-assets ASSETS_FOLDER=scripts/ci/dev-sso && \
		$(MAKE) run-cicd-script SCRIPT_PATH=scripts/ci/setup-dev-sso.sh  SCRIPT_PARAMS="--sso-ns $(DEV_SSO_NS)"; \
	fi
```

make/run-cicd-script.mk
```diff
@@ -1,5 +1,9 @@
 
-OWNER_AND_BRANCH_LOCATION=codeready-toolchain/toolchain-cicd/master
+GH_OWNER=mfrancisc
+GH_REPO=toolchain-cicd
+GH_BRANCH=devkeycloak
```
MatousJobanek, line 4: reminder

MatousJobanek (approved):

make/dev.mk
mfrancisc: good point! This looks better!

mfrancisc: I've used this also in the other dev targets see https://github.com/codeready-toolchain/toolchain-e2e/pull/996/commits/0c742da53aa10ad4ad8c8345ce2278861e80c739

make/run-cicd-script.mk
mfrancisc: thanks! I'll remove those once I merge the cicd PR!

dev_install.adoc
```diff
@@ -16,6 +16,15 @@ Configure authentication for the cluster using one of the following options:
 Option #1: Contact a member of the Dev Sandbox Team for instructions on how to configure the cluster to use our internal Dev SSO. +
 Option #2: Configure your own Keycloak server and set up authentication on the OpenShift cluster: https://docs.openshift.com/container-platform/4.6/authentication/configuring-internal-oauth.html
 
+
+Option #3: Deploy and configure keycloak internally as part of the cluster. Just add `DEV_SSO=true` parameter to the `dev` targets. For eg.: `make dev-deploy-latest DEV_SSO=true` will deploy latest version of the operators with a preconfigured keycloak instance.
```
MatousJobanek, line 20: I know that it's the default user & password is printed at the end of the command, but I didn't notice it at the first glance. You can mention it here as well.
```suggestion
Option #3: Deploy and configure keycloak internally as part of the cluster. Just add `DEV_SSO=true` parameter to the `dev` targets. For eg.: `make dev-deploy-latest DEV_SSO=true` will deploy latest version of the operators with a preconfigured keycloak instance and one default keycloak user `user1@user.us` with password `user1`.
```

dev_install.adoc
```diff
@@ -16,6 +16,15 @@ Configure authentication for the cluster using one of the following options:
...
+If you are presented with the following error, you just need to ignore the certificate problem for the `https://keycloak-<dev-sso-namespace>.<domain>/auth/js/keycloak.js :
```
MatousJobanek, line 21: you can go to the shorter version of the URL which is also printed at the end of the command.
```suggestion
If you are presented with the following error, then you need to accept the self-signed certificate of the dev Keycloak instance first. Go to `https://keycloak-<dev-sso-namespace>.<domain>/auth (the complete link is printed out at the end of the command) and accept the certificate.:
```

MatousJobanek (approved):
It's awesome have this automated. Just tested it and it works like a charm.
Just two suggestions for the README

dev_install.adoc
mfrancisc: thanks a lot for the feedback! makes sense!

make/run-cicd-script.mk
mfrancisc: done in https://github.com/codeready-toolchain/toolchain-e2e/pull/996/commits/09ae4aa053638c9e9741bae7d61b9aaf08345ac6

PR #997, opened 2024-06-14
Make sure kubeconfig is present in the ToolchainCluster secrets on the host cluster.

metlos:
add a simple test that checks for the presence of kubeconfig entry in the ToolchainCluster secret in the host cluster.

Associated PRs:
- host-operator: https://github.com/codeready-toolchain/host-operator/pull/1042
- toolchain-common: https://github.com/codeready-toolchain/toolchain-common/pull/404

### DIFF
--- a/test/e2e/toolchaincluster_test.go
+++ b/test/e2e/toolchaincluster_test.go
@@ -84,6 +84,7 @@ func verifyToolchainCluster(t *testing.T, await *wait.Awaitility, otherAwait *wa
 			secretRef(secretCopy.Name),
 			owner(current.Labels["ownerClusterName"]),
 			namespace(current.Labels["namespace"]),
+			disableTLS(current.Spec.DisabledTLSValidations),
 			capacityExhausted, // make sure this cluster cannot be used in other e2e tests
 		)
 
@@ -130,6 +131,7 @@ func verifyToolchainCluster(t *testing.T, await *wait.Awaitility, otherAwait *wa
 			secretRef(secretCopy.Name),
 			owner(current.Labels["ownerClusterName"]),
 			namespace(current.Labels["namespace"]),
+			disableTLS(current.Spec.DisabledTLSValidations),
 			capacityExhausted, // make sure this cluster cannot be used in other e2e tests
 		)
 
@@ -236,5 +238,12 @@ func apiEndpoint(url string) clusterOption {
 func caBundle(bundle string) clusterOption {
 	return func(c *toolchainv1alpha1.ToolchainCluster) {
 		c.Spec.CABundle = bundle
+	}
+}
+
+// disableTLS sets the DisabledTLSValidations field
+func disableTLS(validations []toolchainv1alpha1.TLSValidation) clusterOption {
+	return func(c *toolchainv1alpha1.ToolchainCluster) {
+		c.Spec.DisabledTLSValidations = validations
 	}
 }
--- a/test/e2e/toolchaincluster_test.go
+++ b/test/e2e/toolchaincluster_test.go
@@ -58,7 +58,7 @@ func verifyToolchainCluster(t *testing.T, await *wait.Awaitility, otherAwait *wa
 	// based on the secrets, we will no longer require this migration step and this test will be removed.
 	t.Run("kubeconfig is generated from the connection details", func(t *testing.T) {
 		if kubeConfigMustExist {
-			targetClient, _ := space.NewKubeClientFromSecret(t, await.Client, current.Spec.SecretRef.Name, current.Namespace)
+			targetClient, _ := space.NewKubeClientFromSecret(t, await.Client, current.Spec.SecretRef.Name, current.Namespace, toolchainv1alpha1.AddToScheme)
 			space.ValidateKubeClient(t, targetClient, await.Namespace, &toolchainv1alpha1.ToolchainClusterList{})
 		} else {
 			secret := &corev1.Secret{}
--- a/test/e2e/toolchaincluster_test.go
+++ b/test/e2e/toolchaincluster_test.go
@@ -8,7 +8,6 @@ import (
 	"time"
 
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
-	"github.com/codeready-toolchain/toolchain-common/pkg/test"
 	. "github.com/codeready-toolchain/toolchain-e2e/testsupport"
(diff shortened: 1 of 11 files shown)


test/e2e/toolchaincluster_test.go
```diff
@@ -46,6 +52,18 @@ func verifyToolchainCluster(t *testing.T, await *wait.Awaitility, otherAwait *wa
...
+	// NOTE: this checks that there is the migration step in place that converts the connection details for the cluster that is currently
+	// spread over the secret and ToolchainCluster is also set as a kubeconfig inside the secret. Once we migrate to create the ToolchainClusters
+	// based on the secrets, we will no longer require this migration step and this test will be removed.
+	t.Run("kubeconfig is generated from the connection details", func(t *testing.T) {
+		secret := &corev1.Secret{}
+		require.NoError(t, await.Client.Get(context.TODO(), client.ObjectKey{Name: current.Spec.SecretRef.Name, Namespace: current.Namespace}, secret))
+
+		// this is enough for the test here. The correctness of the contained kubeconfig is checked in the unit tests of the controller in toolchain-common.
+		_, present := secret.Data["kubeconfig"]
+		assert.Equal(t, kubeConfigMustExist, present)
```
MatousJobanek, line 64: There are actually already functions that helps you with verifying the validity of the kubeconfig
https://github.com/codeready-toolchain/toolchain-e2e/blob/6c68589aa96acba09217ee95e9fdc05d0ba19fb0/testsupport/space/spacerequest.go#L115-L149

There is no need to do anything fancy, I would just export the functions and use it here - something like this:
```suggestion
		if kubeConfigMustExist {
			kubeconfig, _ := space.NewKubeClientFromSecret(t, await.Client, current.Spec.SecretRef.Name, current.Namespace)
			space.ValidateKubeClient(t, kubeconfig, current.Labels["namespace"])
		} else {
			secret := &corev1.Secret{}
			require.NoError(t, await.Client.Get(context.TODO(), client.ObjectKey{Name: current.Spec.SecretRef.Name, Namespace: current.Namespace}, secret))

			_, present := secret.Data["kubeconfig"]
			assert.False(t, present)
		}
```
But if it would require more changes  than I assume in the code above, then please disregard my comment and keep it as it is.

testsupport/space/spacerequest.go
```diff
@@ -141,9 +142,10 @@ func newKubeClientFromSecret(t *testing.T, cl client.Client, secretName, secretN
...
 }
 
-// validateKubeClient validates the the kube client can access the given namespace
-func validateKubeClient(t *testing.T, namespaceAccessClient client.Client, namespace string) {
-	secretsList := &corev1.SecretList{}
-	require.NoError(t, namespaceAccessClient.List(context.TODO(), secretsList, client.InNamespace(namespace)))
-	require.NotEmpty(t, secretsList)
+// ValidateKubeClient validates the the kube client can access the given namespace by listing objects using the provided list instance
+// The list is checked to not be empty.
+func ValidateKubeClient(t *testing.T, namespaceAccessClient client.Client, namespace string, list client.ObjectList) {
```
mfrancisc, line 147: minor - maybe we can move this to the `testsupport/util` or `testsupport/auth` package ? Probably it will look nicer then `space.ValidateKubeClient` from a client perspective. But feel free to leave it as is.

test/e2e/toolchaincluster_test.go
```diff
@@ -23,13 +25,18 @@ func TestToolchainClusterE2E(t *testing.T) {
 	memberAwait := awaitilities.Member1()
 	memberAwait.WaitForToolchainClusterResources(t)
 
-	verifyToolchainCluster(t, hostAwait.Awaitility, memberAwait.Awaitility)
-	verifyToolchainCluster(t, memberAwait.Awaitility, hostAwait.Awaitility)
+	// NOTE: because we cannot introduce changes to host operator and member operator simulateneously to enforce
+	// backwards compatibility, we first introduce the conversion of connection details into kubeconfig to the host
```
mfrancisc, line 29: I'm sure I'm missing something - but why only to host operator if the `kubeconfig` data is not actually used. I mean it would be still backwards compatible , unless I'm missing something. 

Even if I agree with introducing this incrementally.

mfrancisc (commented):
Looks great 👏 

I have few minor comments only.

Would it make sense to wrap temporary/migration code with `TODO` or something that will make it easier to identify later when we need to remove it. But feel free to leave it as is if you think it's clear and you can easly identify the temporary code.

testsupport/space/spacerequest.go
MatousJobanek: this is a temporal check for now - just for the migration time. so, it's fine to keep it where it is

mfrancisc: why is temporal ? aren't we switching to kubeconfig so we will still need to validate it in the toolchaincluster secret even later ? but maybe I'm missing something.

metlos: It will be validated a million times in other places when it is actually used for the connections among the clusters.

But you're right, a quick explicit test for it may reduce the head-scratching when problems with the connections arise.

test/e2e/toolchaincluster_test.go
metlos: I think it is because our CI will not pick up the changes from both host-operator and member-operator at the same time, am I right @MatousJobanek?

testsupport/space/spacerequest.go
MatousJobanek: if there is gonna be a problem with the kubeconfig created for ToolchainCluster, then the ToolchainCluster won't get ready in the first place (which means that we wouldn't be able to do almost anything), which means that we wouldn't even reach to this point when we could try to test it explicitly.

So as @metlos mentioned, the validity of the kubeconfig will be verified by using it, this is not the case for now - the kubeconfigs are created only (as part of the migration process) but not used yet.

test/e2e/toolchaincluster_test.go
MatousJobanek: yes, you are right. The change can be promoted to host and member only in two steps - they are independent operators/deployments. For that reason, the toolchain-e2e is configured to be able to pair with only one PR.

testsupport/space/spacerequest.go
metlos: I moved the functions to `testsupport/util`.

MatousJobanek (approved):

PR #1002, opened 2024-06-26
KUBESAW-124: Makefile target for dependency validations

fbm3307:
This is to create makefile target for dependency validations in each repository so that this can be used to verify if there are any compilation issues after updating the dependencies of toolchain-common/api

Similar PRs
- Member-Operator - https://github.com/codeready-toolchain/member-operator/pull/581 
- Host-Operator - https://github.com/codeready-toolchain/host-operator/pull/1052
- Registration-Service - https://github.com/codeready-toolchain/registration-service/pull/440 
- KSTCL - https://github.com/kubesaw/ksctl/pull/43


make/verify-dependencies.mk
```diff
@@ -0,0 +1,12 @@
+.PHONY: verify-dependencies
+## Runs commands to verify after the updated dependecies of toolchain-common/API(go mod replace), if the repo needs any changes to be made
+verify-dependencies: tidy vet test lint-go-code
```
MatousJobanek, line 3: you can also run 
```
go test ./... -skip '.*'
```
to verify that all tests can be compiled

fbm3307: Thanks, This is a good add-on , 

added it

make/verify-dependencies.mk
```diff
@@ -0,0 +1,16 @@
...
+.PHONY: tidy
+tidy: 
+	go mod tidy
+
+.PHONY: vet
+vet:
+	go vet ./...
+
+.PHONY: compile
+compile:
```
MatousJobanek, line 14: `compile` can be a bit confusing
```suggestion
.PHONY: go-test-skip-all
go-test-skip-all:
```

make/verify-dependencies.mk
```diff
@@ -0,0 +1,16 @@
...
+	go test ./... -skip '.*'
```
MatousJobanek, line 15: I already left a comment in the member-operator - everything is related to `go` commands and it's also verifying validity of the go dependencies, let's move it to `go.mk` file

make/verify-dependencies.mk
fbm3307: done

make/verify-dependencies.mk
fbm3307: i thought about keeping the dependency check separate but adding everything related to go in one file makes sense.. moved it

make/go.mk
```diff
@@ -14,3 +14,19 @@ build:
 	mkdir -p $(OUT_DIR)/bin || true
 	$(Q)CGO_ENABLED=0 GOARCH=${goarch} GOOS=linux \
 		go build ${V_FLAG} ./...
+
+.PHONY: verify-dependencies
+## Runs commands to verify after the updated dependecies of toolchain-common/API(go mod replace), if the repo needs any changes to be made
+verify-dependencies: tidy vet compile test lint-go-code
```
MatousJobanek, line 20:
```suggestion
verify-dependencies: tidy vet go-test-skip-all test lint-go-code
```

make/verify-dependencies.mk
fbm3307: in go.mk the commands are related to build, also registration service doesn't have go.mk it has build.mk for the same.. i was thinking to either have a separate file which is same across all repos, or rename build.mk file in registration service to go.mk (same as other repos) WDYT?

MatousJobanek: let's move it to `go.mk` - that's the place where I would look for any commands/targets related to go binary - and as you mentioned, rename `build.mk` to `go.mk` in reg-service to keep it consistent :+1:

MatousJobanek (approved):

PR #1007, opened 2024-07-09
add verification of (re)generating appstudio tiers via ksctl

MatousJobanek:
1. clone latest version of ksctl
2. install it locally
3. copy appstudio tiers (from host-operator) into /tmp folder
4. run `ksctl generate nstemplatetiers` for the appstudio tiers
5. apply the generated TierTemplates (and add an annotation to all resources)
6. verify in the tests that the NSTemplateTier doesn't get overridden and keeps the generated TierTemplates

you know, just to make sure that this will work as soon as Konflux team moves it to a infra-deployments

[KUBESAW-62](https://issues.redhat.com/browse/KUBESAW-62)

### DIFF
--- a/make/test.mk
+++ b/make/test.mk
@@ -439,7 +439,7 @@ test:
 ## Regenerate and re-apply appstudio tiers using the latest version of ksctl
 tiers-via-ksctl:
 	$(eval HOST_REPO_LOC = $(or ${HOST_REPO_PATH}, /tmp/host-operator-master))
-ifeq ($(HOST_REPO_PATH),/tmp/host-operator-master)
+ifeq ($(HOST_REPO_PATH),)
 	rm -rf /tmp/host-operator-master 2>/dev/null || true
 	git clone --depth 1 https://github.com/codeready-toolchain/host-operator.git /tmp/host-operator-master
 endif
--- a/make/test.mk
+++ b/make/test.mk
@@ -139,7 +139,7 @@ test-e2e-registration-local:
 	$(MAKE) test-e2e REG_REPO_PATH=${PWD}/../registration-service
 
 .PHONY: e2e-run-parallel
-e2e-run-parallel:
+e2e-run-parallel: tiers-via-ksctl
 	@echo "Running e2e tests in parallel..."
 	$(MAKE) execute-tests MEMBER_NS=${MEMBER_NS} MEMBER_NS_2=${MEMBER_NS_2} HOST_NS=${HOST_NS} REGISTRATION_SERVICE_NS=${REGISTRATION_SERVICE_NS} TESTS_TO_EXECUTE="./test/e2e/parallel" E2E_PARALLELISM=100
 	@echo "The parallel e2e tests successfully finished"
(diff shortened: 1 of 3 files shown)


make/test.mk
```diff
@@ -434,3 +434,23 @@ display-eval:
...
+	rm -rf /tmp/e2e-tiers-out 2>/dev/null || true
+	mkdir /tmp/e2e-tiers
+	cp -r ${HOST_REPO_LOC}/deploy/templates/nstemplatetiers/appstudio /tmp/e2e-tiers/
+	cp -r ${HOST_REPO_LOC}/deploy/templates/nstemplatetiers/appstudio-env /tmp/e2e-tiers/
+	cp -r ${HOST_REPO_LOC}/deploy/templates/nstemplatetiers/appstudiolarge /tmp/e2e-tiers
+	rm -rf /tmp/ksctl-master 2>/dev/null || true
+	git clone https://github.com/kubesaw/ksctl.git /tmp/ksctl-master
+	$(MAKE) -C /tmp/ksctl-master install GOBIN=/tmp/ksctl-master/bin
+	/tmp/ksctl-master/bin/ksctl generate nstemplatetiers --source /tmp/e2e-tiers --out-dir /tmp/e2e-tiers-out
+	oc kustomize /tmp/e2e-tiers-out | sed 's/toolchain-host-operator/${HOST_NS}/g;s/^metadata:/metadata:\n  annotations:\n    generated-by: ksctl/g' | oc apply -f -
```
alexeykazakov, line 456: I'm trying to understand what exactly this is supposed to check.
If I'm not mistaken it verifies that:
- the latest (from master) `ksctl generate mstemplatetiers` works as expected.
And it's doing it by:
- generating tiers using the built-in appstudio tiers as source
- applying them as is (but also adding an annotation) before running the test

So my questions:
- Are you trying to override the built-in appstudio tiers here? But the built-in are generated during the operator startup while the ksctl's ones are applied before that. What am I missing? :)
- I thought we want to remove the appstudio tiers from the host operator when they are moved to infradeployment. Or your plan is something else? Can you elaborate on this.
- If the purpose is to actually verify that ksctl generates kubesaw compatible tiers then shouldn't we do it a bit differently:
  - use another tier as source which is not going anywhere (maybe create a special test tier for that?)
  - do ksctl - e2e-tests pairing so we can catch ksctl generate bugs on the ksctl PR step. Not after it gets merged to master.
  - I'm not sure it's worth doing all of this though while we are working on deep refactoring of the tier mechanism. I'm just trying to understand what scenario you are trying to cover by this PR :)

MatousJobanek: Let me clarify a few things here:
* Right now, the appstudio tiers are being managed by host-operator. 
* Konflux team already started working on moving the tiers to infra-deployments. 
* We cannot drop the tiers from the host-operator and exactly at the same time create them in infra-deployments
* There is gonna be some small amount of time when the tiers will be managed by both, host-operator & infra-deployments. 

This PR is verifying two things:
1. That the situation that I described above will work, so the tiers updated by ArgoCD won't get replaced back by the host-operator in the middle of execution. I know that it won't since we update the tiers only once, when we start the host-operator, I just wanted to be 100% sure that we won't break anything by that approach. 
2. That the tiers generated by the ksctl are the same as those generated by host-operator. 

As soon as Konflux moves the tiers to infra-deployments, we can drop the tiers from the host-operator. After that, I want to:
1. move the tiers into toolchain-e2e and generate them by the ksctl command
2. maybe simplify them - keep only a few resources there

In other words, make the appstudio tiers a testing tiers you talked about.

So, to answer your questions:

> Are you trying to override the built-in appstudio tiers here? But the built-in are generated during the operator startup while the ksctl's ones are applied before that. What am I missing? :)

ksctl is executed after startup of the operator

> I thought we want to remove the appstudio tiers from the host operator when they are moved to infradeployment. Or your plan is something else? Can you elaborate on this.

yes, we do want to remove them from host-operator, but as I mentioned above, we need to do it in two step and I'm covering the first step here.

> If the purpose is to actually verify that ksctl generates kubesaw compatible tiers then shouldn't we do it a bit differently:
>    use another tier as source which is not going anywhere (maybe create a special test tier for that?)

We could, but we already have logic and written tests and verification checks for the appstudio tiers, so re-using them is much easier for us than creating a new ones.
At the end of the day, we will have something like tests tiers. We could also rename them with different names, but let's do it in steps.

> do ksctl - e2e-tests pairing so we can catch ksctl generate bugs on the ksctl PR step. Not after it gets merged to master.

Yeah, that's something what I have in my mind as well. I started with something simple, but we will need something more advanced because we should incorporate ksctl into e2e tests even more. TBH, I'm not sure if we want to use the same logic for pairing or not. I need to think about it a bit more.

> I'm not sure it's worth doing all of this though while we are working on deep refactoring of the tier mechanism. I'm just trying to understand what scenario you are trying to cover by this PR :)

I had the same question - if it's worth doing it or not. I wanted to be 100% sure that we won't break anything and adding the simulation of the scenario to e2e tests seemed to be the best approach to me. That's also why I used the simplest approach for now, that's why I didn't create some dummy tiers, and why I didn't deal with pairing for now. So as simple approach as possible.

alexeykazakov: Thanks for the detailed context! The things I missed but you know have clarified it now:
- this PR (as is) just for the migration step. To cover time when we have appstudio tiers in both the host operator and infra-deployment.
- the tiers are generated (and applied) in this PR after the operator started. Not before.
- this PR is just a relatively quick solution for the migration step. All other concerns/improvements will be addressed later since it will take much more time, thinking, effort.

It all makes sense to me :+1:

alexeykazakov (approved):

make/test.mk
MatousJobanek: Just to summarize the next steps:
1. Konflux team will create the tiers in infra-deployments
2. we will move the appstudio tiers as they are from host-operator PR to toolchain-e2e https://issues.redhat.com/browse/KUBESAW-61
3. we will keep testing the tiers in the scenarios we have in toolchain-e2e - no changes for now. The reason why I want to move them all as they are is that they are important to test different scenarios:
    i. `appstudio` to verify a standard tiers generated by ksctl. 
    ii. `appstudiolarge` to verify a tier generated by ksctl and using `based_on_tier.yaml` file
    iii. `appstudio-env` to verify the a tier with "namespace-manager" SA used by SpaceRequest
    iv. all of them testing tiers with multiple SpaceRoles
5. then we can plan the next steps - there are technically two areas:

Tiers:
* simplify the tiers to make them simpler
* maybe rename them
* As soon as we finish the decoupling effort, we can technically drop the appstudiolarge tier as it won't test anything special that won't be covered by other tiers

Ksctl:
* incorporate ksctl into toolchain-e2e more
* support pairing
* maybe start running e2e tests in ksctl repo as well

PR #1008, opened 2024-07-11
Tests for feature toggles

alexeykazakov:
Paired with https://github.com/codeready-toolchain/member-operator/pull/580


testsupport/wait/member.go
```diff
@@ -989,6 +989,48 @@ func (a *MemberAwaitility) WaitForRoleBinding(t *testing.T, namespace *corev1.Na
 	return roleBinding, err
 }
 
+// WaitForObject waits until an Object with the given name exists in the given namespace
+func (a *MemberAwaitility) WaitForObject(t *testing.T, namespaceName string, name string, obj client.Object, objList client.ObjectList) error {
+	t.Logf("waiting for Object '%s' in namespace '%s'", name, namespaceName)
+	err := wait.Poll(a.RetryInterval, a.Timeout, func() (done bool, err error) {
+		if err := a.Client.Get(context.TODO(), types.NamespacedName{Namespace: namespaceName, Name: name}, obj); err != nil {
```
MatousJobanek, line 996: reusing the same object in the loop is risky. We should always use a new instance of the object (or we need to know exactly what resources for which purpose we are getting, which is hard in such a generic function). 
If we don't do that, then the resulting object may contain fields/conditions/values that are not present in the corresponding resource anymore. 
In other words, the final content of the object doesn't have to reflect what is in the cluster.

test/e2e/parallel/nstemplatetier_test.go
```diff
@@ -303,3 +307,132 @@ func TestKsctlGeneratedTiers(t *testing.T) {
...
+			SpaceTier("base1ns").
+			TargetCluster(memberAwait).
+			RequireConditions(wait.ConditionSet(wait.Default(), wait.ApprovedByAdmin())...).
+			Execute(t)
+		// and promote that space to the ftier tier
+		tiers.MoveSpaceToTier(t, hostAwait, "featured-user", tier.Name)
+
+		// then
+
+		// Verify that the space has the feature annotation
```
MatousJobanek, line 348:
```suggestion
		// Verify that the space has the feature annotation - the weight is set to 100 so it should be added to all Spaces
```

test/e2e/parallel/nstemplatetier_test.go
```diff
@@ -303,3 +307,132 @@ func TestKsctlGeneratedTiers(t *testing.T) {
...
+		require.NoError(t, err)
+
+		t.Run("disable feature", func(t *testing.T) {
+			// when
+
+			// Now let's disable the feature for the Space by removing the feature annotation
+			an := space.Annotations
+			delete(an, toolchainv1alpha1.FeatureToggleNameAnnotationKey)
+			space, err := hostAwait.UpdateSpace(t, user.Space.Name, func(s *toolchainv1alpha1.Space) {
+				s.Annotations = an
```
MatousJobanek, line 366: why not
```suggestion
			_, err := hostAwait.UpdateSpace(t, user.Space.Name, func(s *toolchainv1alpha1.Space) {
				delete(space.Annotations, toolchainv1alpha1.FeatureToggleNameAnnotationKey)
```

test/e2e/parallel/nstemplatetier_test.go
```diff
@@ -303,3 +307,132 @@ func TestKsctlGeneratedTiers(t *testing.T) {
...
+		// and promote that space to the ftier tier
+		tiers.MoveSpaceToTier(t, hostAwait, "featured-user", tier.Name)
+
+		// then
+
+		// Verify that the space has the feature annotation
+		space, err := hostAwait.WaitForSpace(t, user.Space.Name)
+		require.NoError(t, err)
+		require.NotEmpty(t, space.Annotations)
+		assert.Equal(t, "test-feature", space.Annotations[toolchainv1alpha1.FeatureToggleNameAnnotationKey])
```
mfrancisc, line 352: should we check the annotation on the NSTemplateSet CR ?

test/e2e/parallel/nstemplatetier_test.go
```diff
@@ -303,3 +307,132 @@ func TestKsctlGeneratedTiers(t *testing.T) {
...
+				// when
+
+				// Now let's re-enable the feature for the Space by restoring the feature annotation
+				an := space.Annotations
+				if an == nil {
+					an = make(map[string]string)
+				}
+				an[toolchainv1alpha1.FeatureToggleNameAnnotationKey] = "test-feature"
+				_, err := hostAwait.UpdateSpace(t, user.Space.Name, func(s *toolchainv1alpha1.Space) {
+					s.Annotations = an
```
MatousJobanek, line 384: related to the previous comment - this is easier to read:
```suggestion
				_, err := hostAwait.UpdateSpace(t, user.Space.Name, func(s *toolchainv1alpha1.Space) {
					if s.Annotations == nil {
						s.Annotations = make(map[string]string)
					}
					s.Annotations[toolchainv1alpha1.FeatureToggleNameAnnotationKey] = "test-feature"
```

mfrancisc (commented):
Looks great 👍 

I have only one minor comment.

I was also wondering if we are planning to add more scenarios later:
- more features
- different weights 
- different objects per feature

but I guess for now this is enough , since we are covered with the unit tests.

test/e2e/parallel/nstemplatetier_test.go
```diff
@@ -303,3 +307,132 @@ func TestKsctlGeneratedTiers(t *testing.T) {
...
+			Email("featured@domain.com").
+			ManuallyApprove().
+			TargetCluster(awaitilities.Member1()).
+			EnsureMUR().
+			SpaceTier("base1ns").
+			TargetCluster(memberAwait).
+			RequireConditions(wait.ConditionSet(wait.Default(), wait.ApprovedByAdmin())...).
+			Execute(t)
+		// and promote that space to the ftier tier
+		tiers.MoveSpaceToTier(t, hostAwait, "featured-user", tier.Name)
```
mfrancisc, line 344: should we create a space with the base1ns tier and check that is doesn't have the object expected for the ftier ?

test/e2e/parallel/nstemplatetier_test.go
```diff
@@ -303,3 +307,132 @@ func TestKsctlGeneratedTiers(t *testing.T) {
...
+	modifiers := []tiers.TierTemplateModifier{
+		func(awaitility *wait.HostAwaitility, template *toolchainv1alpha1.TierTemplate) error {
+			clusterRB := runtime.RawExtension{
+				Raw: tpl.Bytes(),
+			}
+			template.Spec.Template.Objects = append(template.Spec.Template.Objects, clusterRB)
+			return nil
+		},
+	}
+	return tiers.WithClusterResources(t, otherTier, modifiers...)
```
MatousJobanek, line 414: no need to define it as an array
```suggestion
	modifier := func(awaitility *wait.HostAwaitility, template *toolchainv1alpha1.TierTemplate) error {
		clusterRB := runtime.RawExtension{
			Raw: tpl.Bytes(),
		}
		template.Spec.Template.Objects = append(template.Spec.Template.Objects, clusterRB)
		return nil
	}
	return tiers.WithClusterResources(t, otherTier, modifier)
```

test/e2e/parallel/nstemplatetier_test.go
alexeykazakov: NSTemplateSet annotation is an implementation detail. We could check it but IMO it's not necessary in e2e-tests. We just want to check that the object for the enabled feature was indeed created.

testsupport/tiers/checks.go
```diff
@@ -37,6 +37,7 @@ const (
 	baseextendedidling = "baseextendedidling"
 	baselarge          = "baselarge"
 	testTier           = "test"
+	featuredTier       = "ftier" // dynamic tier created by nstemplatetier test to test feature toggles. It's a copy of the "base1ns" tier with additional resources managed by feature toggles. And since it's a copy of another tier we need to keep the name short to avoid exceeding 63 characters for the tier template names
```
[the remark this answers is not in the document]
alexeykazakov: Originally I used the custom tier when provisioned a user:
```
NewSignupRequest(...).SpaceTier(customTier)
```
And it required defining that custom dynamic tier in checks.go. But it turned out not working properly anyway, so I switched to provisioning a space for a built-in tier and then promote it to the custom tier. This dynymic tier definition is a leftover. Let me clean it up now.

test/e2e/parallel/nstemplatetier_test.go
alexeykazakov: https://github.com/codeready-toolchain/toolchain-e2e/pull/1008/commits/fb48abfece5bc03bcf8d5ac2a5d2e937449d723b

test/e2e/parallel/nstemplatetier_test.go
alexeykazakov: https://github.com/codeready-toolchain/toolchain-e2e/pull/1008/commits/4eca4186f0b10e629b7181efa42fb14b5d638ab9

test/e2e/parallel/nstemplatetier_test.go
alexeykazakov: Right. See https://github.com/codeready-toolchain/toolchain-e2e/pull/1008/commits/bb43234bee995e327a3789b83b8512d9822a2a03

testsupport/wait/member.go
alexeykazakov: This is just a theoretical risk in this specific case, isn't it? This specific function just gets the object by this name and if found it returns it. It doesn't check anything else like labels or conditions which would populate the given obj with some found object but instead of returning it would try to populate it again in a loop. This function should work as is but I agree that it's error prone in case we want to change it (and introduce criteria params for example) in the future.
Creating a deep copy would be safer but I'm not sure if there is any easy way of creating a deep copy of a client.Object. We could try to marshal/unmarshal it but not sure if it's worth doing...
Alternatively we could drop it and just add a `WaitForClusterResourceBinding()` function instead but I still find this generic `WaitForObject()` useful.
WDYT?

alexeykazakov:
> I was also wondering if we are planning to add more scenarios later:
> 
> more features
> different weights
> different objects per feature
> but I guess for now this is enough , since we are covered with the unit tests.

All these cases are covered by unit tests. We could have multiple features in e2e-tests too but having any weight which is not 0 or 100 would be quite tricky to test properly. So I just wanted to cover the mechanism of enabling and disabling a feature in the e2e-tests and leave all possible combinations to unit tests.

testsupport/tiers/checks.go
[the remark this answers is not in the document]
alexeykazakov: https://github.com/codeready-toolchain/toolchain-e2e/pull/1008/commits/57aeb2d94e18c4c20124931a96643bae83050448

test/e2e/parallel/nstemplatetier_test.go
alexeykazakov: We could but wouldn't it be an overkill? The object is defined in the custom tier only. We don't check if objects in appstudio tiers are not ended up in spaces with base1ns tier, for example.
What we could test here is promoting a space with tier `featuredTier` to `base1ns` and see that the object is deleted. But I wanted to keep the e2e-tests low-profile and test the main scenario: enable a feature, disable a feature. And leave all the corner cases to unit tests.
Do you think we should have a test for the promotion case in e2e-test anyway?

alexeykazakov:
@mfrancisc @MatousJobanek thank you for the review! I addressed or replied to all the comments. Please take another look when you have time.

testsupport/tiers/checks.go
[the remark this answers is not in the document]
alexeykazakov: OK. I faced the same issue with `VerifyResourcesProvisionedForSpaceWithCustomTier()`. The problem is that exiting tests do not really modify the templates themselves. They create copies of existing tier templates as is.

Modifying the templates after they are already duplicated is not working because such templates (duplicates) already exist: https://github.com/codeready-toolchain/toolchain-e2e/blob/master/testsupport/tiers/tier_setup.go#L143

It is still possible to modify an existing template by creating a new one (using the custom template as the original one, so we would duplicate a template of the existing custom template). However in this case the template.Spec.Tier would refer to the custom tier, so checks.go can't find the check for it.

This change should fix the problem: https://github.com/codeready-toolchain/toolchain-e2e/pull/1008/commits/ddc4d6a1b2907ca085f6d5685499330bcc078b27
It basically allows to provide the template modifiers during the custom tier creation.

testsupport/wait/member.go
MatousJobanek: In your specific case, yes. But this is a generic function that could be used for anything and you cannot control how the caller uses the function. 
I found an old PR where we were fixing this problem https://github.com/codeready-toolchain/toolchain-e2e/pull/43 and I guess that there were a few more cases since then.
Anyway, why couldn't we use the already existing generic functions? https://github.com/codeready-toolchain/toolchain-e2e/blob/82dc979b1fafe3bf61792736cad45b71baead750/testsupport/wait/awaitility.go#L657-L845
Instead of calling:
```go
crb := &v1.ClusterRoleBinding{}
crbName := fmt.Sprintf("%s-%s", user.Space.Name, "test-feature")
err = memberAwait.WaitForObject(t, "", crbName, crb, &v1.ClusterRoleBindingList{})
require.NoError(t, err)
```
you could do
```go
crbName := fmt.Sprintf("%s-%s", user.Space.Name, "test-feature")
crb, err := wait.For(t, memberAwait.Awaitility, &v1.ClusterRoleBinding{}).WithNameThat(crbName)
require.NoError(t, err)
```
or create a new method that would accept also the namespace because the current one automatically takes the operator's namespace - something like that:
```go
crbName := fmt.Sprintf("%s-%s", user.Space.Name, "test-feature")
crb, err := wait.For(t, memberAwait.Awaitility, &v1.ClusterRoleBinding{}).WithNamespacedNameThat("", crbName)
require.NoError(t, err)
```

TBH, I would rather make this generic `Waiter` more powerful and useful than creating yet another function that does more or less the same, but in a slightly risky way and less powerful way. I understand that the `Waiter` may look complicated and doesn't provide the nice logging as the other wait functions, but take it as the start. We can make it better. Mainly the predicate abstraction could have been done in a more useful way.

MatousJobanek (commented):
It looks good, but please consider using the already existing generic Waiter type and its methods and expanding it if needed

test/e2e/parallel/nstemplatetier_test.go
mfrancisc: > We could but wouldn't it be an overkill? The object is defined in the custom tier only. We don't check if objects in appstudio tiers are not ended up in spaces with base1ns tier, for example.

that's right, we can avoid this indeed 👍 

> Do you think we should have a test for the promotion case in e2e-test anyway?

If we think it's a scenario that would be frequent yes, if it's just a corner case - I mean if we will not perform this promotion from featured tier to non featured tiers often then it's not worth adding e2e tests ( increasing overall execution time and adding more code to maintain ) , unit tests covering this would be enough indeed. I don't think we will move users between these tiers often , maybe it's fine to leave it for later if needed 👍

testsupport/tiers/tier_setup.go
```diff
@@ -29,11 +29,13 @@ type CustomNSTemplateTier struct {
 
 type CustomNSTemplateTierModifier func(*HostAwaitility, *CustomNSTemplateTier) error
 
-func WithClusterResources(t *testing.T, otherTier *toolchainv1alpha1.NSTemplateTier) CustomNSTemplateTierModifier {
+type TierTemplateModifier func(*HostAwaitility, *toolchainv1alpha1.TierTemplate) error
```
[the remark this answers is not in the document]
alexeykazakov: Yes, you are right. We don't need it. Cleaned it up here: https://github.com/codeready-toolchain/toolchain-e2e/pull/1008/commits/f964109cb7dcadde48ec94b57b381f6bc244c9e0

testsupport/wait/member.go
alexeykazakov: > In your specific case, yes. But this is a generic function that could be used for anything and you cannot control how the caller uses the function.
I found an old PR where we were fixing this problem https://github.com/codeready-toolchain/toolchain-e2e/pull/43 and I guess that there were a few more cases since then.

It's not actually about how this function can be used but more about how it's implemented. The PR/fix you provided is a bit different because, as I mentioned above, the function accepts criteria. So the function loads the object multiple times in the same struct until it satisfies the criteria. 

But the proposed `WaitForObject()` function does not accept any criteria. So it doesn't load the struct multiple times. As soon as the object is found it returns it as is. Previous attempts should not affect it. The only potential issue can be if you pass non-empty object to that function. But it's no different to using `client.Get(object)`.

The potential risk I was referring to is related to situation if we decide to add some criteria to the function later without taking into account the implementation details. In this case we would introduce a bug into the function.

But anyway. I was not aware of that `wait.For()` function and this is why I introduced `WaitForObject()`. So let me drop this and re-use the existing function: https://github.com/codeready-toolchain/toolchain-e2e/pull/1008/commits/df30805f96d33bf08e9d2eff8bad4ffc1d4334d0
I also added `wait.For().WithNameDeleted()` instead of `WaitForObjectToBeDeleted()`.

testsupport/wait/awaitility.go
```diff
@@ -796,6 +796,41 @@ func (w *Waiter[T]) WithNameThat(name string, predicates ...assertions.Predicate
...
+func (w *Waiter[T]) WithNameDeleted(name string) error {
+	w.t.Logf("waiting for object of GVK '%s' with name '%s' in namespace '%s' to be deleted", w.gvk, name, w.await.Namespace)
+	err := wait.Poll(w.await.RetryInterval, w.await.Timeout, func() (done bool, err error) {
+		obj := &unstructured.Unstructured{}
+		obj.SetGroupVersionKind(w.gvk)
+		if err := w.await.Client.Get(context.TODO(), client.ObjectKey{Name: name, Namespace: w.await.Namespace}, obj); err != nil {
+			if apierrors.IsNotFound(err) {
+				return true, nil
+			}
+			return false, err
```
rajivnathan, line 809: If there is a non-nil error returned here then it will stop retrying right? Maybe it's better to ignore the error to reduce the chance of test flakiness in cases of connection issues or some other temporary issue? It should still fail with a timeout if the object cannot be deleted due to other issues. WDYT? Also, if we do that we can log the error for debugging purposes.

test/e2e/parallel/nstemplatetier_test.go
```diff
@@ -303,3 +307,126 @@ func TestKsctlGeneratedTiers(t *testing.T) {
...
+
+		// Now let's create a Space
+		user := NewSignupRequest(awaitilities).
+			Username("featured-user").
+			Email("featured@domain.com").
+			ManuallyApprove().
+			TargetCluster(awaitilities.Member1()).
+			EnsureMUR().
+			SpaceTier("base1ns").
+			TargetCluster(memberAwait).
```
rajivnathan, line 342: Looks like `TargetCluster` is set twice?

rajivnathan (commented):
Nice! Just a couple minor comments

testsupport/wait/awaitility.go
alexeykazakov: Well.. theoretically I guess you are right. But I don't remember a single case of flaky test because of a connection issue or anything like that. But maybe you've seen it? Returning an error here would make it easier to debug the problem because it would be in the failed test message. It can be hard to look for that error in logs.
And we do it everywhere (or at least in most cases). So I would still keep it as is, so it's consistent at least if not for other reasons.

test/e2e/parallel/nstemplatetier_test.go
alexeykazakov: Yep. Good catch. Cleaned up here: https://github.com/codeready-toolchain/toolchain-e2e/pull/1008/commits/b7288e49915234b9c3bda7290349c8aae90bef01

testsupport/wait/member.go
MatousJobanek: you are actually right, the for loop would immediately stop when the resources is present, so there wouldn't be multiple Get calls that would get different versions of the resources. I had the mindset of multiple executions waiting from some criteria, sorry for confusion :+1: 

Anyway, thanks for reusing the `wait.For()` capability :+1:

MatousJobanek (approved):
Nice and clean :+1: :1st_place_medal: 
Thanks :rocket:

mfrancisc (approved):
Nice Job 👍

rajivnathan (approved):
Nice! 🚀

PR #1009, opened 2024-07-11
add tests for Public-Viewer

filariow:
This PR adds the e2e tests for Public-Viewer 

PR linked to https://github.com/codeready-toolchain/registration-service/pull/443

JIRA [ASC-532](https://issues.redhat.com/browse/ASC-532)


test/e2e/public-viewer/proxy_test.go
```diff
@@ -0,0 +1,181 @@
...
+		identityID:            uuid.New(),
+	}
+	createAppStudioUser(t, awaitilities, communityUser)
+
+	communityUserProxyClient, err := hostAwait.CreateAPIProxyClient(t, communityUser.token, hostAwait.APIProxyURL)
+	require.NoError(t, err)
+
+	t.Run("space is flagged as community", func(t *testing.T) {
+		// when
+		_ = CreateCommunitySpaceBinding(t, hostAwait, space.Name, space.Namespace)
```
rsoaresd, line 56:
```suggestion
                err = CreateCommunitySpaceBinding(t, hostAwait, space.Name, space.Namespace)
```

rsoaresd (dismissed):
Hellooo! Nice 🚀 

Just a minor suggestion:

test/e2e/public-viewer/proxy_test.go
filariow: thanks, Rafaela! I fixed these lines in 86d96ccf3eff37007feb199ae383090c7ca3f338

test/e2e/publicviewer/proxy_test.go
```diff
@@ -0,0 +1,181 @@
...
+// tests access to community-shared spaces
+func TestPublicViewerProxy(t *testing.T) {
+	// given
+
+	// make sure everything is ready before running the actual tests
+	awaitilities := WaitForDeployments(t)
+	hostAwait := awaitilities.Host()
+	memberAwait := awaitilities.Member1()
+	// we create a space to share , a new MUR and a SpaceBindingRequest
+	space, _, _ := NewSpaceBindingRequest(t, awaitilities, memberAwait, hostAwait, "admin")
```
mfrancisc, line 42: why do you need to create the space using spacebinding request function ? It doesn't seem you use the SBR returned by this function, but maybe I'm missing something ? 

If you need only the Space you could use something like:
```
space, signup, _ := testsupportspace.CreateSpace(t, awaitilities, testspace.WithTierName("appstudio"), testspace.WithSpecTargetCluster(memberAwait.ClusterName))
```

test/e2e/publicviewer/spacebindingrequest_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+	testspace "github.com/codeready-toolchain/toolchain-common/pkg/test/space"
+	spacebindingrequesttestcommon "github.com/codeready-toolchain/toolchain-common/pkg/test/spacebindingrequest"
+	. "github.com/codeready-toolchain/toolchain-e2e/testsupport"
+	testsupportspace "github.com/codeready-toolchain/toolchain-e2e/testsupport/space"
+	"github.com/codeready-toolchain/toolchain-e2e/testsupport/spacebinding"
+	"github.com/codeready-toolchain/toolchain-e2e/testsupport/wait"
+)
+
+func NewSpaceBindingRequest(
```
mfrancisc, line 22: just curious - was there any problem with reusing the existing function from https://github.com/codeready-toolchain/toolchain-e2e/blob/82dc979b1fafe3bf61792736cad45b71baead750/test/e2e/parallel/spacebindingrequest_test.go#L231

Plus as mentioned in my other comment you probably don't need this function, but maybe I'm missing something.

test/e2e/publicviewer/proxy_test.go
```diff
@@ -0,0 +1,181 @@
+package publicviewer_test
```
mfrancisc, line 1: as mentioned in the comment for the `test.mk` file maybe you don't need to have a separate test suite, you could move this under the test/e2e folder ( not the parallel one maybe ), and update the toolchainconfig at runtime.

mfrancisc (commented):
Overall the tests are looking good! Great Job 👍 

I have left few comments.

mfrancisc:
Also I think we should have a test case for the scenario in which a community Space is flagged as non community and the `KubesawAuthenticatedUsername` looses access. WDYT?

alexeykazakov (commented):
Could you please add the link to the paired PR to the description?

test/e2e/publicviewer/proxy_test.go
filariow: great catch!

make/test.mk
```diff
@@ -138,6 +148,25 @@ test-e2e-host-local:
 test-e2e-registration-local:
 	$(MAKE) test-e2e REG_REPO_PATH=${PWD}/../registration-service
 
+.PHONY: e2e-run-public-viewer
+e2e-run-public-viewer:
+	@echo "Running e2e public-viewer tests..."
+	oc patch toolchainconfigs.toolchain.dev.openshift.com config \
+		-n ${HOST_NS} \
+		--patch='{"spec":{"host":{"publicViewerConfig":{"enabled":true}}}}' \
+		--type=merge
```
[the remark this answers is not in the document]
filariow: done in c4660e0332eb71ee3bec7645110ea5df8869ccb9

test/e2e/publicviewer/proxy_test.go
filariow: done in c4660e0332eb71ee3bec7645110ea5df8869ccb9

test/e2e/publicviewer/spacebindingrequest_test.go
filariow: done in c4660e0332eb71ee3bec7645110ea5df8869ccb9

test/e2e/publicviewer/proxy_test.go
filariow: done in c4660e0332eb71ee3bec7645110ea5df8869ccb9

filariow:
> Also I think we should have a test case for the scenario in which a community Space is flagged as non community and the `KubesawAuthenticatedUsername` looses access. WDYT?

Interesting idea. This basically means that the CommunitySpaceBinding is not there (i.e., it has been deleted). For a normal sso-user I expect this test to already exist (maybe in the SpaceBinding suite?). For the special user `KubesawAuthenticatedUsername` I'd say it's worth to add a test!

test/e2e/proxy_publicviewer_test.go
```diff
@@ -0,0 +1,225 @@
...
+				err := joeCli.Create(context.TODO(), &cm)
+				require.Error(t, err)
+			})
+		})
+	})
+
+	t.Run("space is not flagged as community", func(t *testing.T) {
+		sp := toolchainv1alpha1.Space{}
+		err = hostAwait.Client.Get(context.TODO(), client.ObjectKeyFromObject(space), &sp)
+		require.Error(t, err)
```
filariow, line 163:
```suggestion
		require.NoError(t, err)
```

filariow:
```
 No resources found
reached timeout of waiting for CRD toolchainconfigs.toolchain.dev.openshift.com to be available in the cluster and the InstallPlan to be complete - see following info for debugging: 
```

/retest

test/e2e/proxy_publicviewer_test.go
```diff
@@ -0,0 +1,225 @@
...
+					},
+					client.InNamespace(space.Namespace),
+				}
+				sbs := &toolchainv1alpha1.SpaceBindingList{}
+				if err := hostAwait.Client.List(context.TODO(), sbs, opts...); err != nil {
+					return false, err
+				}
+
+				return len(sbs.Items) == 1, nil
+			}))
```
mfrancisc, line 80: you could use
```suggestion
				require.NoError(t, err)
				_, err = awaitilities.Host().WaitForSpaceBinding(t, toolchainv1alpha1.KubesawAuthenticatedUsername, space.Name,
					UntilSpaceBindingHasSpaceRole("viewer"))
				require.NoError(t, err)
```
this way you also check the expected role.

test/e2e/proxy_publicviewer_test.go
```diff
@@ -0,0 +1,225 @@
...
+
+			t.Run("community user cannot create config maps into space", func(t *testing.T) {
+				cm := corev1.ConfigMap{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "test-cm",
+						Namespace: sp.Status.ProvisionedNamespaces[0].Name,
+					},
+				}
+				err := communityUserProxyClient.Create(context.TODO(), &cm)
+				require.Error(t, err)
```
mfrancisc, line 115: minor - but maybe we could check that the returned error is the expected one using `require.EqualError`

test/e2e/proxy_publicviewer_test.go
```diff
@@ -0,0 +1,225 @@
...
+			t.Run("sso user cannot create config maps into space", func(t *testing.T) {
+				// then
+				cm := corev1.ConfigMap{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "test-cm",
+						Namespace: sp.Status.ProvisionedNamespaces[0].Name,
+					},
+				}
+				err := joeCli.Create(context.TODO(), &cm)
+				require.Error(t, err)
```
mfrancisc, line 155: same as above - if possible I would check for the specific error

test/e2e/proxy_publicviewer_test.go
```diff
@@ -0,0 +1,225 @@
...
+
+			t.Run("community user cannot list config maps from community space", func(t *testing.T) {
+				// then
+				cms := corev1.ConfigMapList{}
+
+				communityUserProxyClient, err := hostAwait.CreateAPIProxyClient(t, communityUser.token, hostAwait.ProxyURLWithWorkspaceContext(sp.Name))
+				require.NoError(t, err)
+
+				err = communityUserProxyClient.List(context.TODO(), &cms, client.InNamespace(sp.Status.ProvisionedNamespaces[0].Name))
+				require.Error(t, err)
```
mfrancisc, line 183: same here if possible

test/e2e/proxy_publicviewer_test.go
```diff
@@ -0,0 +1,225 @@
...
+
+			t.Run("community user cannot create config maps into space", func(t *testing.T) {
+				cm := corev1.ConfigMap{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "test-cm",
+						Namespace: sp.Status.ProvisionedNamespaces[0].Name,
+					},
+				}
+				err := communityUserProxyClient.Create(context.TODO(), &cm)
+				require.Error(t, err)
```
mfrancisc, line 194: and here

test/e2e/proxy_publicviewer_test.go
```diff
@@ -0,0 +1,225 @@
...
+		Username(user.username).
+		IdentityID(user.identityID).
+		ManuallyApprove().
+		TargetCluster(user.expectedMemberCluster).
+		EnsureMUR().
+		RequireConditions(wait.ConditionSet(wait.Default(), wait.ApprovedByAdmin())...).
+		Execute(t)
+	user.signup = req.UserSignup
+	user.token = req.Token
+	tiers.MoveSpaceToTier(t, awaitilities.Host(), user.signup.Status.CompliantUsername, "appstudio")
```
mfrancisc, line 212: you could create the usersignup with the appstudio tier
```suggestion
	req := NewSignupRequest(awaitilities).
		Username(user.username).
		IdentityID(user.identityID).
		ManuallyApprove().
		TargetCluster(user.expectedMemberCluster).
		EnsureMUR().
		RequireConditions(wait.ConditionSet(wait.Default(), wait.ApprovedByAdmin())...).
		Execute(t)
         req := NewSignupRequest(awaitilities).
		Username(user.username).
		IdentityID(user.identityID).
		ManuallyApprove().
		TargetCluster(user.expectedMemberCluster).
		EnsureMUR().
		SpaceTier("appstudio").
		RequireConditions(ConditionSet(wait.Default(), wait.ApprovedByAdmin())...).
		Execute(t)
	user.signup = req.UserSignup
	user.token = req.Token
```

test/e2e/proxy_publicviewer_test.go
```diff
@@ -0,0 +1,225 @@
...
+	VerifyResourcesProvisionedForSignup(t, awaitilities, user.signup, "deactivate30", "appstudio")
+	user.compliantUsername = user.signup.Status.CompliantUsername
+	_, err := awaitilities.Host().WaitForMasterUserRecord(t, user.compliantUsername, wait.UntilMasterUserRecordHasCondition(wait.Provisioned()))
+	require.NoError(t, err)
```
mfrancisc, line 216: also I think this is already covered by the `EnsureMUR()` option in the signup creation right ? 
```suggestion
```

mfrancisc (commented):
Looks good! Thanks for the simplifications!

I only have few minor comments.

test/e2e/proxy_publicviewer_test.go
filariow: If I'm getting it correctly, `EnsureMUR` is not checking that the MUR has the Provisioned condition

test/e2e/proxy_publicviewer_test.go
filariow: done in 8d2872804d812e52a62d4144e4451b21db20e798

test/e2e/proxy_publicviewer_test.go
filariow: done in 8d2872804d812e52a62d4144e4451b21db20e798

test/e2e/proxy_publicviewer_test.go
filariow: done in 8d2872804d812e52a62d4144e4451b21db20e798

test/e2e/proxy_publicviewer_test.go
filariow: done in 8d2872804d812e52a62d4144e4451b21db20e798

test/e2e/proxy_publicviewer_test.go
filariow: done in 8d2872804d812e52a62d4144e4451b21db20e798

test/e2e/proxy_publicviewer_test.go
filariow: done in 8d2872804d812e52a62d4144e4451b21db20e798

test/e2e/proxy_publicviewer_test.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+func createAppStudioUser(t *testing.T, awaitilities wait.Awaitilities, user *proxyUser) {
+	// Create and approve signup
+	req := NewSignupRequest(awaitilities).
+		Username(user.username).
+		IdentityID(user.identityID).
+		ManuallyApprove().
+		TargetCluster(user.expectedMemberCluster).
+		EnsureMUR().
```
mfrancisc, line 192: minor optimization - If I'm not mistaken you don't need a space to be created for this user right ? . 
```suggestion
		EnsureMUR().
		NoSpace().
```

test/e2e/proxy_publicviewer_test.go
```diff
@@ -0,0 +1,210 @@
...
+	hostAwait.UpdateToolchainConfig(t, testconfig.PublicViewerConfig(true))
+
+	// we create a space to share and a user for 'community-user'
+	space, _, _ := testsupportspace.CreateSpace(t, awaitilities, testspace.WithTierName("appstudio"), testspace.WithSpecTargetCluster(memberAwait.ClusterName))
+	communityUser := &proxyUser{
+		expectedMemberCluster: memberAwait,
+		username:              "community-user",
+		identityID:            uuid.New(),
+	}
+	createAppStudioUser(t, awaitilities, communityUser)
```
mfrancisc, line 54: minor - since apparently this function is used only for creating the community user maybe we can rename it to `createAppStudioCommunityUser` ? feel free to leave it as is in case there might future use-cases that might be different.

testsupport/spacebinding/spacebindingrequest.go
```diff
@@ -34,6 +34,11 @@ func WithNamespace(ns string) SpaceBindingRequestOption {
 func CreateSpaceBindingRequest(t *testing.T, awaitilities wait.Awaitilities, memberName string, opts ...SpaceBindingRequestOption) *toolchainv1alpha1.SpaceBindingRequest {
 	memberAwait, err := awaitilities.Member(memberName)
 	require.NoError(t, err)
+
+	return CreateSpaceBindingRequestOnMember(t, memberAwait, opts...)
+}
+
+func CreateSpaceBindingRequestOnMember(t *testing.T, memberAwait *wait.MemberAwaitility, opts ...SpaceBindingRequestOption) *toolchainv1alpha1.SpaceBindingRequest {
```
mfrancisc, line 41: I think those changes are left overs from when you were using SBRs , but it doesn't harm to have them anyways. So feel free to keep those or drop them.

mfrancisc (approved):
Great job with the tests 👏 🏅 

Thanks for addressing my comments 🙏

testsupport/spacebinding/spacebindingrequest.go
filariow: done in 843f770711828865aff04721ba020e9c7762c48e

test/e2e/proxy_publicviewer_test.go
filariow: done in c00f15122abb7185057104da91dd426921dc3255

test/e2e/proxy_publicviewer_test.go
filariow: good idea, done in c00f15122abb7185057104da91dd426921dc3255

mfrancisc (approved):

mfrancisc:
Just a reminder - please wait for the paired PR to be green before merging, and other's review as well of course.

test/e2e/proxy_publicviewer_test.go
```diff
@@ -0,0 +1,214 @@
...
+		ManuallyApprove().
+		TargetCluster(user.expectedMemberCluster).
+		EnsureMUR().
+		NoSpace().
+		SpaceTier("appstudio").
+		RequireConditions(wait.ConditionSet(wait.Default(), wait.ApprovedByAdmin())...).
+		Execute(t)
+	user.signup = req.UserSignup
+	user.token = req.Token
+	VerifyResourcesProvisionedForSignup(t, awaitilities, user.signup, "deactivate30", "appstudio")
```
mfrancisc, line 200: one more thing - do you need to run those verifications for the community user ? I don't think so, since we are creating this without the space and now the test is failing.
```suggestion
```

filariow: Thanks! I don't think it's needed. I committed your suggestion.

mfrancisc:
@filariow I see the test started to fail with:
```
 proxy_publicviewer_test.go:51: 
        	Error Trace:	/go/src/github.com/codeready-toolchain/toolchain-e2e/test/e2e/proxy_publicviewer_test.go:51
        	            	an error on the server ("unable to get target cluster: no member cluster found for the user") has prevented the request from succeeding
        	Test:       	TestProxyPublicViewer
```

it seems like the proxy returns an error from here , and this seems to be happening since the removal of the creation of the space for the community user : https://github.com/codeready-toolchain/toolchain-e2e/pull/1009/commits/c00f15122abb7185057104da91dd426921dc3255 . Could it be ?

filariow:
> @filariow I see the test started to fail with:
> 
> ```
>  proxy_publicviewer_test.go:51: 
>         	Error Trace:	/go/src/github.com/codeready-toolchain/toolchain-e2e/test/e2e/proxy_publicviewer_test.go:51
>         	Error:      	Received unexpected error:
>         	            	an error on the server ("unable to get target cluster: no member cluster found for the user") has prevented the request from succeeding
>         	Test:       	TestProxyPublicViewer
> ```
> 
> it seems like the proxy returns an error from here , and this seems to be happening since the removal of the creation of the space for the community user : [c00f151](https://github.com/codeready-toolchain/toolchain-e2e/commit/c00f15122abb7185057104da91dd426921dc3255) . Could it be ?

Yes, that's also my understanding. I'm looking into it.

filariow:
@mfrancisc PTAL: https://github.com/codeready-toolchain/toolchain-e2e/pull/1009/commits/002b8acd6635fc3d8cc9965d3237b60a74babbf5

mfrancisc:
@filariow does this mean that the community users will need to have a space or this is temporary rollback only ?

dperaza4dustbit (approved):
/lgtm

test/e2e/proxy_publicviewer_test.go
```diff
@@ -0,0 +1,289 @@
...
+
+		// Wait until space is flagged as community
+		_, err := hostAwait.WaitForSpaceBinding(t, toolchainv1alpha1.KubesawAuthenticatedUsername, space.Name, wait.UntilSpaceBindingHasSpaceRole("viewer"))
+		require.NoError(t, err)
+
+		// wait until the space has ProvisionedNamespaces
+		sp, err := hostAwait.WaitForSpace(t, space.Name, wait.UntilSpaceHasAnyProvisionedNamespaces())
+		require.NoError(t, err)
+
+		t.Run("user is not banned", func(t *testing.T) {
```
mfrancisc, line 141: TBH I'm not sure we should have the subtests for "user is not banned" and "user is banned", since the not approved user and the sso user cannot be banned, or at least I don't think we ban those right now, it's not a real scenario. 

I would maybe just add an entry for the banned user in the `tt` ( testing table ) and verify that it cannot access community spaces.

test/e2e/proxy_publicviewer_test.go
```diff
@@ -0,0 +1,289 @@
...
+							err := proxyClient.Create(context.TODO(), &cm)
+							require.True(t, errors.IsForbidden(err), "expected Create ConfigMap as SSO user to return a Forbidden error, actual: %v", err)
+						})
+					})
+				})
+			}
+		})
+	})
+
+	t.Run("space is not flagged as community", func(t *testing.T) {
```
mfrancisc, line 205: isn't this already covered in the existing https://github.com/codeready-toolchain/toolchain-e2e/blob/58eee93627f35d2d9b67def93d9c72232facbb87/test/e2e/parallel/proxy_test.go ?

filariow: Those tests are testing user access to workspaces when Public-Viewer is disabled. This one is testing user access to private workspaces with Public-Viewer enabled

mfrancisc: Ok so this makes sure that the feature toggle doesn't affect the current functionality for private workspaces. Thanks for explaining that, I've missed it.

mfrancisc: I would maybe just add a check to make sure the `space` doesn't have the `KubesawAuthenticatedUsername` spacebinding before starting with all the other subtests, unless I've missed it , this doesn't seem to be there now.

filariow: good idea, it's not there. Let me add it.

filariow: eea5b2d5df84b7ec82338430ba94e4b8e971bd7e

test/e2e/proxy_publicviewer_test.go
```diff
@@ -0,0 +1,313 @@
...
+						t.Run("cannot create config maps", func(t *testing.T) {
+							cm := corev1.ConfigMap{
+								ObjectMeta: metav1.ObjectMeta{
+									Name:      "test-cm",
+									Namespace: sp.Status.ProvisionedNamespaces[0].Name,
+								},
+							}
+							err := proxyClient.Create(context.TODO(), &cm)
+							require.True(t, errors.IsForbidden(err), "expected Create ConfigMap as SSO user to return a Forbidden error, actual: %v", err)
+						})
```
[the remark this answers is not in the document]
filariow: here we test that the proxy returns `403 - Forbidden` to banned users' `GET` and `CREATE` requests to the endpoint `/api/v1/namespaces/{{sp.Status.ProvisionedNamespaces[0].Name}}/configmap`.
To obtain this result we use a [pre-warmed client](https://github.com/codeready-toolchain/toolchain-e2e/blob/9767104d581561200788d66e0a3d4029ad6e95d5/test/e2e/proxy_publicviewer_test.go#L177) - one that already has retrieved info from `/api`, to perform the requests.

Client initialization error is tested [here](https://github.com/codeready-toolchain/toolchain-e2e/blob/9767104d581561200788d66e0a3d4029ad6e95d5/test/e2e/proxy_publicviewer_test.go#L183-L198). In this scenario we create a new client. Before executing the first request it tries to retrieve the info from `/api`, receiving a `403 - Forbidden`. The subsequent operations will Fail with a `NoMatch` error as the client does not recognize any resource.

Let me add more comments and fix the generic error check at [L204](https://github.com/codeready-toolchain/toolchain-e2e/blob/9767104d581561200788d66e0a3d4029ad6e95d5/test/e2e/proxy_publicviewer_test.go#L204)

test/e2e/proxy_publicviewer_test.go
filariow: With the current implementation of the [ensureUserIsNotBanned Middleware](https://github.com/codeready-toolchain/registration-service/blob/e15d7cc1730b9b07523bd5c91d89250634d360cb/pkg/proxy/proxy.go#L412-L443)  they can be banned. Indeed, the current implementation checks if a `BannedUser` with the same mail it finds in the JWT token exists. This means that we can technically ban not approved and SSO users.

We are not making use of this scenario ATM, but it's supported by the current implementation.

mfrancisc: ok makes sense. Thanks for the explanation. Let's keep them there.

test/e2e/proxy_publicviewer_test.go
```diff
@@ -0,0 +1,313 @@
...
+					// the client needs to be created before the ban,
+					// otherwise it won't initialize properly
+					url := hostAwait.ProxyURLWithWorkspaceContext(sp.Name)
+					proxyClient, err := hostAwait.CreateAPIProxyClient(t, user.Token(), url)
+					require.NoError(t, err)
+
+					banUser(t, hostAwait, user)
+
+					t.Run(s, func(t *testing.T) {
+						t.Run("user can not initialize a new client", func(t *testing.T) {
```
mfrancisc, line 183:
```suggestion
						t.Run("user cannot initialize a new client", func(t *testing.T) {
```

test/e2e/proxy_publicviewer_test.go
```diff
@@ -0,0 +1,313 @@
...
+		testCases := map[string]bool{
+			"user is not banned": false,
+			"user is banned":     true,
+		}
+		for s, tc := range testCases {
+			t.Run(s, func(t *testing.T) {
+				for s, c := range tt {
+					t.Run(s, func(t *testing.T) {
+						user := c.proxyClientUser()
+						if tc {
```
mfrancisc, line 243: minor - maybe we can rename this to something more self explanatory like `banUser` , `isBanned` or something similar. Just to make it easier to read.

test/e2e/proxy_publicviewer_test.go
```diff
@@ -0,0 +1,313 @@
...
+							banUser(t, hostAwait, user)
+						}
+
+						t.Run("community user cannot access to non-community space", func(t *testing.T) {
```
mfrancisc, line 247: minor - why defining the user as "community user" ? Those are just "regular" users right ? They are created with this function `createAppStudioCommunityUser` which creates a `UserSignup` and `Space` resource right ?
```suggestion
						t.Run("user cannot access to non-community space", func(t *testing.T) {
```

maybe we can rename those and remove the community part.

test/e2e/proxy_publicviewer_test.go
```diff
@@ -0,0 +1,313 @@
...
+							require.NotEmpty(t, sp.Status.ProvisionedNamespaces)
+
+							proxyWorkspaceURL := hostAwait.ProxyURLWithWorkspaceContext(sp.Name)
+							communityUserProxyClient, err := hostAwait.CreateAPIProxyClient(t, user.Token(), proxyWorkspaceURL)
+							require.NoError(t, err)
+
+							t.Run("community user cannot list config maps from community space", func(t *testing.T) {
```
mfrancisc, line 254: shouldn't this be "non-community space" ?
```suggestion
							t.Run("user cannot list config maps from non-community space", func(t *testing.T) {
```

test/e2e/proxy_publicviewer_test.go
```diff
@@ -0,0 +1,313 @@
...
+								cms := corev1.ConfigMapList{}
+
+								err = communityUserProxyClient.List(context.TODO(), &cms, client.InNamespace(sp.Status.ProvisionedNamespaces[0].Name))
+								require.True(t, meta.IsNoMatchError(err), "expected List ConfigMap as community user to return a NoMatch error, actual: %v", err)
+							})
+
+							t.Run("community user cannot create config maps into space", func(t *testing.T) {
```
mfrancisc, line 261: also here, to be consistent
```suggestion
							t.Run("user cannot create config maps into non-community space", func(t *testing.T) {
```

mfrancisc (approved):
Great Job with creating those e2e tests 🏅

test/e2e/proxy_publicviewer_test.go
filariow: right, I also renamed the auxiliary function `createAppStudioCommunityUser`: 213786f928afe7d19746e789077523dbf18ee118

test/e2e/proxy_publicviewer_test.go
filariow: done 9213645a5bbc6c7e2b3bebaa528d2abe05f5ecde

PR #1010, opened 2024-07-12
install ksctl via go install in a separate file

MatousJobanek:
simplified the installation of the ksctl binary and moved it to a separate makefile to not mix the things in the huge test file.
Later, we can think about:
1. either incorporating ksctl into the pairing mechanism
2. or specify the ksctl dependency in the go.mod file

But for now, let's do this simple install process and use always the master branch. In this way, we can also easily point any PR to a specific fork/branch if needed (to test any fixes of bugs).

The new `ksctl` makefile target can be also used later for making sure that the binary is present for example before calling `ksctl adm register-member` command.


rajivnathan (approved):
Nice!

alexeykazakov (approved):

PR #1012, opened 2024-07-12
Check that ToolchainCluster status is populated

metlos:
This adds a simple test that the host-operator adds the information about the cluster connection to the ToolchainCluster status. 

In subsequent PR, this information will also be added by the member-operator but we split this into 2 phases to conform to the requirements of the CI that ensure that we maintain backwards compatibility.

Related PRs:
- host-operator: https://github.com/codeready-toolchain/host-operator/pull/1058
- toolchain-common: https://github.com/codeready-toolchain/toolchain-common/pull/412
- toolchain-common: https://github.com/codeready-toolchain/toolchain-common/pull/426


test/e2e/toolchaincluster_test.go
```diff
@@ -55,6 +61,26 @@ func verifyToolchainCluster(t *testing.T, await *wait.Awaitility, otherAwait *wa
...
 	})
 
+	t.Run("status is populated with info from kubeconfig", func(t *testing.T) {
+		if !tcStatusShouldBePopulated {
+			t.Skip("not applicable for this cluster type")
+			return
+		}
+		secret := corev1.Secret{}
+		require.NoError(t, await.Client.Get(context.TODO(), client.ObjectKey{Name: current.Spec.SecretRef.Name, Namespace: current.Namespace}, &secret))
+		apiConfig, err := clientcmd.Load(secret.Data["kubeconfig"])
```
rsoaresd, line 71: should we ensure that kubeconfig key exists/is not empty in the secret data or isn't necessary?

rsoaresd, line 71: should we ensure that kubeconfig key exists/is not empty in the secret data or isn't necessary?

rsoaresd (commented):
Hellooo, nice 🚀

Just a minor question:

test/e2e/toolchaincluster_test.go
```diff
@@ -55,6 +61,26 @@ func verifyToolchainCluster(t *testing.T, await *wait.Awaitility, otherAwait *wa
...
+
+		configContext, present := apiConfig.Contexts[apiConfig.CurrentContext]
+		require.True(t, present)
+
+		operatorNamespace := configContext.Namespace
+		apiEndpoint := apiConfig.Clusters[configContext.Cluster].Server
+
+		assert.Equal(t, operatorNamespace, current.Status.OperatorNamespace)
+		assert.Equal(t, apiEndpoint, current.Status.APIEndpoint)
+	})
```
MatousJobanek, line 82: it would also maybe make sense to double check that it's the same as in the other fields:
```suggestion
		assert.Equal(t, current.Spec.APIEndpoint, current.Status.APIEndpoint)
		assert.Equal(t, current.Labels["namespace"], current.Status.OperatorNamespace)
	})
```

MatousJobanek (commented):
Looks good, just two small comments

test/e2e/toolchaincluster_test.go
MatousJobanek: this is already verified at previous lines https://github.com/codeready-toolchain/toolchain-e2e/blob/b64e9efef25296c7abaf51ce1bfe3936b9a776cb/test/e2e/toolchaincluster_test.go#L59-L62

rsoaresd: Thanks!!

test/e2e/toolchaincluster_test.go
metlos: I'm not sure this would bring much to the table since the spec field is going away and is not actually used for anything as soon as we start using kubeconfig for the client construction.

test/e2e/toolchaincluster_test.go
```diff
@@ -12,8 +12,10 @@ import (
 	"github.com/codeready-toolchain/toolchain-e2e/testsupport/util"
 	"github.com/codeready-toolchain/toolchain-e2e/testsupport/wait"
 	"github.com/stretchr/testify/require"
+	"gotest.tools/assert"
```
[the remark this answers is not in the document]
metlos: this! happens! every! time! 🤦🏼 Thanks for noticing.

test/e2e/toolchaincluster_test.go
```diff
@@ -107,9 +130,8 @@ func verifyToolchainCluster(t *testing.T, await *wait.Awaitility, otherAwait *wa
 		// given
 		name := generateNewName("new-offline-", current.Name)
 		secretCopy := &corev1.Secret{}
-		wait.CopyWithCleanup(t, await,
-			client.ObjectKey{Name: current.Spec.SecretRef.Name, Namespace: current.Namespace},
-			client.ObjectKey{Name: name, Namespace: current.Namespace}, secretCopy)
+		wait.CopyWithCleanup(t, await, client.ObjectKey{Name: current.Spec.SecretRef.Name, Namespace: current.Namespace},
+			client.ObjectKey{Name: name, Namespace: current.Namespace}, secretCopy, kubeconfig.Modify(t, kubeconfig.ApiEndpoint("https://1.2.3.4:8443")))
```
mfrancisc, line 134: minor thing - unless I've missed something the `ApiEndpoint` from the `secretCopy` kubeconfig is not actually used.

testsupport/kubeconfig/kubeconfig.go
```diff
@@ -0,0 +1,54 @@
+package kubeconfig
```
mfrancisc, line 1: I was wondering if the name is accurate, as the functionality in here seems to be modifying secrets ( actually kubeconfig in secrets) , but I cannot come up with a better name.

mfrancisc (commented):
Nice Job! 

I have only few minor comments.

test/e2e/toolchaincluster_test.go
metlos: Au contraire, the `secretCopy` is stored with a modified kubeconfig that contains an invalid api endpoint. The test then makes sure that the toolchain cluster CR that reference that secret correctly reports the status as not ready, because the api endpoint (obtained from the kubeconfig) is invalid.

testsupport/kubeconfig/kubeconfig.go
metlos: Yeah, I was the same. Along with cache invalidation and off-by-one errors, naming is one of the 2 hard things in CS 😉

test/e2e/toolchaincluster_test.go
mfrancisc: I got confused since also the toolchaincluster resource gets created with the same API endpoint, my bad. Thanks for clarifying that.

Maybe a small comment explaining why are we setting that api endpoint would help , something like: _"set invalid api endpoint so that tc goes into unreachable state"_ or something like that.

mfrancisc (approved):
Nice job 👍 

Thanks for answering my questions/comments.

MatousJobanek (approved):
Looks good :+1:

fbm3307 (approved):
Looks Good 👍

PR #1019, opened 2024-07-26
attempt fix flaky TestAutomaticClusterAssignment

mfrancisc:
Attempt to fix : https://redhat-internal.slack.com/archives/C06MJ2DVBU4/p1721985825426319?thread_ts=1721925588.145229&cid=C06MJ2DVBU4

### DIFF
--- a/test/e2e/space_autocompletion_test.go
+++ b/test/e2e/space_autocompletion_test.go
@@ -117,7 +117,7 @@ func TestAutomaticClusterAssignment(t *testing.T) {
 		require.NoError(t, err)
 		for _, m := range toolchainStatus.Status.Members {
 			if memberAwait1.ClusterName == m.ClusterName {
-				spaceprovisionerconfig.UpdateForCluster(t, hostAwait.Awaitility, memberAwait1.ClusterName, testSpc.MaxNumberOfSpaces(uint(m.SpaceCount)))
+				spaceprovisionerconfig.UpdateForCluster(t, hostAwait.Awaitility, memberAwait1.ClusterName, testSpc.MaxNumberOfSpaces(0)) // temporarily disable provisioning on this cluster
 			} else if memberAwait2.ClusterName == m.ClusterName {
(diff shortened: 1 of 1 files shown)


mfrancisc:
/retest

infra

mfrancisc:
/retest 

infra

test/e2e/space_autocompletion_test.go
```diff
@@ -117,7 +117,7 @@ func TestAutomaticClusterAssignment(t *testing.T) {
 		require.NoError(t, err)
 		for _, m := range toolchainStatus.Status.Members {
 			if memberAwait1.ClusterName == m.ClusterName {
-				spaceprovisionerconfig.UpdateForCluster(t, hostAwait.Awaitility, memberAwait1.ClusterName, testSpc.MaxNumberOfSpaces(uint(m.SpaceCount)))
+				spaceprovisionerconfig.UpdateForCluster(t, hostAwait.Awaitility, memberAwait1.ClusterName, testSpc.Enabled(false)) // temporarily disable provisioning on this cluster
```
MatousJobanek, line 120: this removes the flakiness, but it won't test the actual behavior. This test is there to verify space thresholds and it should stay there. 
For now, I would maybe log out which value the SPC was updated with.
Maybe it's just matter of bad timing when the cache of the host-operator wasn't updated soon enough, we use some bad numbers in the tests, or we have a bug somewhere. 
Anyway, since we are not sure what could be the culprit of these failures, I would log the additional information only.

mfrancisc: Makes sense, I've checked but I cannot see this failure in the last 14days in our repos. For now I've added some more logs that will print out the SPC values before and after an update. See https://github.com/codeready-toolchain/toolchain-e2e/pull/1019/commits/9066794655c8974eb8c79ac5ccec79d7d35db93a

MatousJobanek (approved):
/lgtm

I just figured out (based on the latest failure) what is the reason for the flakiness. I'll open a PR with a fix in host-operator.
But it's fine to merge this one (just in case we face another issue with it)

MatousJobanek:
see https://github.com/codeready-toolchain/host-operator/pull/1074 for the fix

PR #1025, opened 2024-07-31
KUBESAW-156: Use ksctl adm register-member command

fbm3307:
This to replace the add-cluster.sh script and use ksctl register-member command

### DIFF
--- a/make/test.mk
+++ b/make/test.mk
@@ -261,10 +261,8 @@ print-operator-logs:
 
 .PHONY: setup-toolchainclusters
 setup-toolchainclusters:
-	$(MAKE) run-cicd-script SCRIPT_PATH=scripts/add-cluster.sh  SCRIPT_PARAMS="-t member -mn $(MEMBER_NS) -hn $(HOST_NS) ${LETS_ENCRYPT_PARAM}"
-	if [[ ${SECOND_MEMBER_MODE} == true ]]; then $(MAKE) run-cicd-script SCRIPT_PATH=scripts/add-cluster.sh  SCRIPT_PARAMS="-t member -mn $(MEMBER_NS_2) -hn $(HOST_NS) -mm 2 ${LETS_ENCRYPT_PARAM}"; fi
-	$(MAKE) run-cicd-script SCRIPT_PATH=scripts/add-cluster.sh  SCRIPT_PARAMS="-t host   -mn $(MEMBER_NS) -hn $(HOST_NS) ${LETS_ENCRYPT_PARAM}"
-	if [[ ${SECOND_MEMBER_MODE} == true ]]; then $(MAKE) run-cicd-script SCRIPT_PATH=scripts/add-cluster.sh  SCRIPT_PARAMS="-t host   -mn $(MEMBER_NS_2) -hn $(HOST_NS) -mm 2 ${LETS_ENCRYPT_PARAM}"; fi
+	register-member -host-ns $(HOST_NS) -member-ns $(MEMBER_NS)
+	if [[ ${SECOND_MEMBER_MODE} == true ]]; then register-member -host-ns $(HOST_NS) -member-ns $(MEMBER_NS_2); fi
 	echo "Restart host operator pods so it can get the ToolchainCluster CRs while it's starting up".
 	oc delete pods --namespace ${HOST_NS} -l control-plane=controller-manager
 


fbm3307:
looks like a infra failure retesting it 
/test e2e

make/test.mk
```diff
@@ -30,10 +30,13 @@ endif
 
 E2E_TEST_EXECUTION ?= true
 
+LETS_ENCRYPT_PARAM_VALUE := false
+
 ifeq ($(IS_OSD),true)
 LETS_ENCRYPT_PARAM := --lets-encrypt
```
mfrancisc, line 36: I think you should set it to true here right? I mean you're not setting it to true anywhere, unless I'm missing something.
```suggestion
LETS_ENCRYPT_PARAM := --lets-encrypt
LETS_ENCRYPT_PARAM_VALUE=true
```

MatousJobanek: or rather this:
```
ifneq ($(IS_OSD),true)
LETS_ENCRYPT_PARAM := --lets-encrypt=false
```

fbm3307: it should be that way.. i was just trying to make it work in the CI.. 
changed it now.

make/test.mk
```diff
@@ -30,10 +30,12 @@ endif
 
 E2E_TEST_EXECUTION ?= true
 
+
 ifeq ($(IS_OSD),true)
```
MatousJobanek, line 34: this has to be 
```suggestion
ifneq ($(IS_OSD),true)
```

make/test.mk
```diff
@@ -260,11 +262,9 @@ print-operator-logs:
...
-	if [[ ${SECOND_MEMBER_MODE} == true ]]; then $(MAKE) run-cicd-script SCRIPT_PATH=scripts/add-cluster.sh  SCRIPT_PARAMS="-t member -mn $(MEMBER_NS_2) -hn $(HOST_NS) -mm 2 ${LETS_ENCRYPT_PARAM}"; fi
-	$(MAKE) run-cicd-script SCRIPT_PATH=scripts/add-cluster.sh  SCRIPT_PARAMS="-t host   -mn $(MEMBER_NS) -hn $(HOST_NS) ${LETS_ENCRYPT_PARAM}"
-	if [[ ${SECOND_MEMBER_MODE} == true ]]; then $(MAKE) run-cicd-script SCRIPT_PATH=scripts/add-cluster.sh  SCRIPT_PARAMS="-t host   -mn $(MEMBER_NS_2) -hn $(HOST_NS) -mm 2 ${LETS_ENCRYPT_PARAM}"; fi
+setup-toolchainclusters: ksctl
+	${BIN_DIR}/ksctl adm register-member --host-ns="$(HOST_NS)" --member-ns="$(MEMBER_NS)" --host-kubeconfig="$(or ${KUBECONFIG}, ${HOME}/.kube/config)" --member-kubeconfig="$(or ${KUBECONFIG}, ${HOME}/.kube/config)" --lets-encrypt="${LETS_ENCRYPT_PARAM}"
+	if [[ ${SECOND_MEMBER_MODE} == true ]]; then ${BIN_DIR}/ksctl adm register-member --host-ns "$(HOST_NS)" --member-ns "$(MEMBER_NS_2)" --host-kubeconfig "$(or ${KUBECONFIG}, ${HOME}/.kube/config)" --member-kubeconfig "$(or ${KUBECONFIG}, ${HOME}/.kube/config)" --lets-encrypt="${LETS_ENCRYPT_PARAM}" ; fi
```
MatousJobanek, line 267: the variable contains the whole flag now
```suggestion
	${BIN_DIR}/ksctl adm register-member --host-ns="$(HOST_NS)" --member-ns="$(MEMBER_NS)" --host-kubeconfig="$(or ${KUBECONFIG}, ${HOME}/.kube/config)" --member-kubeconfig="$(or ${KUBECONFIG}, ${HOME}/.kube/config)" ${LETS_ENCRYPT_PARAM}"
	if [[ ${SECOND_MEMBER_MODE} == true ]]; then ${BIN_DIR}/ksctl adm register-member --host-ns "$(HOST_NS)" --member-ns "$(MEMBER_NS_2)" --host-kubeconfig "$(or ${KUBECONFIG}, ${HOME}/.kube/config)" --member-kubeconfig "$(or ${KUBECONFIG}, ${HOME}/.kube/config)" ${LETS_ENCRYPT_PARAM} ; fi
```

fbm3307: done

make/test.mk
fbm3307: yes i was thinking about the same.. but since it was also being used here.. https://github.com/codeready-toolchain/toolchain-e2e/blob/a5241b5effc70aec391331d8cd44ccffdf801479/make/test.mk#L80
i wasnt still trying to check that it doesnt wont change the usage at the above place..

MatousJobanek: the line only passes the current value of the variable to the called target - to preserve the value. Or what issue do you see with this? If you didn't fix the if statement, then it would work in the opposite way than it is supposed to.

Anyway, what kind of concern did you see with applying the change?

fbm3307: now that we have changed the whole param to `--lets-encrypt=false` instead of just `--lets-encrypt` i was just trying to understand if/how that would still work along with the boolean value we are passing.. 

i have already applied the change .

MatousJobanek:
/retest 
ksctl PR was merged

MatousJobanek (approved):
Looks good, but please double check that the dev and appstudio makefine targets work as expected :+1:

make/test.mk
mfrancisc: This looks fine, even if it relies on the fact that the ksctl command has lets-encrypt set to true by default and it doesn't pass explicitly true. But it's a minor thing.

mfrancisc (approved):
Great job 👍

PR #1030, opened 2024-08-12
enable pairing with branches from ksctl repository

MatousJobanek:
Since the e2e test setup is increasingly dependent on ksctl commands, I introduced a possibility of pairing between toolchain-e2e and ksctl.
* It's only one-directional  - the e2e tests are not executed in ksctl repo for now. It can be changed later.
* It's not included into the limit that we can pair only one PR (host/member/reg-service) with only one PR from toolchain-e2e repo. In other words, it can be combined with it.
* it's not included in the migration test suite for now
* local execution uses ksctl version from master if not specified otherwise via env variables `USE_INSTALLED_KSCTL` or `KSCTL_REPO_PATH`. The UX can be changed later.

paired PR https://github.com/kubesaw/ksctl/pull/58 (expecting to fail when generating tier manifests)

### DIFF
--- a/make/ksctl.mk
+++ b/make/ksctl.mk
@@ -1,7 +1,62 @@
-KSCTL_GH_OWNER=kubesaw
-KSCTL_GH_BRANCH=master
+USE_INSTALLED_KSCTL = false
 
 BIN_DIR := $(shell pwd)/build/_output/bin
 
+ifneq ($(USE_INSTALLED_KSCTL),true)
+KSCTL_BIN_DIR = ${BIN_DIR}/
+endif
+
+.PHONY: ksctl
 ksctl:
-	GOBIN=${BIN_DIR} CGO_ENABLED=0 go install github.com/${KSCTL_GH_OWNER}/ksctl/cmd/ksctl@${KSCTL_GH_BRANCH}
+ifneq ($(USE_INSTALLED_KSCTL),true)
+	@$(MAKE) get-ksctl-and-install --no-print-directory
+else
+	@echo "Using local version of ksctl"
+endif
+
+get-ksctl-and-install:
+ifeq ($(strip $(KSCTL_REPO_PATH)),)
+    ifneq ($(CI),)
+        ifneq ($(GITHUB_ACTIONS),)
+			$(eval BRANCH_NAME = ${GITHUB_HEAD_REF})
+			$(eval AUTHOR_LINK = https://github.com/${AUTHOR})
+        else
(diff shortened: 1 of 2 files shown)


make/ksctl.mk
```diff
@@ -1,7 +1,62 @@
...
+			git config --global user.email "devtools@redhat.com"
+			git config --global user.name "Devtools"
+			# clone
+			git clone --depth=1 https://github.com/kubesaw/ksctl.git ${KSCTL_REPO_PATH}
+			# add the user's fork as remote repo
+			git --git-dir=${KSCTL_REPO_PATH}/.git --work-tree=${KSCTL_REPO_PATH} remote add external ${AUTHOR_LINK}/ksctl.git
+			# fetch the branch
+			git --git-dir=${KSCTL_REPO_PATH}/.git --work-tree=${KSCTL_REPO_PATH} fetch external ${REMOTE_KSCTL_BRANCH}
+			# merge the branch with master
+			git --git-dir=${KSCTL_REPO_PATH}/.git --work-tree=${KSCTL_REPO_PATH} merge --allow-unrelated-histories --no-commit FETCH_HEAD
```
mfrancisc, line 49: just curious - why do we need to clone both the upstream and the fork and merge them ? was there any problem with cloning what was specified in the `KSCTL_GH_OWNER` variable and `KSCTL_GH_BRANCH` and building and installing that ?

mfrancisc: also could there be merge conflicts when doing this ?

mfrancisc: Ok now I see , I think we want it to work in the same way as the pairing for the other repos, so if there is a conflict the author must merge/fix and then rerun the tests. 

Not sure why - but I was thinking that initially we would have a simpler mechanism that just clones a given ksctl repo , builds and installs it.

alexeykazakov (approved):
:+1:

make/ksctl.mk
MatousJobanek: we want to test the final state - when the PR is merged to master. We don't want test only the changes in the PR (without combining them with master) as the PR may be outdated and it doesn't have to contain the latest changes important for the functions of the commands.

MatousJobanek:
/retest
the paired PR was closed and the branch was deleted

mfrancisc (approved):
Looks great 👍  

Thanks a lot for creating this and for answering my comments! 

minor note -  maybe we could document somewhere these variables:
`USE_INSTALLED_KSCTL=true` would use the installed ksctl version
`KSCTL_REPO_PATH=/my/local/path/to/ksctl/` would build your local version of ksctl 


WDYT?

PR #1034, opened 2024-08-15
fix ksctl pairing

MatousJobanek:
* pairing is not executed if `DEPLOY_LATEST` is set to `true` - this means, in appstudio makefile targets
* added more logs so it's clear what version is gonna be used
* extracted git cloning and merging to a separate target - the reason is that the if statements are evaluated at the beginning of the target, so setting it in the middle of it doesn't have any effect on the result of the if statement. Thus, there were problems with the `ifneq ($(strip $(REMOTE_KSCTL_BRANCH)),)`

TBH, the special behavior of makefile targets drives me crazy and I'm thinking about moving it to a bash script.
I'll take a look at the options and think about the best approach.

### DIFF
--- a/make/ksctl.mk
+++ b/make/ksctl.mk
@@ -16,48 +16,61 @@ endif
 
 get-ksctl-and-install:
 ifeq ($(strip $(KSCTL_REPO_PATH)),)
-    ifneq ($(CI),)
-        ifneq ($(GITHUB_ACTIONS),)
-			$(eval BRANCH_NAME = ${GITHUB_HEAD_REF})
-			$(eval AUTHOR_LINK = https://github.com/${AUTHOR})
+    ifneq ($(DEPLOY_LATEST),true)
+        ifneq ($(CI),)
+            ifneq ($(GITHUB_ACTIONS),)
+				$(eval BRANCH_NAME = ${GITHUB_HEAD_REF})
+				$(eval AUTHOR_LINK = https://github.com/${AUTHOR})
+            else
+				$(eval AUTHOR_LINK = $(shell jq -r '.refs[0].pulls[0].author_link' <<< $${CLONEREFS_OPTIONS} | tr -d '[:space:]'))
+				@echo "found author link ${AUTHOR_LINK}"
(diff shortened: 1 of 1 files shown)


sadlerap (approved):
I'm fine with this in a makefile, but I agree with your sentiment - this should maybe get moved to a script at some point.

alexeykazakov (approved):

mfrancisc (approved):
Thanks for fixing it 👍 

I've tried it locally ( forcing the CI variables ) and is working!

mfrancisc:
> TBH, the special behavior of makefile targets drives me crazy and I'm thinking about moving it to a bash script.
I'll take a look at the options and think about the best approach.

Agree , moving it to bash would be better indeed . I wasn't aware of this special behavior of the `if` statement. Thanks for letting us know.

sadlerap:
It's been pointed out to me that `DEPLOY_LATEST` doesn't work for us on workspaces, since we build our own version of kubesaw for testing purposes that may or may not be based off of the latest commit that's landed.

Would it be possible to have some kind of way of disabling PR pairing without requiring the deployment of the latest version of kubesaw?  Maybe have a parameter like `CI_DISABLE_PAIRING` or something similar?

MatousJobanek:
Interesting, I wasn't aware of that. 
In infra-deployments, in the PR that you made https://github.com/redhat-appstudio/infra-deployments/pull/4324, you updated logic that is running the `appstudio-dev-deploy-latest` target. This target automatically sets the `DEPLOY_LATEST` variable to `true` and I cannot find any other place that would do something else :man_shrugging: 

A few people reach out to us that the infra-deployments logic is failing for them because of missing ksctl binary. The reason was your logic that installs the latest ksctl to the default GOBIN directory, which they didn't have added to their PATH. That's why I proposed this change so you can remove the line that installs ksctl to the default GOBIN directory, and toolchain-e2e can use the local temporal location for the binary, and thus infra-deployments setup doesn't cause any other issues when executed locally.

MatousJobanek:
anyway, let me create another PR that should avoid running the pairing logic in other environments

MatousJobanek:
@sadlerap see https://github.com/codeready-toolchain/toolchain-e2e/pull/1041

PR #1036, opened 2024-08-19
add tests for proxy and banned user

filariow:
Signed-off-by: Francesco Ilario <filario@redhat.com>

### DIFF
--- a/test/e2e/parallel/banned_user_test.go
+++ b/test/e2e/parallel/banned_user_test.go
@@ -1,0 +1,80 @@
+package parallel
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
+	testspace "github.com/codeready-toolchain/toolchain-common/pkg/test/space"
+	. "github.com/codeready-toolchain/toolchain-e2e/testsupport"
+	testsupportspace "github.com/codeready-toolchain/toolchain-e2e/testsupport/space"
+	"github.com/codeready-toolchain/toolchain-e2e/testsupport/wait"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestBannedUser(t *testing.T) {
+	t.Parallel()
+
+	// make sure everything is ready before running the actual tests
+	awaitilities := WaitForDeployments(t)
+	hostAwait := awaitilities.Host()
+	memberAwait := awaitilities.Member1()
+
+	// create an user and a space
+	sp, us, _ := testsupportspace.CreateSpaceWithRoleSignupResult(t, awaitilities, "admin",
+		testspace.WithSpecTargetCluster(memberAwait.ClusterName),
+		testspace.WithTierName("appstudio"),
+		testspace.WithStatusProvisionedNamespaces([]toolchainv1alpha1.SpaceNamespace{
+			{},
+		}),
+	)
+
+	// ban the user
+	_ = CreateBannedUser(t, hostAwait, us.UserSignup.Spec.IdentityClaims.Email)
+
+	// wait until the space has ProvisionedNamespaces
+	sp, err := hostAwait.WaitForSpace(t, sp.Name, wait.UntilSpaceHasAnyProvisionedNamespaces())
+	require.NoError(t, err)
+
+	// wait until the user is banned
+	_, err = hostAwait.
+		WithRetryOptions(wait.TimeoutOption(time.Second*10), wait.RetryInterval(time.Second*2)).
+		WaitForUserSignup(t, us.UserSignup.Name,
+			wait.UntilUserSignupHasConditions(
+				wait.ConditionSet(wait.Default(), wait.ApprovedByAdmin(), wait.Banned())...))
+	require.NoError(t, err)
+
+	t.Run("banned user cannot list config maps from space", func(t *testing.T) {
+		// then
+		cms := corev1.ConfigMapList{}
+
+		proxyWorkspaceURL := hostAwait.ProxyURLWithWorkspaceContext(sp.Name)
+		userProxyClient, err := hostAwait.CreateAPIProxyClient(t, us.Token, proxyWorkspaceURL)
+		require.NoError(t, err)
+
+		err = userProxyClient.List(context.TODO(), &cms, client.InNamespace(sp.Status.ProvisionedNamespaces[0].Name))
+		require.True(t, meta.IsNoMatchError(err), "expected List ConfigMap to return a NoMatch error, actual: %v", err)
+	})
+
+	t.Run("banned user cannot create config maps into space", func(t *testing.T) {
+		cm := corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "test-cm",
+				Namespace: sp.Status.ProvisionedNamespaces[0].Name,
+			},
+		}
+		proxyWorkspaceURL := hostAwait.ProxyURLWithWorkspaceContext(sp.Name)
+		communityUserProxyClient, err := hostAwait.CreateAPIProxyClient(t, us.Token, proxyWorkspaceURL)
+		require.NoError(t, err)
+
+		err = communityUserProxyClient.Create(context.TODO(), &cm)
+		require.True(t, meta.IsNoMatchError(err), "expected Create ConfigMap to return a NoMatch error, actual: %v", err)
+	})
+}
--- a/testsupport/space/space.go
+++ b/testsupport/space/space.go
@@ -26,6 +26,13 @@ func CreateSpace(t *testing.T, awaitilities wait.Awaitilities, opts ...testspace
 // CreateSpaceWithRole initializes a new Space object using the NewSpace function, and then creates it in the cluster
 // It also automatically provisions MasterUserRecord and creates SpaceBinding for it
 func CreateSpaceWithRole(t *testing.T, awaitilities wait.Awaitilities, role string, opts ...testspace.Option) (*toolchainv1alpha1.Space, *toolchainv1alpha1.UserSignup, *toolchainv1alpha1.SpaceBinding) {
+	sp, sr, sb := CreateSpaceWithRoleSignupResult(t, awaitilities, role, opts...)
+	return sp, sr.UserSignup, sb
+}
+
+// CreateSpaceWithRole initializes a new Space object using the NewSpace function, and then creates it in the cluster
+// It also automatically provisions MasterUserRecord and creates SpaceBinding for it
+func CreateSpaceWithRoleSignupResult(t *testing.T, awaitilities wait.Awaitilities, role string, opts ...testspace.Option) (*toolchainv1alpha1.Space, *testsupport.SignupResult, *toolchainv1alpha1.SpaceBinding) {
 	// we need to create a MUR & SpaceBinding, otherwise, the Space could be automatically deleted by the SpaceCleanup controller
 	username := uuid.Must(uuid.NewV4()).String()
 	user := testsupport.NewSignupRequest(awaitilities).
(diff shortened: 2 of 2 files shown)


testsupport/space/space.go
```diff
@@ -66,7 +73,7 @@ func CreateSpaceWithRole(t *testing.T, awaitilities wait.Awaitilities, role stri
 		require.NoError(t, err)
 	}
 
-	return space, signup, spaceBinding
+	return space, user, spaceBinding
```
mfrancisc, line 76: is there any benefit for returning the `user` instead of just the `user.UserSignup` if that's the only thing that used? Alternatively I see that the `user` object also contains the space.

testsupport/space/space.go
```diff
@@ -26,6 +26,13 @@ func CreateSpace(t *testing.T, awaitilities wait.Awaitilities, opts ...testspace
 // CreateSpaceWithRole initializes a new Space object using the NewSpace function, and then creates it in the cluster
 // It also automatically provisions MasterUserRecord and creates SpaceBinding for it
 func CreateSpaceWithRole(t *testing.T, awaitilities wait.Awaitilities, role string, opts ...testspace.Option) (*toolchainv1alpha1.Space, *toolchainv1alpha1.UserSignup, *toolchainv1alpha1.SpaceBinding) {
+	sp, sr, sb := CreateSpaceWithRoleSignupResult(t, awaitilities, role, opts...)
+	return sp, sr.UserSignup, sb
+}
+
+// CreateSpaceWithRole initializes a new Space object using the NewSpace function, and then creates it in the cluster
+// It also automatically provisions MasterUserRecord and creates SpaceBinding for it
+func CreateSpaceWithRoleSignupResult(t *testing.T, awaitilities wait.Awaitilities, role string, opts ...testspace.Option) (*toolchainv1alpha1.Space, *testsupport.SignupResult, *toolchainv1alpha1.SpaceBinding) {
```
mfrancisc, line 35: maybe we can rename it to something like `createSignupWithRole` ? And also update the comment.

mfrancisc (commented):
Looks great! Nice job 👏 . 
Thanks a lot for creating this test and filling in this gap 🏅 

I have only few minor comments.

testsupport/space/space.go
filariow: The main reason is to retrieve the token. I'd too prefer not to change this method, but I could not find a better approach. Do you have any suggestions for retrieving it?

testsupport/space/space.go
filariow: I forgot to update the comment, I'll fix that.

This method needs to be exported for being used in`parallel` and it is also creating the space, so the name you suggested won't work. But -I agree- we can find a better name.

mfrancisc: Sorry why does it needs to be exported ? I see the `CreateSpace` is used in the `parallel` test suite, this one is just referenced in this file. But I'm sure I'm missing something. Also it's a minor thing. 

Yeah naming is hard, I don't have a better name. Maybe `CreateSpaceSignupAndBindingWithRole` , or just keep the current name.

mfrancisc: Oh never mind . I'm blind . I saw now you are calling it from the test 🤦‍♂️ . Sorry for the confusion.

testsupport/space/space.go
mfrancisc: Ok makes sense. I've noticed now that is also called from the test. So never mind. Please disregards my comment.

testsupport/space/space.go
mfrancisc: But yeah maybe we can find a different name for the function.

test/e2e/parallel/banned_user_test.go
```diff
@@ -0,0 +1,80 @@
...
+				wait.ConditionSet(wait.Default(), wait.ApprovedByAdmin(), wait.Banned())...))
+	require.NoError(t, err)
+
+	t.Run("banned user cannot list config maps from space", func(t *testing.T) {
+		// then
+		cms := corev1.ConfigMapList{}
+
+		proxyWorkspaceURL := hostAwait.ProxyURLWithWorkspaceContext(sp.Name)
+		userProxyClient, err := hostAwait.CreateAPIProxyClient(t, us.Token, proxyWorkspaceURL)
+		require.NoError(t, err)
```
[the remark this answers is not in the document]
filariow: great catch, done in 8d5a9a6f3359c9d233d965e2662b1368c766d83e

testsupport/space/space.go
filariow: fixed comment here 322e7688618c5ca7f161a920a9a2242d7d9f5fda

test/e2e/parallel/banned_user_test.go
```diff
@@ -0,0 +1,78 @@
...
+	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
+	testspace "github.com/codeready-toolchain/toolchain-common/pkg/test/space"
+	. "github.com/codeready-toolchain/toolchain-e2e/testsupport"
+	testsupportspace "github.com/codeready-toolchain/toolchain-e2e/testsupport/space"
+	"github.com/codeready-toolchain/toolchain-e2e/testsupport/wait"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestBannedUser(t *testing.T) {
```
mfrancisc, line 22: one more thing - since it's specific to the proxy, should we move it to the `proxy_test.go` ?

test/e2e/parallel/banned_user_test.go
```diff
@@ -0,0 +1,78 @@
...
+	// make sure everything is ready before running the actual tests
+	awaitilities := WaitForDeployments(t)
+	hostAwait := awaitilities.Host()
+	memberAwait := awaitilities.Member1()
+
+	// create an user and a space
+	sp, us, _ := testsupportspace.CreateSpaceWithRoleSignupResult(t, awaitilities, "admin",
+		testspace.WithSpecTargetCluster(memberAwait.ClusterName),
+		testspace.WithTierName("appstudio"),
+		testspace.WithStatusProvisionedNamespaces([]toolchainv1alpha1.SpaceNamespace{
```
mfrancisc, line 34: do we need this condition ? I think no, since initially it will be empty and later we wait at line 39 for the provisioned namespaces to be filled.

filariow: I don't remember why we needed but we can try without it and revert if required. Let me remove this line

filariow: done in b6d91539da97de88797fe1994abe954ef9c39a60

test/e2e/parallel/banned_user_test.go
mfrancisc: yeah, renaming sounds good me too! I don't think there is any plan to split it for now. but sounds a good idea. That said we could just rename the test and file so it's clear that is for proxy only and not banned users in general which is covered in the `user_management_test.go`

filariow: done in d3ea52a10ddb8199ed1fbd799b15bc0cf8c13187

alexeykazakov (commented):
Looks good! But shouldn't we merge it with the existing proxy tests?
Similar to how this PR https://github.com/codeready-toolchain/toolchain-e2e/pull/1037 adds the banned user to the space lister tests, this PR could add a banned user in the proxy flow test setup here: https://github.com/codeready-toolchain/toolchain-e2e/blob/master/test/e2e/parallel/proxy_test.go#L189
And then add those two test runs there.

filariow:
> Looks good! But shouldn't we merge it with the existing proxy tests? Similar to how this PR #1037 adds the banned user to the space lister tests, this PR could add a banned user in the proxy flow test setup here: https://github.com/codeready-toolchain/toolchain-e2e/blob/master/test/e2e/parallel/proxy_test.go#L189 And then add those two test runs there.


IIAUIC, these tests checks the workflows for non-banned users, so I don't think it would be a good fit for adding banned user test cases. Am I understanding it correctly?

However the comment above is very insightful, and makes me think we should not use ConfigMap in these tests and use instead the Application.

> // !!! Additional context !!!
> // The test uses a dummy HAS API type called Application. The reason is that the regular
> // user doesn't have full permission for the standard types like ConfigMap. This means
> // that we could do create/read operations on that resource from this test.
> // To work around this limitation, we created a dummy HAS API type that has the same name
> // and the same group as the actual one. The CRD is created as part of the test setup
> // and since the CRD name & group name matches, then RBAC allow us to execute create/read
> // operations on that resource using the user permissions.

test/e2e/parallel/banned_user_test.go
filariow: it passed

mfrancisc:
> However the comment above is very insightful, and makes me think we should not use ConfigMap in these tests and use instead the Application.

TBH Application is something very specific to Konflux, and as we want to decouple more KubeSaw from the products ( Konflux , DevSandbox ) those tests might have to change or be moved to the products repos/org. WDYT?

filariow:
> > However the comment above is very insightful, and makes me think we should not use ConfigMap in these tests and use instead the Application.
> 
> TBH Application is something very specific to Konflux, and as we want to decouple more KubeSaw from the products ( Konflux , DevSandbox ) those tests might have to change or be moved to the products repos/org. WDYT?

That's true, I agree. As these tests are working for banned user and failing for non banned users, I'd say we can merge them. WDYT?

alexeykazakov:
> TBH Application is something very specific to Konflux, and as we want to decouple more KubeSaw from the products ( Konflux , DevSandbox ) those tests might have to change or be moved to the products repos/org. WDYT?

It doesn't really matter. We use our own "Application" CRD in the tests: https://github.com/codeready-toolchain/toolchain-e2e/tree/master/deploy/crds

We don't need or want to use the real Application CRDs from Konflux. The point is to use something which we have full control over (like our own test CRD) and which is included into the tier (space user roles) templates:
https://github.com/codeready-toolchain/toolchain-e2e/blob/master/test/e2e/parallel/proxy_test.go#L990
The Appstudio tier, which is used in the tests, now comes with the tests (we don't have a built-in appstudio tier in our host operator anymore): https://github.com/codeready-toolchain/toolchain-e2e/blob/master/deploy/nstemplatetiers/appstudio/spacerole_admin.yaml#L17

So we can treat all these appstudio references as internal e2e tests abstractions. 
We could rename it to something more generic but IMO it's not so important. The important part is that we do not use anything from Konflux. We use our own custom test artifacts (appstudio tier, crds, etc) which just happen to have the same names as Konflux stuff. But they are decoupled.

> IIAUIC, these tests checks the workflows for non-banned users, so I don't think it would be a good fit for adding banned user test cases. Am I understanding it correctly?

I'm not sure I follow... 
These tests check the entire proxy flow: https://github.com/codeready-toolchain/toolchain-e2e/blob/master/test/e2e/parallel/proxy_test.go#L158
It tests successful proxy requests and also unauthorized ones. Why checking that banned users do not have access to workspace resources should be separate? 

I'm talking about a similar approach as in https://github.com/codeready-toolchain/toolchain-e2e/pull/1037 (but that was for the ws lister api).

filariow:
> > IIAUIC, these tests checks the workflows for non-banned users, so I don't think it would be a good fit for adding banned user test cases. Am I understanding it correctly?
> 
> I'm not sure I follow... These tests check the entire proxy flow: https://github.com/codeready-toolchain/toolchain-e2e/blob/master/test/e2e/parallel/proxy_test.go#L158 It tests successful proxy requests and also unauthorized ones. Why checking that banned users do not have access to workspace resources should be separate?
> 
> I'm talking about a similar approach as in #1037 (but that was for the ws lister api).

@alexeykazakov thanks for clarifying. Is e107354d0f6e09a4db769e3964427ab1115f0405 going in the direction you're suggesting?

Originally, I put them in a separate file as I didn't identified this suite in `proxy_test.go` and I though this file was already huge. We discussed here https://github.com/codeready-toolchain/toolchain-e2e/pull/1036#discussion_r1722105516 to move them to `proxy_test.go`, and I proposed to leave them in a separate file mainly for this reason.

alexeykazakov:
> Originally, I put them in a separate file as I didn't identified this suite in proxy_test.go and I though this file was already huge. We discussed here https://github.com/codeready-toolchain/toolchain-e2e/pull/1036#discussion_r1722105516 to move them to proxy_test.go, and I proposed to leave them in a separate file mainly for this reason.

Yes, the file is quite long. I was wondering if by moving the test into proxy_test.go you could reuse some code there. If we can't really re-use match then splitting the tests could be a good idea but I would maybe leave it for a separate PR. Just extracting this specific banned-user test case might be confusing. So maybe it's time to refactor this test (move to a separate proxy package and and split them to multiple test files???) but let's keep it is for now, in a single test file. Just for consistency sake.

alexeykazakov (approved):
Thanks!

mfrancisc (approved):
Nice job 👏

mfrancisc:
/retest

infra

PR #1045, opened 2024-09-09
add e2e tests for banneduser

filariow:
Linked to https://github.com/codeready-toolchain/registration-service/pull/455 and https://github.com/codeready-toolchain/host-operator/pull/1079

Signed-off-by: Francesco Ilario <filario@redhat.com>

### DIFF
--- a/test/e2e/parallel/proxy_test.go
+++ b/test/e2e/parallel/proxy_test.go
@@ -1049,7 +1049,7 @@ func TestSpaceLister(t *testing.T) {
 			require.NoError(t, err)
 
 			req.Header.Add("Authorization", fmt.Sprintf("Bearer %s", bannedUser.token))
-			tr := &http.Transport{TLSClientConfig: &tls.Config{InsecureSkipVerify: true}}
+			tr := &http.Transport{TLSClientConfig: &tls.Config{InsecureSkipVerify: true}} // nolint:gosec
 			client := &http.Client{Transport: tr}
 			res, err := client.Do(req)
 			require.NoError(t, err)
@@ -1125,7 +1125,7 @@ func TestSpaceLister(t *testing.T) {
 			require.NoError(t, err)
 
 			req.Header.Add("Authorization", fmt.Sprintf("Bearer %s", lazyBannedUser.token))
-			tr := &http.Transport{TLSClientConfig: &tls.Config{InsecureSkipVerify: true}}
+			tr := &http.Transport{TLSClientConfig: &tls.Config{InsecureSkipVerify: true}} // nolint:gosec
 			client := &http.Client{Transport: tr}
 			res, err := client.Do(req)
 			require.NoError(t, err)
--- a/test/e2e/parallel/proxy_test.go
+++ b/test/e2e/parallel/proxy_test.go
@@ -729,11 +729,6 @@ func TestSpaceLister(t *testing.T) {
 			username:              "banned.user",
 			identityID:            uuid.New(),
 		},
-		"lazybanneduser": {
-			expectedMemberCluster: memberAwait,
-			username:              "lazybanned.user",
-			identityID:            uuid.New(),
-		},
 	}
 	appStudioTierRolesWSOption := commonproxy.WithAvailableRoles([]string{"admin", "contributor", "maintainer", "viewer"})
 
@@ -743,23 +738,11 @@ func TestSpaceLister(t *testing.T) {
 	}
 
 	// share workspaces
-	_ = users["car"].shareSpaceWith(t, awaitilities, users["banneduser"])
-	_ = users["car"].shareSpaceWith(t, awaitilities, users["lazybanneduser"])
 	busSBROnCarSpace := users["car"].shareSpaceWith(t, awaitilities, users["bus"])
 	bicycleSBROnCarSpace := users["car"].shareSpaceWith(t, awaitilities, users["bicycle"])
 	bicycleSBROnBusSpace := users["bus"].shareSpaceWith(t, awaitilities, users["bicycle"])
 	// let's also create a failing SBR so that we can check if it's being added in the bindings list
 	failingSBR := users["bus"].invalidShareSpaceWith(t, awaitilities, users["car"])
-
-	// ban a user,
-	// this will be used in order to make sure this type of user doesn't have access to any of the requests in the workspace lister
-	_ = CreateBannedUser(t, hostAwait, users["banneduser"].signup.Spec.IdentityClaims.Email)
-	// wait until the user is banned
-	_, err := hostAwait.
-		WaitForUserSignup(t, users["banneduser"].signup.Name,
-			wait.UntilUserSignupHasConditions(
-				wait.ConditionSet(wait.Default(), wait.ApprovedByAdmin(), wait.Banned())...))
-	require.NoError(t, err)
 
 	t.Run("car lists workspaces", func(t *testing.T) {
 		// when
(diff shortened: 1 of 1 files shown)


mfrancisc:
You can only pair this with one of regsvc or host-operator, I see it is failing since all 3 PRs now have the same branch name.

mfrancisc:
since it's related to proxy test enhancements I think it should go with the regsvc PR right ?

mfrancisc:
> yes, it should. Would e2e tests run fine even if https://github.com/codeready-toolchain/host-operator/pull/1083 is not merged? I was not able to have them running in my local without this change to host-operator.

Correct! You need to merge first the host-operator PR since the actual registration-service deployment is the one that comes with host-operator as briefly documented here https://github.com/codeready-toolchain/registration-service/blob/master/deploy/README.md. 

So yes feel free to merge the [host-operator PR](https://github.com/codeready-toolchain/host-operator/pull/1083) once you're ready.

test/e2e/parallel/proxy_test.go
```diff
@@ -1033,13 +1041,141 @@ func TestSpaceLister(t *testing.T) {
...
+			client := &http.Client{Transport: tr}
+			res, err := client.Do(req)
+			require.NoError(t, err)
+			defer res.Body.Close()
+
+			require.Equal(t, http.StatusForbidden, res.StatusCode)
+		})
+
+		t.Run("cannot get workspace", func(t *testing.T) {
+			_, err := bannedUser.getWorkspace(t, hostAwait, bannedUser.compliantUsername)
```
alexeykazakov, line 1062: Do we really need to use `getWorkspace()` here? It uses a poll (it's slow because it will keep trying until it's timed out) and hides the actual error. Shouldn't we make sure that everything is ready (spaces, space bindings, etc) and then just use the client directly `proxyCl.Get(...)` and then check that the error is `Forbidden` with the expected error message?

test/e2e/parallel/proxy_test.go
```diff
@@ -1033,13 +1041,141 @@ func TestSpaceLister(t *testing.T) {
...
+							require.Error(t, err)
+							require.True(t, meta.IsNoMatchError(err), "expected NoMatch error, got %v", err)
+						})
+					}
+				})
+			}
+		})
+	})
+
+	t.Run("lazy banned user", func(t *testing.T) {
```
alexeykazakov, line 1105: What is the difference comparing to `banneduser` tests?

filariow: the difference is that `banneduser`'s client does not initialize correctly -successful call to `/apis`, so we expect it to return `NoMatch` errors. In this case the client initialized correctly and we have `Forbidden` errors.

test/e2e/parallel/proxy_test.go
filariow: great point, let me fix that

filariow: done in a533a2b

test/e2e/parallel/proxy_test.go
mfrancisc: should we rename it to something like "user creates client and after is banned" ? But more important does it matter when the user is banned , if before or after the initialization of a kube client ? 

I'm suggesting that probably we don't need this test. WDYT?

mfrancisc (commented):
Nice Job! Thanks a lot for the enhancements to the e2e tests 🙏 

I have only one comment.

test/e2e/parallel/proxy_test.go
filariow: AFAIK, for how Controller-Runtime's Client v0.13.0 is implemented, the Controller-Runtime's client performs a call to the `/apis` endpoint during initialization. If we ban the user before creating the client, this call will fail and subsequent requests will fail with `NoMatch` without really performing the request against the proxy, as the client was not able to initialize the list of resources the cluster is supporting. 

Indeed, the [client.New()](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.13.0/pkg/client#New) function [instantiates a new Dynamic RESTMapper](https://github.com/kubernetes-sigs/controller-runtime/blob/d242fe21e646f034995c4c93e9bba388a0fdaab9/pkg/client/client.go#L108) (if not provided via client's Config).
The [mapper invokes](https://github.com/kubernetes-sigs/controller-runtime/blob/d242fe21e646f034995c4c93e9bba388a0fdaab9/pkg/client/apiutil/dynamicrestmapper.go#L85) the `/apis` endpoint to fetch the list of resources that the cluster is supporting. That's done [invoking the setStaticMapper function here](https://github.com/kubernetes-sigs/controller-runtime/blob/d242fe21e646f034995c4c93e9bba388a0fdaab9/pkg/client/apiutil/dynamicrestmapper.go#L99).
Next usages of the client will lookup the cached list of supported resources to understand if the cluster supports the provided resource and return a `NoMatch` error without really executing the request.

Conversely, is we ban the user lately, the client will effectively perform the call and the proxy will reject it with `Forbidden`.

As the former scenario is mainly testing a Controller-Runtime's Client specific behavior, I'd say we could remove the `banneduser` suite, and rename the `lazy banned user` as `banneduser`. WDYT?

alexeykazakov: Thanks for the detailed explanation.
I don't think we need to test the client (initialization). We want to test our proxy here. So I agree with @mfrancisc and would drop the tests which fails with the client initialization.
Create the client. Then ban the user. Then call the proxy and check the expected 403 error.

filariow: done in 89a664f52fc50c2ac8187c417e84bbb73384bfed

mfrancisc (approved):
Nice Job 👍 

Thanks for the additional changes.

test/e2e/parallel/proxy_test.go
```diff
@@ -1033,13 +1024,86 @@ func TestSpaceLister(t *testing.T) {
...
+		})
+
+		t.Run("cannot list resources", func(t *testing.T) {
+			tt := map[string]client.Client{
+				"home workspace":   proxyClDefault,
+				"shared workspace": proxyClCar,
+			}
+			for k, wc := range tt {
+				t.Run(k, func(t *testing.T) {
+					tt := map[string]client.ObjectList{
```
rsoaresd, line 1092:
```suggestion
                                        resourceLists := map[string]client.ObjectList{
```

rsoaresd (commented):
Great job 🚀

Just a minor suggestion to rename `tt` for better readability to something like:

test/e2e/parallel/proxy_test.go
```diff
@@ -1033,13 +1024,86 @@ func TestSpaceLister(t *testing.T) {
...
+
+		t.Run("cannot list workspaces", func(t *testing.T) {
+			err := proxyClDefault.List(context.TODO(), &toolchainv1alpha1.WorkspaceList{})
+
+			require.Error(t, err)
+			require.True(t, k8serr.IsForbidden(err), "expected Forbidden error, got %v", err)
+		})
+
+		t.Run("cannot list resources", func(t *testing.T) {
+			tt := map[string]client.Client{
```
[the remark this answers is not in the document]
filariow: good idea, let me rename them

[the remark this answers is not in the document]
filariow: done in a8d808fac8142c19e1b53030b5af8d8e8e539d24

alexeykazakov (approved):
Nice! Thanks for addressing the comments.

PR #1046, opened 2024-09-09
Update the tests to not distinguish between member and host

metlos:
Update the tests to not distinguish between member and host because both should now read the connection details from the kubeconfig in TC secret.

Related PRs:
* member-operator: https://github.com/codeready-toolchain/member-operator/pull/595


MatousJobanek (approved):
Cool :+1:

PR #1047, opened 2024-09-17
remove -parallel flag

MatousJobanek:
from the docs https://pkg.go.dev/cmd/go#hdr-Testing_flags
```
...
By default, -parallel is set to the value of GOMAXPROCS.
Setting -parallel to values higher than GOMAXPROCS may cause degraded
performance due to CPU contention, especially when fuzzing.
...
```

### DIFF
--- a/make/test.mk
+++ b/make/test.mk
@@ -37,8 +37,6 @@ ifneq ($(IS_OSD),true)
 LETS_ENCRYPT_PARAM := --lets-encrypt=false
 endif
 
-E2E_PARALLELISM=1
-
 TESTS_RUN_FILTER_REGEXP ?= ""
 
 .PHONY: test-e2e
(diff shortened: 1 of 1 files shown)


mfrancisc (approved):
Good catch!

make/test.mk
```diff
@@ -166,7 +164,7 @@ execute-tests:
 	@echo "Status of ToolchainStatus"
 	-oc get ToolchainStatus -n ${HOST_NS} -o yaml
 	@echo "Starting test $(shell date)"
-	MEMBER_NS=${MEMBER_NS} MEMBER_NS_2=${MEMBER_NS_2} HOST_NS=${HOST_NS} REGISTRATION_SERVICE_NS=${REGISTRATION_SERVICE_NS} go test ${TESTS_TO_EXECUTE} -run ${TESTS_RUN_FILTER_REGEXP} -p 1 -parallel ${E2E_PARALLELISM} -v -timeout=90m -failfast || \
+	MEMBER_NS=${MEMBER_NS} MEMBER_NS_2=${MEMBER_NS_2} HOST_NS=${HOST_NS} REGISTRATION_SERVICE_NS=${REGISTRATION_SERVICE_NS} go test ${TESTS_TO_EXECUTE} -run ${TESTS_RUN_FILTER_REGEXP} -p 1 -v -timeout=90m -failfast || \
```
rajivnathan, line 167: Should we also try adjusting the `-p` value?

```
The 'go test' command may run tests for different packages
    in parallel as well, according to the setting of the -p flag
    (see 'go help build').
```

rajivnathan (approved):
Nice find!

alexeykazakov (approved):

make/test.mk
MatousJobanek: Good question. Yeah, I was thinking about this as well. We test each package separately, so technically we don't need to set the flag at all because the result is the same. 
Previously, we tested multiple packages together (with one command) and we needed to ensure sequential execution, so the flag was needed.
I left the flag there for now is it doesn't change anything and it explicitly says that concurrent package testing is not enabled.

MatousJobanek: TBH, I was wondering if it would make sense to experiment a bit and move the parallel test files into separate packages and see if it improves concurrency and the execution time. 
The code structure would look a bit weird, but if we were able to shorten the execution time by tens of minutes, then it would be worth trying that :-). But I don't expect any huge difference.

PR #1048, opened 2024-09-23
Add a target to get unstuck after a failed e2e test run

metlos:
Add a target to get unstuck after a failed e2e test run with trailing e2e resources with finalizers and no operator running.

This may happen when the migration testsuite fails.


mfrancisc (approved):
Nice 👍

Another option would be to use the `--force` flag but that doesn't always work IIRC.

alexeykazakov (approved):

PR #1053, opened 2024-10-02
Allow proxy to forward requests to namespaces outside of workspace

alexeykazakov:
Paired with https://github.com/codeready-toolchain/registration-service/pull/471

### DIFF
--- a/test/e2e/parallel/proxy_test.go
+++ b/test/e2e/parallel/proxy_test.go
@@ -325,34 +325,27 @@ func TestProxyFlow(t *testing.T) {
 				anotherNamespace := fmt.Sprintf("outside-%s", user.compliantUsername)
 				user.expectedMemberCluster.CreateNamespace(t, anotherNamespace)
 
-				tests := map[string]struct {
-					ProxyClient func() client.Client
-				}{
-					"using explicit user's workspace context": {
-						ProxyClient: func() client.Client {
-							proxyWorkspaceURL := hostAwait.ProxyURLWithWorkspaceContext(user.compliantUsername)
-							proxyCl, err := hostAwait.CreateAPIProxyClient(t, user.token, proxyWorkspaceURL)
-							require.NoError(t, err)
-							return proxyCl
-						},
-					},
-					"not using any workspace context": {
-						ProxyClient: func() client.Client {
-							return user.createProxyClient(t, hostAwait)
-						},
-					},
-				}
+				proxyWorkspaceURL := hostAwait.ProxyURLWithWorkspaceContext(user.compliantUsername)
+				explicitWorkspaceCtxClient, err := hostAwait.CreateAPIProxyClient(t, user.token, proxyWorkspaceURL) // a client with a workspace context
+				require.NoError(t, err)
+				withoutWorkspaceCtxClient := user.createProxyClient(t, hostAwait) // a client with no workspace context
 
 				t.Run("fail when accessing unshared namespace", func(t *testing.T) {
-					for k, tc := range tests {
-						t.Run(k, func(t *testing.T) {
-							// then
-							cm := corev1.ConfigMap{}
-							// every namespace has a CM "kube-root-ca.crt" but we can't get it because the user doesn't have access to the namespace
-							err := tc.ProxyClient().Get(context.TODO(), types.NamespacedName{Namespace: anotherNamespace, Name: "kube-root-ca.crt"}, &cm)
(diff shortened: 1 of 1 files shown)


mfrancisc (approved):

test/e2e/parallel/proxy_test.go
```diff
@@ -317,25 +318,83 @@ func TestProxyFlow(t *testing.T) {
...
+						},
 					},
-					Spec: appstudiov1.ApplicationSpec{
-						DisplayName: "Should be forbidden",
+					"not using any workspace context": {
+						ProxyClient: func() client.Client {
+							return user.createProxyClient(t, hostAwait)
+						},
 					},
 				}
```
MatousJobanek, line 344: You don't need to instantiate the client for every iteration separately, you can do it once and then reuse it, right?
```go
explicitWorkspaceCtxClient, err := hostAwait.CreateAPIProxyClient(t, user.token, proxyWorkspaceURL)
require.NoError(t, err)
withoutWorkspaceCtxClient := user.createProxyClient(t, hostAwait)
```

so you can simplify the code by dropping the iterations (as I mentioned in the other PR, they are sometimes hard to follow and investigate the failures in the complicated tests) and replacing the test with:
```go
t.Run("fail when accessing unshared namespace", func(t *testing.T) {
	cm := corev1.ConfigMap{}
	t.Run("using explicit user's workspace context", func(t *testing.T) {
		// when
		// every namespace has a CM "kube-root-ca.crt" but we can't get it because the user doesn't have access to the namespace
		err = explicitWorkspaceCtxClient.Get(context.TODO(), types.NamespacedName{Namespace: anotherNamespace, Name: "kube-root-ca.crt"}, &cm)
		// then
		require.EqualError(t, err, fmt.Sprintf(`configmaps "kube-root-ca.crt" is forbidden: User "%s" cannot get resource "configmaps" in API group "" in the namespace "%s"`, user.compliantUsername, anotherNamespace))
	})
	t.Run("not using any workspace context", func(t *testing.T) {
		// when
		// every namespace has a CM "kube-root-ca.crt" but we can't get it because the user doesn't have access to the namespace
		err := withoutWorkspaceCtxClient.Get(context.TODO(), types.NamespacedName{Namespace: anotherNamespace, Name: "kube-root-ca.crt"}, &cm)
		// then
		require.EqualError(t, err, fmt.Sprintf(`configmaps "kube-root-ca.crt" is forbidden: User "%s" cannot get resource "configmaps" in API group "" in the namespace "%s"`, user.compliantUsername, anotherNamespace))
	})
})
```
and the next one with:
```go 
t.Run("using explicit user's workspace context", func(t *testing.T) {
	// when
	cm := corev1.ConfigMap{}
	// every namespace has a CM "kube-root-ca.crt" but we can't get it because the user doesn't have access to the namespace
	err = explicitWorkspaceCtxClient.Get(context.TODO(), types.NamespacedName{Namespace: anotherNamespace, Name: "kube-root-ca.crt"}, &cm)
	// then
	require.NoError(t, err)
	assert.Equal(t, anotherNamespace, cm.Namespace)
})
t.Run("not using any workspace context", func(t *testing.T) {
	// when
	cm := corev1.ConfigMap{}
	// every namespace has a CM "kube-root-ca.crt" but we can't get it because the user doesn't have access to the namespace
	err := withoutWorkspaceCtxClient.Get(context.TODO(), types.NamespacedName{Namespace: anotherNamespace, Name: "kube-root-ca.crt"}, &cm)
	// then
	require.NoError(t, err)
	assert.Equal(t, anotherNamespace, cm.Namespace)
})
```
it's only one line longer, but it's easier to read.

alexeykazakov: > You don't need to instantiate the client for every iteration separately, you can do it once and then reuse it, right?

Yes. This is what I was thinking about too... Would safe two additional client initializations.

> so you can simplify the code by dropping the iterations (as I mentioned in the other PR, they are sometimes hard to follow and investigate the failures in the complicated tests) and replacing the test with:

This is what I originally did but then decided to make the code shorter ;) But it's a good point regarding investigation and debugging. It's a bit harder when it's in a loop. Let me re-organize it.

alexeykazakov: I reorganized the tests here: https://github.com/codeready-toolchain/toolchain-e2e/pull/1053/commits/dd57b32e7564a78b30d489637a4295e7f7b23e3a

MatousJobanek (approved):
Thanks for the additional changes

rajivnathan (approved):

PR #1054, opened 2024-10-02
Don't use the ToolchainCluster.Spec fields

metlos:
Instead, use their replacements in the ToolchainCluster status. This will enable us to remove the fields from the spec and only rely on the kubeconfig (and the status fields that project the values out of the kubeconfig).

### DIFF
--- a/testsupport/wait/awaitility.go
+++ b/testsupport/wait/awaitility.go
@@ -172,9 +172,7 @@ func (a *Awaitility) WaitForToolchainClusterWithCondition(t *testing.T, namespac
 // and running in the given expected namespace. It also checks if the CR has the ClusterConditionType
 func (a *Awaitility) GetToolchainCluster(t *testing.T, namespace string, cdtype toolchainv1alpha1.ConditionType) (toolchainv1alpha1.ToolchainCluster, bool, error) {
 	clusters := &toolchainv1alpha1.ToolchainClusterList{}
-	if err := a.Client.List(context.TODO(), clusters, client.InNamespace(a.Namespace), client.MatchingLabels{
-		"namespace": namespace,
-	}); err != nil {
+	if err := a.Client.List(context.TODO(), clusters, client.InNamespace(a.Namespace)); err != nil {
 		return toolchainv1alpha1.ToolchainCluster{}, false, err
 	}
 	if len(clusters.Items) == 0 {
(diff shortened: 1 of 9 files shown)


MatousJobanek (approved):
:+1:

MatousJobanek:
> /test publish-operators-for-e2e-tests

this is a GH action workflow. If you want to restart it, then go to the GH Actions UI and click on "re-run".

However, the workflow is configured to automatically run also when you leave any of the following comments: `/retest`, `/test all`, `/test e2e`, or `/test` https://github.com/codeready-toolchain/toolchain-e2e/blob/315bbf10a2c64090096f6b110beb609866746268/.github/workflows/publish-operators-for-e2e-tests.yml#L16

ranakan19 (approved):

mfrancisc (approved):
👍 

There is probably one small leftover: https://github.com/metlos/toolchain-e2e/blob/374daa079348432f62fd4da7a50e09946b46ad43/testsupport/wait/host.go#L1419

testsupport/wait/awaitility.go
```diff
@@ -172,17 +172,15 @@ func (a *Awaitility) WaitForToolchainClusterWithCondition(t *testing.T, namespac
...
 	if len(clusters.Items) == 0 {
 		t.Logf("no toolchaincluster resource with expected labels: namespace='%s'", namespace)
 	}
 	// assume there is zero or 1 match only
 	for _, cl := range clusters.Items {
-		if cd.IsTrue(cl.Status.Conditions, cdtype) {
+		if cl.Status.OperatorNamespace == namespace && cd.IsTrue(cl.Status.Conditions, cdtype) {
 			return cl, true, nil
 		}
 	}
```
MatousJobanek, line 186: btw, the message in the middle is not up-to-date anymore. Could be modified to:
```
func (a *Awaitility) GetToolchainCluster(t *testing.T, namespace string, cdtype toolchainv1alpha1.ConditionType) (toolchainv1alpha1.ToolchainCluster, bool, error) {
	clusters := &toolchainv1alpha1.ToolchainClusterList{}
	if err := a.Client.List(context.TODO(), clusters, client.InNamespace(a.Namespace)); err != nil {
		return toolchainv1alpha1.ToolchainCluster{}, false, err
	}
	// assume there is zero or 1 match only
	for _, tCluster := range clusters.Items {
		if tCluster.Status.OperatorNamespace == namespace && cd.IsTrue(tCluster.Status.Conditions, cdtype) {
			return tCluster, true, nil
		}
	}
	t.Logf("no ToolchainCluster resource that would be ready and would contain Status.OperatorNamespace: '%s'", namespace)
	return toolchainv1alpha1.ToolchainCluster{}, false, nil
```

PR #1058, opened 2024-11-05
Tests for intel-x-large tier

alexeykazakov:
Paired with https://github.com/codeready-toolchain/host-operator/pull/1102


ranakan19 (approved):
Looks Good,
minor but I'm unsure if using "Large" in the function name is ideal when reusing the same function for both large and xlarge.

xcoulon (approved):

PR #1059, opened 2024-11-06
test: SANDBOX-814 align replica count in registration-service

rsoaresd:
# Description
Currently, running `make e2e-tests` deploys registration-service with 2 replicas, while `make dev-deploy-e2e` deploys it with 3 replicas. So, this PR aligns replica count in registration-service within the e2e and dev environment and adds a test to ensure that we can update the replica count trough ToolchainConfig.

[SANDBOX-814](https://issues.redhat.com/browse/SANDBOX-814)

### DIFF
--- a/test/e2e/toolchain_config_test.go
+++ b/test/e2e/toolchain_config_test.go
@@ -1,0 +1,31 @@
+package e2e
+
+import (
+	"context"
+	"testing"
+
+	testconfig "github.com/codeready-toolchain/toolchain-common/pkg/test/config"
+	"github.com/codeready-toolchain/toolchain-e2e/testsupport"
+	"github.com/stretchr/testify/require"
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func TestToolchainConfig(t *testing.T) {
+	awaitilities := testsupport.WaitForDeployments(t)
+	hostAwait := awaitilities.Host()
+
+	t.Run("update replica count for registration-service", func(t *testing.T) {
+		// get current replica count in registration-service
+		registrationServiceName := "registration-service"
+		rs := &appsv1.Deployment{}
+		err := hostAwait.Client.Get(context.TODO(), types.NamespacedName{Namespace: hostAwait.Namespace, Name: registrationServiceName}, rs)
+		require.NoError(t, err)
+		currentNrReplicas := *rs.Spec.Replicas
+
+		// update replica count in registration-service
+		newNrReplicas := currentNrReplicas + 1
+		hostAwait.UpdateToolchainConfig(t, testconfig.RegistrationService().Replicas(newNrReplicas))
+		hostAwait.WaitForDeploymentToGetReady(t, registrationServiceName, int(newNrReplicas))
+	})
+}
--- a/deploy/host-operator/e2e-tests/toolchainconfig.yaml
+++ b/deploy/host-operator/e2e-tests/toolchainconfig.yaml
@@ -15,7 +15,8 @@ spec:
           segmentWriteKey: 'test devspaces segment write key'
         segmentWriteKey: 'test sandbox segment write key'
       environment: 'e2e-tests'
-      replicas: 2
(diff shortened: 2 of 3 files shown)


test/e2e/toolchain_config_test.go
```diff
@@ -0,0 +1,31 @@
...
+		// get current replica count in registration-service
+		registrationServiceName := "registration-service"
+		rs := &appsv1.Deployment{}
+		err := hostAwait.Client.Get(context.TODO(), types.NamespacedName{Namespace: hostAwait.Namespace, Name: registrationServiceName}, rs)
+		require.NoError(t, err)
+		currentNrReplicas := *rs.Spec.Replicas
+
+		// update replica count in registration-service
+		newNrReplicas := currentNrReplicas + 1
+		hostAwait.UpdateToolchainConfig(t, testconfig.RegistrationService().Replicas(newNrReplicas))
```
alexeykazakov, line 28: Hm... wouldn't it affect other tests? This test will increase the replicas to +1 but we still expect 3 replicas in other tests, rigth?

mfrancisc: I think for this to not happen, we need to reset the value the original one once the test ends. 

Since the test is not a parallel one it shouldn't affect other ongoing executions, unless I've missed something.

rsoaresd: @alexeykazakov @mfrancisc 

This test will not affect other tests because it does not run in parallel and`UpdateToolchainConfig` [already reverts the ToolchainConfig to the original one](https://github.com/codeready-toolchain/toolchain-e2e/blob/master/testsupport/wait/host.go#L1732-L1747) when the test finishes. 

But I had the same thought when I started to write the test. I actually added the revert but then I deleted it since it is unnecessary. Sorry for not mentioning in the PR description. I will add :)

rsoaresd: I will also put a comment on the test

mfrancisc (approved):
Nice Job 👍 

Thanks for adding the new test and for the additional comments.

test/e2e/toolchain_config_test.go
mfrancisc: > UpdateToolchainConfig [already reverts the ToolchainConfig to the original one](https://github.com/codeready-toolchain/toolchain-e2e/blob/master/testsupport/wait/host.go#L1732-L1747) when the test finishes.

Good point, sorry I've forgot about that 🤦‍♂️

MatousJobanek (approved):
Thanks for aligning that :rocket: 
As I mentioned in the comment, it should be completely OK to run it as part of the parallel test suite to minimize the impact on the test execution time.

alexeykazakov (approved):

test/e2e/toolchain_config_test.go
rsoaresd: @MatousJobanek All the parallel tests call `WaitForDeployments` which calls `waitForOpetators` that checks the number of replicas of the registration service.

If we run the test in parallel, we will get errors like this:

```
    user_assertions.go:267: 
        	Error Trace:	/Users/rsoaresd/toolchain-e2e/testsupport/user_assertions.go:267
        	            				/Users/rsoaresd/toolchain-e2e/testsupport/user_assertions.go:36
        	            				/Users/rsoaresd/toolchain-e2e/test/e2e/parallel/e2e_test.go:185
        	            	Get ".../apis/toolchain.dev.openshift.com/v1alpha1/namespaces/toolchain-member-11123235/nstemplatesets/johnsmith": local error: tls: bad record MAC
        	Test:       	TestE2EFlow/try_to_break_UserAccount/delete_identity_mapping_and_wait_until_recreated
```

MatousJobanek: the function should be called only once for the whole parallel suite execution
https://github.com/codeready-toolchain/toolchain-e2e/blob/0c385c92fd63315225dde4092f46be0479ef296d/testsupport/init.go#L50
but let me know if I'm missing some small detail

PR #1062, opened 2024-11-12
test: SANDBOX-816 refactor WaitForDeployments

rsoaresd:
# Description
To reuse the function `WaitForDeployments` in WA E2E tests, we need to refactor `WaitForDeployments` to take into consideration when SECOND_MEMBER_MODE is set to false.

[SANDBOX-816](https://issues.redhat.com/browse/SANDBOX-816)

### DIFF
--- a/make/test.mk
+++ b/make/test.mk
@@ -164,7 +164,7 @@ execute-tests:
 	@echo "Status of ToolchainStatus"
 	-oc get ToolchainStatus -n ${HOST_NS} -o yaml
 	@echo "Starting test $(shell date)"
-	MEMBER_NS=${MEMBER_NS} MEMBER_NS_2=${MEMBER_NS_2} HOST_NS=${HOST_NS} REGISTRATION_SERVICE_NS=${REGISTRATION_SERVICE_NS} go test ${TESTS_TO_EXECUTE} -run ${TESTS_RUN_FILTER_REGEXP} -p 1 -v -timeout=90m -failfast || \
+	MEMBER_NS=${MEMBER_NS} MEMBER_NS_2=${MEMBER_NS_2} HOST_NS=${HOST_NS} REGISTRATION_SERVICE_NS=${REGISTRATION_SERVICE_NS} SECOND_MEMBER_MODE=true go test ${TESTS_TO_EXECUTE} -run ${TESTS_RUN_FILTER_REGEXP} -p 1 -v -timeout=90m -failfast || \
 	($(MAKE) print-logs HOST_NS=${HOST_NS} MEMBER_NS=${MEMBER_NS} MEMBER_NS_2=${MEMBER_NS_2} REGISTRATION_SERVICE_NS=${REGISTRATION_SERVICE_NS} && exit 1)
 
 .PHONY: print-logs
--- a/make/test.mk
+++ b/make/test.mk
@@ -164,7 +164,7 @@ execute-tests:
 	@echo "Status of ToolchainStatus"
 	-oc get ToolchainStatus -n ${HOST_NS} -o yaml
 	@echo "Starting test $(shell date)"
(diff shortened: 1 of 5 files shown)


testsupport/init.go
```diff
@@ -263,3 +276,10 @@ func schemeWithAllAPIs(t *testing.T) *runtime.Scheme {
 	require.NoError(t, builder.AddToScheme(s))
 	return s
 }
+
+func IsSecondMemberMode(t *testing.T) bool {
+	secondMemberMode := os.Getenv(wait.SecondMemberModeVar)
```
MatousJobanek, line 281: Are you sure that this variable is available at the time of the test execution? It is not explicitly set in the makefile https://github.com/codeready-toolchain/toolchain-e2e/blob/0c385c92fd63315225dde4092f46be0479ef296d/make/test.mk#L167
What is set is the `MEMBER_NS_2` that is already retrieved in the test suite, so you could technically rely on that one - if it's empty then it's running in a "single member mode"

rsoaresd: I forgot to update there! Sorry, thanks!

I thought about using `MEMBER_NS_2`, but then I saw that we have `SECOND_MEMBER_MODE`, and I decided to use that one.  Essentially because I think it would be more intuitive if we point `SECOND_MEMBER_MODE=false`, when using the make targets like `make prepare-e2e deploy-e2e SECOND_MEMBER_MODE=false (...)` in WA

testsupport/init.go
```diff
@@ -209,13 +214,17 @@ func WaitForDeployments(t *testing.T) wait.Awaitilities {
 		// Also verify the autoscaling buffer in both members
 		webhookImage := initMemberAwait.GetContainerEnv(t, "MEMBER_OPERATOR_WEBHOOK_IMAGE")
 		require.NotEmpty(t, webhookImage, "The value of the env var MEMBER_OPERATOR_WEBHOOK_IMAGE wasn't found in the deployment of the member operator.")
-		err = initMember2Await.WaitUntilWebhookDeleted(t) // webhook on member2 should be deleted
-		require.NoError(t, err)
+		if isSecondMemberMode {
```
MatousJobanek, line 217: btw, you could do 
```suggestion
		if IsSecondMemberMode(t) {
```
without storing it in a variable and passing it through the functions. I know that it would call the `os.Getenv` a few times, but it's fine and it would be easier to read, WDYT?

testsupport/init.go
MatousJobanek: either option is fine :-)

testsupport/init.go
rsoaresd: I agree! Tysm!

mfrancisc (approved):
Looks good 👍

make/test.mk
```diff
@@ -164,7 +164,7 @@ execute-tests:
 	@echo "Status of ToolchainStatus"
 	-oc get ToolchainStatus -n ${HOST_NS} -o yaml
 	@echo "Starting test $(shell date)"
-	MEMBER_NS=${MEMBER_NS} MEMBER_NS_2=${MEMBER_NS_2} HOST_NS=${HOST_NS} REGISTRATION_SERVICE_NS=${REGISTRATION_SERVICE_NS} go test ${TESTS_TO_EXECUTE} -run ${TESTS_RUN_FILTER_REGEXP} -p 1 -v -timeout=90m -failfast || \
+	MEMBER_NS=${MEMBER_NS} MEMBER_NS_2=${MEMBER_NS_2} HOST_NS=${HOST_NS} REGISTRATION_SERVICE_NS=${REGISTRATION_SERVICE_NS} SECOND_MEMBER_MODE=true go test ${TESTS_TO_EXECUTE} -run ${TESTS_RUN_FILTER_REGEXP} -p 1 -v -timeout=90m -failfast || \
```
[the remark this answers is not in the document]
rsoaresd: Sorry! Thanks!

alexeykazakov (approved):
Looks good but shouldn't we update our README to mention this new `SECOND_MEMBER_MODE` env var option?

rsoaresd:
> Looks good but shouldn't we update our README to mention this new `SECOND_MEMBER_MODE` env var option?

Yes, pushed! Thanks

alexeykazakov (approved):

PR #1068, opened 2024-11-21
add deploy-toolchain-resources-e2e

rsoaresd:
# Description
- add `deploy-toolchain-resources-e2e` to deploy the same resources as `make test-e2e` but doesn't run tests. It will be useful for wa

[SANDBOX-811](https://issues.redhat.com/browse/SANDBOX-811)

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -149,9 +149,11 @@ If you are still confused by the different e2e/operator location, execution and 
 
 All e2e resources (host operator, member operator, registration-service, CRDs, etc) can be deployed without running tests:
 
-* `make dev-deploy-e2e-local` - deploys the same resources as `make test-e2e-local` but doesn't run tests.
+* `make dev-deploy-e2e-local` - deploys the same resources as `make test-e2e-local` in dev environment but doesn't run tests.
 
-* `make dev-deploy-e2e` - deploys the same resources as `make test-e2e` but doesn't run tests.
+* `make dev-deploy-e2e` - deploys the same resources as `make test-e2e` in dev environment but doesn't run tests.
+
+* `make deploy-toolchain-resources-e2e` - deploys the same resources as `make test-e2e` but doesn't run tests.
 
 NOTE: By default these targets deploy resources to `toolchain-host-operator` and `toolchain-member-operator` namespaces.
 
--- a/make/dev.mk
+++ b/make/dev.mk
@@ -1,8 +1,4 @@
-DEV_MEMBER_NS := toolchain-member-operator
-DEV_MEMBER_NS_2 := toolchain-member2-operator
-DEV_HOST_NS := toolchain-host-operator
 DEV_SSO_NS := toolchain-dev-sso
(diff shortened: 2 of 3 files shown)


rajivnathan (approved):
Nice 👍 Just a couple very minor suggestions

mfrancisc (commented):
Looks good 👍 

Alternatively you could have something like this:

```
.PHONY: deploy-single-member-e2e
## Deploy operators in e2e mode with single member without running tests
deploy-single-member-e2e: SECOND_MEMBER_MODE=false
deploy-single-member-e2e: HOST_NS=toolchain-host-operator
deploy-single-member-e2e: MEMBER_NS=toolchain-member-operator
deploy-single-member-e2e: REGISTRATION_SERVICE_NS=toolchain-host-operator
deploy-single-member-e2e: prepare-e2e deploy-e2e
        @echo "The e2e deployment successfully finished"
```

rsoaresd:
> Looks good 👍
> 
> Alternatively you could have something like this:
> 
> ```
> .PHONY: deploy-single-member-e2e
> ## Deploy operators in e2e mode with single member without running tests
> deploy-single-member-e2e: SECOND_MEMBER_MODE=false
> deploy-single-member-e2e: HOST_NS=toolchain-host-operator
> deploy-single-member-e2e: MEMBER_NS=toolchain-member-operator
> deploy-single-member-e2e: REGISTRATION_SERVICE_NS=toolchain-host-operator
> deploy-single-member-e2e: prepare-e2e deploy-e2e
>         @echo "The e2e deployment successfully finished"
> 	@echo "To clean the cluster run 'make clean-e2e-resources'"
> ```

Thanks, @mfrancisc! Addressed it! I did not include the echo commands because `deploy-e2e` already [prints](https://github.com/codeready-toolchain/toolchain-e2e/blob/master/make/test.mk#L94) that the deployment finished successfully. Regarding the clean-up, I changed to use `prepare-and-deploy-e2e` which [prints](https://github.com/codeready-toolchain/toolchain-e2e/blob/master/make/test.mk#L61) it.

If we set the vars in the suggested way, it will fail because [ifeq](https://github.com/codeready-toolchain/toolchain-e2e/blob/master/make/test.mk#L373) is not dynamic and does not re-evaluate variables updated during execution, so even if we changed `SECOND_MEMBER_MODE` to false, it was being evaluated as true and it was deploying the second member.

mfrancisc:
> If we set the vars in the suggested way, it will fail because [ifeq](https://github.com/codeready-toolchain/toolchain-e2e/blob/master/make/test.mk#L373) is not dynamic and does not re-evaluate variables updated during execution, so even if we changed SECOND_MEMBER_MODE to false, it was being evaluated as true and it was deploying the second member.

Thanks for looking into it @rsoaresd , you're right about the `ifeq` , I would need to change that to bash if syntax. But that was just a suggestion, your approach works fine 👍

mfrancisc (approved):
Great job 👍 

Thanks for answering my comments/questions.

PR #1070, opened 2024-11-22
Update webhook test for VM webhook update

rajivnathan:
Paired with https://github.com/codeready-toolchain/member-operator/pull/607


alexeykazakov (approved):

mfrancisc (commented):
Should we add test cases for creating and updating vms with `runStrategy` set and expect that the webhook rejects that ? 

Basically updating this test https://github.com/codeready-toolchain/toolchain-e2e/blob/9f9588d65506587fa42d2761b98f97bb9cad8d6d/test/e2e/parallel/vm_webhook_test.go#L24 .

rajivnathan:
> Should we add test cases for creating and updating vms with `runStrategy` set and expect that the webhook rejects that ?
> 
> Basically updating this test
> 
> https://github.com/codeready-toolchain/toolchain-e2e/blob/9f9588d65506587fa42d2761b98f97bb9cad8d6d/test/e2e/parallel/vm_webhook_test.go#L24
> 
> .

We mainly rely on unit testing for all the validating webhooks if I'm not mistaken. But if we think it's needed, I can give it a try.

PR #1071, opened 2024-11-25
fix: SANDBOX-838 check that SocialEvent is in a ready state when WaitForSocialEvent after creation

rsoaresd:
# Description
"TestActivationCodeVerification" failed twice, for now, with `Operation cannot be fulfilled on socialevents.toolchain.dev.openshift.com`. I guess this is happening because when we call`WaitForSocialEvent`, we are not checking that the resource was in a ready state. So, when we update the resource right after, the resource is not ready in some edge cases and we do not have the latest version of the resource, causing the `Operation cannot be fulfilled on socialevents.toolchain.dev.openshift.com`.

[SANDBOX-838](https://issues.redhat.com/browse/SANDBOX-838)

### DIFF
--- a/test/e2e/parallel/registration_service_test.go
+++ b/test/e2e/parallel/registration_service_test.go
@@ -26,6 +26,7 @@ import (
 	routev1 "github.com/openshift/api/route/v1"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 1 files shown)


alexeykazakov (approved):

mfrancisc (approved):
Great catch 👍

PR #1072, opened 2024-11-26
Reinstate webhook tests

rajivnathan:
It's the follow up to https://github.com/codeready-toolchain/toolchain-e2e/pull/1070 to re-enable the webhook tests

### DIFF
--- a/test/migration/setup/setup_migration_test.go
+++ b/test/migration/setup/setup_migration_test.go
@@ -9,10 +9,6 @@ import (
 
 func TestSetupMigration(t *testing.T) {
 	// given
-	// set env var to skip the mutating webhook check on migration setup temporarily since the old deployment
-	// will deploy the webhooks with the old configuration but the tests will be expecting the new configuration
-	// This should be removed after PR https://github.com/codeready-toolchain/toolchain-e2e/pull/809 is merged
-	t.Setenv("skip-webhook-checks-on-setup", "true")
 	awaitilities := WaitForOperators(t)
 
 	runner := migration.SetupMigrationRunner{
--- a/testsupport/wait/member.go
+++ b/testsupport/wait/member.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"os"
 	"reflect"
 	"sort"
 	"strings"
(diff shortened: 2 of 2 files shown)


testsupport/wait/member.go
```diff
@@ -2296,11 +2295,6 @@ func (a *MemberAwaitility) verifySecret(t *testing.T) []byte {
 }
 
 func (a *MemberAwaitility) verifyMutatingWebhookConfig(t *testing.T, ca []byte) {
-	if val := os.Getenv("skip-webhook-checks-on-setup"); val == "true" {
```
alexeykazakov, line 2299: I guess we can keep these checks for the future use and just remove the `t.Setenv("skip-webhook-checks-on-setup", "true")` above?

alexeykazakov (approved):

testsupport/wait/member.go
mfrancisc: +1

test/migration/setup/setup_migration_test.go
```diff
@@ -9,10 +9,6 @@ import (
 
 func TestSetupMigration(t *testing.T) {
 	// given
-	// set env var to skip the mutating webhook check on migration setup temporarily since the old deployment
-	// will deploy the webhooks with the old configuration but the tests will be expecting the new configuration
-	// This should be removed after PR https://github.com/codeready-toolchain/toolchain-e2e/pull/809 is merged
-	t.Setenv("skip-webhook-checks-on-setup", "true")
```
mfrancisc, line 15: in addition to Alexey's comment maybe we could keep this commented so that we can just uncomment when needed. WDYT? Of course adjusting a bit the description above.

rajivnathan: The webhook test was failing so I thought it was due to the migration thing. I forgot that we made the change to the wait logic. I guess there was something actually wrong with the webhook tests then.

MatousJobanek: Yeah, I checked the failure in the other PR and it wasn't related to setup migration test suite.
Anyway, let's drop this and keep an eye on any flakiness/failures related to migration.

MatousJobanek (approved):

PR #1074, opened 2024-11-28
SANDBOX-838: loop update calls

rsoaresd:
# Description
#1071 addressed a potential fix, but to ensure we do not face it in the future, we should do the update call in a loop. 

Since other update calls are not being done in a loop, this PR addressed it as well. 

[SANDBOX-838](https://issues.redhat.com/browse/SANDBOX-838)

### DIFF
--- a/test/e2e/parallel/proxy_test.go
+++ b/test/e2e/parallel/proxy_test.go
@@ -233,14 +233,14 @@ func TestProxyFlow(t *testing.T) {
 				}
 
 				t.Run("use proxy to update a HAS Application CR in the user appstudio namespace via proxy API", func(t *testing.T) {
-					// Get application name
+					// Update application
 					applicationName := user.getApplicationName(0)
+					// Get application
+					proxyApp := user.getApplication(t, proxyCl, applicationName)
 					// Update DisplayName
 					changedDisplayName := fmt.Sprintf("Proxy test for user %s - updated application", tenantNsName(user.compliantUsername))
-					// Update application
-					proxyApp, err := awaitilities.Host().UpdateApplication(t, proxyCl, applicationName, tenantNsName(user.compliantUsername), func(app *appstudiov1.Application) {
-						app.Spec.DisplayName = changedDisplayName
-					})
+					proxyApp.Spec.DisplayName = changedDisplayName
+					err := proxyCl.Update(context.TODO(), proxyApp)
 					require.NoError(t, err)
 
 					// Find application and check, if it is updated
@@ -320,6 +320,7 @@ func TestProxyFlow(t *testing.T) {
 
 			t.Run("get resource from namespace outside of user's workspace", func(t *testing.T) {
 				// given
+
 				// create a namespace which does not belong to the user's workspace
 				anotherNamespace := fmt.Sprintf("outside-%s", user.compliantUsername)
 				user.expectedMemberCluster.CreateNamespace(t, anotherNamespace)
--- a/test/e2e/parallel/proxy_test.go
+++ b/test/e2e/parallel/proxy_test.go
@@ -8,7 +8,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	rbacv1 "k8s.io/api/rbac/v1"
 	"net/http"
 	"net/url"
 	"os"
@@ -17,6 +16,8 @@ import (
 	"sync"
 	"testing"
 	"time"
+
+	rbacv1 "k8s.io/api/rbac/v1"
 
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
 	commonproxy "github.com/codeready-toolchain/toolchain-common/pkg/proxy"
@@ -232,14 +233,14 @@ func TestProxyFlow(t *testing.T) {
 				}
 
 				t.Run("use proxy to update a HAS Application CR in the user appstudio namespace via proxy API", func(t *testing.T) {
-					// Update application
+					// Get application name
 					applicationName := user.getApplicationName(0)
-					// Get application
-					proxyApp := user.getApplication(t, proxyCl, applicationName)
 					// Update DisplayName
 					changedDisplayName := fmt.Sprintf("Proxy test for user %s - updated application", tenantNsName(user.compliantUsername))
-					proxyApp.Spec.DisplayName = changedDisplayName
-					err := proxyCl.Update(context.TODO(), proxyApp)
+					// Update application
+					proxyApp, err := awaitilities.Host().UpdateApplication(t, proxyCl, applicationName, tenantNsName(user.compliantUsername), func(app *appstudiov1.Application) {
+						app.Spec.DisplayName = changedDisplayName
+					})
 					require.NoError(t, err)
 
 					// Find application and check, if it is updated
@@ -319,7 +320,6 @@ func TestProxyFlow(t *testing.T) {
 
 			t.Run("get resource from namespace outside of user's workspace", func(t *testing.T) {
 				// given
-
 				// create a namespace which does not belong to the user's workspace
 				anotherNamespace := fmt.Sprintf("outside-%s", user.compliantUsername)
 				user.expectedMemberCluster.CreateNamespace(t, anotherNamespace)
--- a/test/e2e/parallel/socialevent_test.go
+++ b/test/e2e/parallel/socialevent_test.go
@@ -1,7 +1,6 @@
 package parallel
 
 import (
-	"context"
 	"testing"
 	"time"
 
@@ -79,7 +78,10 @@ func TestCreateSocialEvent(t *testing.T) {
 			event.Spec.UserTier = "deactivate30"
 
 			// when
-			err := hostAwait.Client.Update(context.TODO(), event)
+			event, err = hostAwait.UpdateSocialEvent(t, false, event.Name,
+				func(ev *toolchainv1alpha1.SocialEvent) {
+					ev.Spec.UserTier = "deactivate30"
+				})
 
 			// then
 			require.NoError(t, err)
@@ -115,7 +117,10 @@ func TestCreateSocialEvent(t *testing.T) {
 			event.Spec.SpaceTier = "base"
 
 			// when
-			err := hostAwait.Client.Update(context.TODO(), event)
+			event, err = hostAwait.UpdateSocialEvent(t, false, event.Name,
+				func(ev *toolchainv1alpha1.SocialEvent) {
+					ev.Spec.SpaceTier = "base"
+				})
 
 			// then
 			require.NoError(t, err)
--- a/test/e2e/parallel/socialevent_test.go
+++ b/test/e2e/parallel/socialevent_test.go
@@ -74,9 +74,6 @@ func TestCreateSocialEvent(t *testing.T) {
 		require.NoError(t, err)
 
 		t.Run("update with valid tier name", func(t *testing.T) {
-			// given
-			event.Spec.UserTier = "deactivate30"
-
 			// when
 			event, err = hostAwait.UpdateSocialEvent(t, false, event.Name,
 				func(ev *toolchainv1alpha1.SocialEvent) {
@@ -113,9 +110,6 @@ func TestCreateSocialEvent(t *testing.T) {
 		require.NoError(t, err)
 
 		t.Run("update with valid tier name", func(t *testing.T) {
-			// given
-			event.Spec.SpaceTier = "base"
-
 			// when
 			event, err = hostAwait.UpdateSocialEvent(t, false, event.Name,
 				func(ev *toolchainv1alpha1.SocialEvent) {
--- a/test/e2e/parallel/socialevent_test.go
+++ b/test/e2e/parallel/socialevent_test.go
@@ -75,10 +75,11 @@ func TestCreateSocialEvent(t *testing.T) {
 
 		t.Run("update with valid tier name", func(t *testing.T) {
 			// when
-			event, err = hostAwait.UpdateSocialEvent(t, false, event.Name,
-				func(ev *toolchainv1alpha1.SocialEvent) {
-					ev.Spec.UserTier = "deactivate30"
-				})
+			event, err = For(t, hostAwait.Awaitility, &toolchainv1alpha1.SocialEvent{}).
+				Update(event.Name,
+					func(ev *toolchainv1alpha1.SocialEvent) {
+						ev.Spec.UserTier = "deactivate30"
+					})
 
 			// then
 			require.NoError(t, err)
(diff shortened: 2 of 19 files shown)


rsoaresd:
/retest

hitting known flaky test [SANDBOX-837](https://issues.redhat.com/browse/SANDBOX-837)

rsoaresd:
/retest

hitting known flaky test [SANDBOX-836](https://issues.redhat.com/browse/SANDBOX-836)

test/e2e/parallel/proxy_test.go
```diff
@@ -232,14 +233,14 @@ func TestProxyFlow(t *testing.T) {
...
-					proxyApp := user.getApplication(t, proxyCl, applicationName)
 					// Update DisplayName
 					changedDisplayName := fmt.Sprintf("Proxy test for user %s - updated application", tenantNsName(user.compliantUsername))
-					proxyApp.Spec.DisplayName = changedDisplayName
-					err := proxyCl.Update(context.TODO(), proxyApp)
+					// Update application
+					proxyApp, err := awaitilities.Host().UpdateApplication(t, proxyCl, applicationName, tenantNsName(user.compliantUsername), func(app *appstudiov1.Application) {
+						app.Spec.DisplayName = changedDisplayName
+					})
 					require.NoError(t, err)
```
MatousJobanek, line 244: Not related to this PR, but I think that we can drop the "AppStudio API"-specific test cases. It's completely fine to test modification on standard k8s resources (eg. ConfigMap), no need to maintain the other API

test/e2e/parallel/socialevent_test.go
```diff
@@ -79,7 +78,10 @@ func TestCreateSocialEvent(t *testing.T) {
 			event.Spec.UserTier = "deactivate30"
```
MatousJobanek, line 78: you can drop this line, right?

test/e2e/parallel/socialevent_test.go
```diff
@@ -115,7 +117,10 @@ func TestCreateSocialEvent(t *testing.T) {
 			event.Spec.SpaceTier = "base"
```
MatousJobanek, line 117: and this one too

test/e2e/parallel/proxy_test.go
MatousJobanek: Apart from that, there is actually no Application controller running that would watch/modify these resources. This means that we don't expect any conflicts for this resource and the update through the proxy should work fine without having the loop.

testsupport/spaceprovisionerconfig/spaceprovisionerconfig.go
```diff
@@ -54,12 +57,16 @@ func UpdateForCluster(t *testing.T, await *wait.Awaitility, referencedClusterNam
...
-		// make the originalSpc look like we freshly obtained it from the server and updated its fields
-		// to look like the original.
-		originalSpc.Generation = currentSpc.Generation
-		originalSpc.ResourceVersion = currentSpc.ResourceVersion
+		_, err = await.UpdateSpaceProvisionerConfig(t, originalSpc.Name,
+			func(spcfg *toolchainv1alpha1.SpaceProvisionerConfig) {
+				// make the originalSpc look like we freshly obtained it from the server and updated its fields
+				// to look like the original.
+				spcfg.Generation = currentSpc.Generation
+				spcfg.ResourceVersion = currentSpc.ResourceVersion
```
MatousJobanek, line 65: there is no need for this hack, the loop ensures that it gets the fresh object
```suggestion
```

testsupport/wait/host.go
```diff
@@ -2574,3 +2583,60 @@ func (a *HostAwaitility) CreateSpaceAndSpaceBinding(t *testing.T, mur *toolchain
...
+		socialEvent = freshSocialEvent
+		return true, nil
+	})
+	return socialEvent, err
+}
+
+// NSTemplateTier updates the given NSTemplateTier using the provided modifiers.
+// If it encounters an error (e.g., object has been modified), it retrieves the latest version and retries.
+// Returns the updated NSTemplateTier.
+func (hostAwait *HostAwaitility) UpdateNSTemplateTier(t *testing.T, tierName string, modifyTier func(tier *toolchainv1alpha1.NSTemplateTier)) (*toolchainv1alpha1.NSTemplateTier, error) {
```
MatousJobanek, line 2621: the receiver names should stay consistent for all methods
```suggestion
func (a *HostAwaitility) UpdateNSTemplateTier(t *testing.T, tierName string, modifyTier func(tier *toolchainv1alpha1.NSTemplateTier)) (*toolchainv1alpha1.NSTemplateTier, error) {
```

MatousJobanek (commented):
Thanks a lot for taking care of this. 
I believe that we could lower the amount of copy-pasting by expanding the `Waiter` generic type https://github.com/codeready-toolchain/toolchain-e2e/blob/f22abed08672f8d820cba7a24d9f5398650e24a1/testsupport/wait/awaitility.go#L669-L675
cc @metlos 
Something like this could work:
```go 
func (w *Waiter[T]) Update(status bool, objectName string, modify func(T)) (T, error) {
	var objectToReturn T
	err := wait.PollUntilContextTimeout(context.TODO(), w.await.RetryInterval, w.await.Timeout, true, func(ctx context.Context) (done bool, err error) {
		obj := &unstructured.Unstructured{}
		obj.SetGroupVersionKind(w.gvk)
		if err := w.await.Client.Get(context.TODO(), types.NamespacedName{Namespace: w.await.Namespace, Name: objectName}, obj); err != nil {
			return true, err
		}
		object, err := w.cast(obj)
		if err != nil {
			return false, fmt.Errorf("failed to cast the object to GVK %v: %w", w.gvk, err)
		}
		modify(object)
		if status {
			// Update the Status
			if err := w.await.Client.Status().Update(context.TODO(), object); err != nil {
				w.t.Logf("error updating '%v' Status '%s': %s. Will retry again...", w.gvk, objectName, err.Error())
				return false, err
			}
		} else {
			// Update the Spec
			if err := w.await.Client.Update(context.TODO(), object); err != nil {
				w.t.Logf("Error updating '%v' Spec '%s': %s. Will retry again...", w.gvk, objectName, err.Error())
				return false, err
			}
		}
		objectToReturn = object
		return true, nil
	})
	return objectToReturn, err
}
```
so you could then call it:
```go
_, err = For(t, hostAwait.Awaitility, &toolchainv1alpha1.NSTemplateTier{}).
	Update(false, tier.NSTemplateTier.Name, func(nstt *toolchainv1alpha1.NSTemplateTier) {
		nstt.Spec = tier.NSTemplateTier.Spec
	})
```
or:
```go
event, err = For(t, hostAwait.Awaitility, &toolchainv1alpha1.SocialEvent{}).
	Update(false, event.Name, func(ev *toolchainv1alpha1.SocialEvent) {
		ev.Spec.SpaceTier = "base"
	})
```

You could also add other methods with some syntactic sugar:
```go
func (w *Waiter[T]) Update(objectName string, modify func(T)) (T, error) {
	return w.doUpdate(false, objectName, modify)
}

func (w *Waiter[T]) UpdateStatus(objectName string, modify func(T)) (T, error) {
	return w.doUpdate(true, objectName, modify)
}

func (w *Waiter[T]) doUpdate(status bool, objectName string, modify func(T)) (T, error) {
	var objectToReturn T
...
```

test/e2e/parallel/proxy_test.go
rsoaresd: Thank you so much! I will remove the loop here. And take care of the "AppStudio API"-specific test cases in a different PR

testsupport/spaceprovisionerconfig/spaceprovisionerconfig.go
rsoaresd: My bad!! Thanks for noticing it!

rsoaresd:
> Thanks a lot for taking care of this. I believe that we could lower the amount of copy-pasting by expanding the `Waiter` generic type
> 
> https://github.com/codeready-toolchain/toolchain-e2e/blob/f22abed08672f8d820cba7a24d9f5398650e24a1/testsupport/wait/awaitility.go#L669-L675
> 
> 
> cc @metlos
> Something like this could work:
> ```go
> func (w *Waiter[T]) Update(status bool, objectName string, modify func(T)) (T, error) {
> 	var objectToReturn T
> 	err := wait.PollUntilContextTimeout(context.TODO(), w.await.RetryInterval, w.await.Timeout, true, func(ctx context.Context) (done bool, err error) {
> 		obj := &unstructured.Unstructured{}
> 		obj.SetGroupVersionKind(w.gvk)
> 		if err := w.await.Client.Get(context.TODO(), types.NamespacedName{Namespace: w.await.Namespace, Name: objectName}, obj); err != nil {
> 			return true, err
> 		}
> 		object, err := w.cast(obj)
> 		if err != nil {
> 			return false, fmt.Errorf("failed to cast the object to GVK %v: %w", w.gvk, err)
> 		}
> 		modify(object)
> 		if status {
> 			// Update the Status
> 			if err := w.await.Client.Status().Update(context.TODO(), object); err != nil {
> 				w.t.Logf("error updating '%v' Status '%s': %s. Will retry again...", w.gvk, objectName, err.Error())
> 				return false, err
> 			}
> 		} else {
> 			// Update the Spec
> 			if err := w.await.Client.Update(context.TODO(), object); err != nil {
> 				w.t.Logf("Error updating '%v' Spec '%s': %s. Will retry again...", w.gvk, objectName, err.Error())
> 				return false, err
> 			}
> 		}
> 		objectToReturn = object
> 		return true, nil
> 	})
> 	return objectToReturn, err
> }
> ```
> 
> so you could then call it:
> 
> ```go
> _, err = For(t, hostAwait.Awaitility, &toolchainv1alpha1.NSTemplateTier{}).
> 	Update(false, tier.NSTemplateTier.Name, func(nstt *toolchainv1alpha1.NSTemplateTier) {
> 		nstt.Spec = tier.NSTemplateTier.Spec
> 	})
> ```
> 
> or:
> 
> ```go
> event, err = For(t, hostAwait.Awaitility, &toolchainv1alpha1.SocialEvent{}).
> 	Update(false, event.Name, func(ev *toolchainv1alpha1.SocialEvent) {
> 		ev.Spec.SpaceTier = "base"
> 	})
> ```
> 
> You could also add other methods with some syntactic sugar:
> 
> ```go
> func (w *Waiter[T]) Update(objectName string, modify func(T)) (T, error) {
> 	return w.doUpdate(false, objectName, modify)
> }
> 
> func (w *Waiter[T]) UpdateStatus(objectName string, modify func(T)) (T, error) {
> 	return w.doUpdate(true, objectName, modify)
> }
> 
> func (w *Waiter[T]) doUpdate(status bool, objectName string, modify func(T)) (T, error) {
> 	var objectToReturn T
> ...
> ```

@MatousJobanek excellent idea!! 🚀 Should we move the `Waiter` generic type to toolchain-common or remain here?

MatousJobanek:
as discussed in the KubeSaw sync call, it should stay here :+1:

testsupport/wait/awaitility.go
```diff
@@ -894,65 +893,51 @@ func For[T client.Object](t *testing.T, a *Awaitility, obj T) *Waiter[T] {
...
-		modifyIdentity(freshIdentity)
-		if err := a.Client.Update(context.TODO(), freshIdentity); err != nil {
-			t.Logf("error updating Identity '%s': %s.Will retry again...", identityName, err.Error())
-			return false, nil
-		}
-		identity = freshIdentity
-		return true, nil
-	})
-	return identity, err
+// Update tries to update the Spec of the given object
```
MatousJobanek, line 931: it can update also labels and annotations, the only thing it cannot update is the `status` sub-resource, thus I would drop the `Spec` part
```suggestion
// Update tries to update the given object
```

MatousJobanek (approved):
Nice, thanks for taking care of this, and thanks for implementing the generic approach :+1:

testsupport/wait/awaitility.go
```diff
@@ -892,3 +892,52 @@ func For[T client.Object](t *testing.T, a *Awaitility, obj T) *Waiter[T] {
...
+			// Update the Status
+			if err := w.await.Client.Status().Update(context.TODO(), object); err != nil {
+				w.t.Logf("error updating '%v' Status '%s': %s. Will retry again...", w.gvk, objectName, err.Error())
+				return false, err
+			}
+		} else {
+			// Update the Spec
+			if err := w.await.Client.Update(context.TODO(), object); err != nil {
+				w.t.Logf("Error updating '%v' Spec '%s': %s. Will retry again...", w.gvk, objectName, err.Error())
+				return false, err
```
MatousJobanek, line 922: I just noticed that I made a mistake in my suggestion, we should not return an error here; otherwise, the poll loop immediately exists with an error:
```suggestion
				return false, nil
			}
		} else {
			// Update the Spec
			if err := w.await.Client.Update(context.TODO(), object); err != nil {
				w.t.Logf("Error updating '%v' Spec '%s': %s. Will retry again...", w.gvk, objectName, err.Error())
				return false, nil
```

mfrancisc: is this true also for the status update faliure : https://github.com/codeready-toolchain/toolchain-e2e/pull/1074/files#diff-4abc90985c7fe1e595e0d6f5131e1f4757a6f00c9239f40ae09375d12f15e530R916

should we not return an error here as well and retry ?

rsoaresd: Yes, it is addressed in line 916 of @MatousJobanek  suggestion! Goinggggg!!

Locally, it passed for me. That's why I did not notice it. Sorry!

mfrancisc: oh nice, sorry I've missed that from the suggestion, never mind 👍

test/e2e/parallel/space_test.go
```diff
@@ -222,11 +222,11 @@ func TestSpaceRoles(t *testing.T) {
 	})
 
 	t.Run("set owner user as contributor instead", func(t *testing.T) {
-		// given an appstudio space with `owner` user as an admin of it
-		ownerBinding.Spec.SpaceRole = "contributor"
-
 		// when
-		err := hostAwait.Client.Update(context.TODO(), ownerBinding)
+		ownerBinding, err = hostAwait.UpdateSpaceBinding(t, ownerBinding.Name, func(sb *toolchainv1alpha1.SpaceBinding) {
```
mfrancisc, line 226: I was wondering if there something special with `UpdateSpaceBinding` and `UpdateSpace`, or they could be replaced by the new generic update function as well ? 

Not something we need to address in this PR anyways.

mfrancisc (approved):
Great Job 👏 
I like the new generic update functions 👍 

I have only one minor comment/question.

test/e2e/parallel/space_test.go
MatousJobanek: As mentioned in the other comment, they will be replaced in the following PR https://github.com/codeready-toolchain/toolchain-e2e/pull/1074#issuecomment-2511409880

mfrancisc: Got it , I've missed that ! Thanks 👍

rsoaresd:
/retest

hitting known flaky test [SANDBOX-837](https://issues.redhat.com/browse/SANDBOX-837)

PR #1075, opened 2024-11-29
SANDBOX-858: make getAndPrint exportable

rsoaresd:
# Description
This PR makes `getAndPrint` exportable. It will be useful for wa

[SANDBOX-858](https://issues.redhat.com/browse/SANDBOX-858)

### DIFF
--- a/testsupport/wait/awaitility.go
+++ b/testsupport/wait/awaitility.go
@@ -474,7 +474,7 @@ func (a *Awaitility) WaitForDeploymentToGetReady(t *testing.T, name string, repl
 		return true, nil
 	})
 	if err != nil {
-		a.getAndPrint(t, "Deployment", a.Namespace, name, &appsv1.Deployment{})
+		a.GetAndPrint(t, "Deployment", a.Namespace, name, &appsv1.Deployment{})
 	}
 	require.NoError(t, err)
 	return deployment
@@ -654,7 +654,7 @@ func (a *Awaitility) listAndReturnContent(resourceKind, namespace string, list c
 	return fmt.Sprintf("\n%s present in the namespace:\n%s\n", resourceKind, string(content))
 }
 
-func (a *Awaitility) getAndPrint(t *testing.T, resourceKind, namespace, name string, obj client.Object, additionalOptions ...client.GetOption) {
+func (a *Awaitility) GetAndPrint(t *testing.T, resourceKind, namespace, name string, obj client.Object, additionalOptions ...client.GetOption) {
 	t.Logf(a.getAndReturnContent(resourceKind, namespace, name, obj, additionalOptions...))
 }
 


mfrancisc (approved):
minor - should we move the packages used by WA to toolchain-common or some new library/repo ?

I'm asking this since importing from toolchain-e2e might be problematic - for example in case toolchain-e2e and WA depend on different versions of `api` and `toolchain-common`. It is very unlikely that this might happen with the current workflow, since we have the habit to keep those up to date in all repositories, but there is no restriction or enforcement on doing that.

alexeykazakov (approved):

MatousJobanek:
Good point @mfrancisc.
Theoretically, moving the code to toolchain-common would make complete sense and would follow what we've been doing in all other cases till now. 
However, practically it wouldn't make much of a difference in this case. WA code will rely also on other parts of the toolchain-e2e code (makefile targets, `WaitForDeployments` func, awatitilities, and maybe more). And moving all these things to toolchain-common doesn't make much sense. 
Also, when you think about the usage from the WA point of view, it's something different compared to the KubeSaw components. WA uses toolchain-e2e to setup KubeSaw in the e2e environment, it doesn't use it to test KubeSaw functionality, so there is no risk of messing up with the proper separation of concerns.

That being said, I think it's fine to have the direct dependency from WA to toolchain-e2e repo

mfrancisc:
> in go code where the version of the toolchain-e2e (that is being used) is defined in go.mod

this type of dependency could break as well, if you need to update the `toolchain-e2e` version and the new version has changes affecting the functions used by WA.

What I'm saying is that there is no way for someone doing changes on `toolchain-e2e` to figure out if the changes are still compatible with how WA uses this functionality, or to know which functionality is actually used in WA, unless it opens a PR in WA with the new version of `toolchain-e2e`. 
But again - I understand that extracting those reusable functionality to a common library/repo might be premature now, and that getting WA to a production state has higher priority right now. For now we can live with it knowing that there is `toolchain-e2e` functionality used in WA, and it's worth checking before/while working on changes for the e2e tests ( same thing as we do when working at changes for `api` and `toolchain-common` repo ) .

rsoaresd:
@mfrancisc @MatousJobanek, thanks a lot for this discussion and your insights!

I thought about reusing toolchain-e2e make targets and functions (`WaitForDeployments`, `NewSignupRequest`, etc.) to not duplicate "unnecessary" code on wa side and to make the implementation faster. As both (make targets and these functions) are something that we do not change often, the risk of something going wrong is less. But, of course, the risk is there.

The "most critical" aspect is indeed the setup, as @MatousJobanek mentioned. We rely on the `deploy-single-member-e2e` to deploy the toolchain e2e resources on wa side.

I think we are good for now. But, in the future, we need to reconsider if we should move the logic to a common repo.

alexeykazakov:
I would try to re-use as much form the e2e repo in the wa tests as possible. So +1 to not duplicating the code but depending on / reusing it.
However it mean that we can always break something. But it's not different than depending on common or anything else (I'm talking about go.mod dependencies, not make targets). You can now commit something into common which would break downstream decencies IF you upgrade the dependency to the latest common.

PR #1076, opened 2024-11-29
Add a simple e2e test for the presence of consumed capacity in SPCs

metlos:
Add a simple e2e test for the presence of consumed capacity in the space provisioner configs. We cannot do much more because we cannot mess with the registered members, nor can we add/remove new members.

Also, tests for appearance and disappearance of toolchain clusters and the reaction of SPCs on that were removed, because the readiness of the SPCs is now more complex and cannot be easily tested in e2e tests without the ability to also add/remove whole new members clusters. This functionality is extensively covered in the host-operator unit tests though.

Paired PRs:
- host-operator: https://github.com/codeready-toolchain/host-operator/pull/1108

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,7 +1,7 @@
 module github.com/codeready-toolchain/toolchain-e2e
 
 require (
-	github.com/codeready-toolchain/api v0.0.0-20241114213029-44333bf24bcf
+	github.com/codeready-toolchain/api v0.0.0-20241119094246-f6581d52dc80
 	github.com/codeready-toolchain/toolchain-common v0.0.0-20241114215157-a6a85252b2f5
 	github.com/davecgh/go-spew v1.1.1
 	github.com/fatih/color v1.15.0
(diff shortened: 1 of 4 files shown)


mfrancisc (commented):
Looks good 👍 

I do have one question - why can't we keep using the current approach that in the tests , I'm referring to copying the toolchainclusters and creating new SPC's with the reference to the copy? I see that the SPC status is influenced by the TC ready condition : https://github.com/codeready-toolchain/host-operator/pull/1108/files#diff-9b844eb6c3d858337767be70add36b6915bd0d4dcc4df8cee332b5e9eb057dcdR169 .

metlos:
> Looks good 👍
> 
> I do have one question - why can't we keep using the current approach that's in the tests now , I'm referring to copying the toolchainclusters and creating new SPC's with the reference to the copy? I see that the SPC status is influenced by the TC ready condition : https://github.com/codeready-toolchain/host-operator/pull/1108/files#diff-9b844eb6c3d858337767be70add36b6915bd0d4dcc4df8cee332b5e9eb057dcdR169 .

The readiness of the SPC was before only reliant on the status of the TC. You could have a disabled SPC that was ready. This enabled us to copy SPCs and TCs without affecting the state of the cluster for the rest of the parallel tests. E.g. Having an additional TC, in effect a new member cluster, wouldn't affect the rest of the testsuite because the SPC was always disabled in these tests and therefore these new "synthetic" members wouldn't participate in the provisioning.

With the SPC readiness now being a true reflection of whether the associated cluster can be provisioned to, this no longer is true. Copying the TC and having a ready SPC for it would make it eligible for provisioning, essentially creating another member. This would mess up other tests that assume there are 2 members.

mfrancisc (approved):
Great Job 👍 

Thanks for explaining the reason behind the simplification of the e2e test. I agree that unit tests should suffice.

MatousJobanek (approved):
Nice simplification :+1:

MatousJobanek:
@metlos you could add a check to the [usersignup_test.go](https://github.com/codeready-toolchain/toolchain-e2e/blob/master/test/e2e/usersignup_test.go#L48) verifying that when the threshold is reached, then the SPC CR is put into an unready state. We already have these test cases there, so it's just matter of adding the checks of the condition of the SPC CRs. And it's completely sufficient to do it in one or two cases, we don't need to update all the cases there.

metlos:
Closing this in favor of https://github.com/codeready-toolchain/toolchain-e2e/pull/1079

PR #1077, opened 2024-12-03
SANDBOX-770: fix WaitForBannedUser

rsoaresd:
# Description
This function is not used in toolchain-e2e. However, it will be useful for WA.
Unfortunately, it was not working as expected since when no error on list, it was accessing `&bannedUserList.Items[0]` without checking the array length.

[SANDBOX-770](https://issues.redhat.com/browse/SANDBOX-770)

### DIFF
--- a/testsupport/wait/host.go
+++ b/testsupport/wait/host.go
@@ -771,7 +771,7 @@ func (a *HostAwaitility) WaitAndVerifyThatUserSignupIsNotCreated(t *testing.T, n
 // WaitForBannedUser waits until there is a BannedUser available with the given email hash
 // !!! WARNING: for now, just used for WA
 func (a *HostAwaitility) WaitForBannedUser(t *testing.T, userEmailHash string) (*toolchainv1alpha1.BannedUser, error) {
-	t.Logf("waiting for BannedUser for user '%s' in namespace '%s'", userEmailHash, a.Namespace)
+	t.Logf("waiting for BannedUser for user email hash '%s' in namespace '%s'", userEmailHash, a.Namespace)
 	var bannedUser *toolchainv1alpha1.BannedUser
 	emailHashLabelMatch := client.MatchingLabels(map[string]string{
 		toolchainv1alpha1.BannedUserEmailHashLabelKey: userEmailHash,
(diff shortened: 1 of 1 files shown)


alexeykazakov (approved):
Alternately we could move it to WA but I'm fine with keeping it here in case we need it use in the main e2e tests later too.

testsupport/wait/host.go
```diff
@@ -770,24 +769,27 @@ func (a *HostAwaitility) WaitAndVerifyThatUserSignupIsNotCreated(t *testing.T, n
 }
 
 // WaitForBannedUser waits until there is a BannedUser available with the given email
-func (a *HostAwaitility) WaitForBannedUser(t *testing.T, email string) (*toolchainv1alpha1.BannedUser, error) {
-	t.Logf("waiting for BannedUser for user '%s' in namespace '%s'", email, a.Namespace)
+// !!! WARNING: for now, just used for WA
+func (a *HostAwaitility) WaitForBannedUser(t *testing.T, userEmailHash string) (*toolchainv1alpha1.BannedUser, error) {
+	t.Logf("waiting for BannedUser for user '%s' in namespace '%s'", userEmailHash, a.Namespace)
```
mfrancisc, line 774: probably you'll want to print the decoded email here , but not sure.

testsupport/wait/host.go
```diff
@@ -770,24 +769,27 @@ func (a *HostAwaitility) WaitAndVerifyThatUserSignupIsNotCreated(t *testing.T, n
 }
 
 // WaitForBannedUser waits until there is a BannedUser available with the given email
```
mfrancisc, line 771:
```suggestion
// WaitForBannedUser waits until there is a BannedUser available with the given email hash
```
we could update the comment since now it expects an email hash and not email if I'm seeing this correctly

mfrancisc (approved):
Looks good 👍

rsoaresd:
> Alternately we could move it to WA but I'm fine with keeping it here in case we need it use in the main e2e tests later too.

@alexeykazakov I did not move it because I do not know if that function is there for some reason. I tried to track when it was added but i was not able. i think it was added a long time ago

testsupport/wait/host.go
rsoaresd: Yes, thanks! I missed!

testsupport/wait/host.go
rsoaresd: There is no way, I think. User email hash is encoded in MD5: https://github.com/codeready-toolchain/toolchain-common/blob/master/pkg/hash/encode.go

PR #1079, opened 2024-12-06
Add a simple e2e test for the presence of consumed capacity

metlos:
Add a simple e2e test for the presence of consumed capacity in the space provisioner configs. We cannot do much more because we cannot mess with the registered members, nor can we add/remove new members.

Paired PRs:
- host-operator: https://github.com/codeready-toolchain/host-operator/pull/1109

Note to self:
- [x] make sure all the comments from https://github.com/codeready-toolchain/toolchain-e2e/pull/1076 are addressed


test/e2e/usersignup_test.go
```diff
@@ -533,8 +534,13 @@ func (s *userSignupIntegrationTest) TestCapacityManagementWithManualApproval() {
 			Execute(s.T())
 		userSignup := user.UserSignup
 
+		_, spcNotReadyError := wait.For(t, hostAwait.Awaitility, &toolchainv1alpha1.SpaceProvisionerConfig{}).FirstThat(
+			assertions.Has(spaceprovisionerconfig.ReferenceToToolchainCluster(memberAwait1.ClusterName)),
+			assertions.Is(testSpc.NotReady()))
+
 		// then
 		s.userIsNotProvisioned(t, userSignup)
+		require.NoError(t, spcNotReadyError)
```
MatousJobanek, line 543: Is there any reason for not checking the error right after calling the function?
```suggestion
			assertions.Is(testSpc.NotReady()))
			require.NoError(t, spcNotReadyError)

		// then
		s.userIsNotProvisioned(t, userSignup)
```

test/e2e/usersignup_test.go
```diff
@@ -566,8 +572,13 @@ func (s *userSignupIntegrationTest) TestCapacityManagementWithManualApproval() {
...
```
MatousJobanek, line 581: the same

test/e2e/usersignup_test.go
```diff
@@ -602,6 +613,43 @@ func (s *userSignupIntegrationTest) TestCapacityManagementWithManualApproval() {
 
 		assert.Equal(t, toolchainv1alpha1.UserSignupStateLabelValueApproved, user.UserSignup.Labels[toolchainv1alpha1.UserSignupStateLabelKey])
 	})
+
+	s.T().Run("last provisioned space flips SpaceProvisionerConfig to not ready", func(t *testing.T) {
+		// given
+		// there are 4 already deployed spaces before this test is run, so we set the capacity to 5 so that 1 last space can be deployed.
```
MatousJobanek, line 619: I don't think that this will really work as the Spaces were deployed in other sub-tests and the users should be cleaned automatically at the end of each of the sub-tests.
This may also lead to flakiness.
I would either transform it to a separate test:
```go
func (s *userSignupIntegrationTest) TestSpaceProvisionerConfigFlip() {
	hostAwait := s.Host()
	memberAwait1 := s.Member1()
	memberAwait2 := s.Member2()
	// given
	s.T().Run("last provisioned space flips SpaceProvisionerConfig to not ready", func(t *testing.T) {
		// given
		spaceprovisionerconfig.UpdateForCluster(t, hostAwait.Awaitility, memberAwait1.ClusterName, testSpc.MaxNumberOfSpaces(1))
		spaceprovisionerconfig.UpdateForCluster(t, hostAwait.Awaitility, memberAwait2.ClusterName, testSpc.Enabled(false))
```
or you can reuse an already existing test that does something similar
eg: https://github.com/codeready-toolchain/toolchain-e2e/blob/59268cf3699545643a490c841ea18fcdd05119c8/test/e2e/usersignup_test.go#L241-L282
where after updating the SPC you can also check the readiness:
```go
spaceprovisionerconfig.UpdateForCluster(t, hostAwait.Awaitility, memberAwait1.ClusterName, testSpc.MaxNumberOfSpaces(1))
spaceprovisionerconfig.UpdateForCluster(t, hostAwait.Awaitility, memberAwait2.ClusterName, testSpc.MaxNumberOfSpaces(1))
_, err := wait.For(t, hostAwait.Awaitility, &toolchainv1alpha1.SpaceProvisionerConfig{}).FirstThat(
	assertions.Has(spaceprovisionerconfig.ReferenceToToolchainCluster(memberAwait1.ClusterName)),
	assertions.Is(testSpc.Ready()))
require.NoError(t, err)
_, err = wait.For(t, hostAwait.Awaitility, &toolchainv1alpha1.SpaceProvisionerConfig{}).FirstThat(
	assertions.Has(spaceprovisionerconfig.ReferenceToToolchainCluster(memberAwait2.ClusterName)),
	assertions.Is(testSpc.Ready()))
require.NoError(t, err)
```
and then expect the flip to an unready state after provisioning the users

test/e2e/usersignup_test.go
metlos: The reason is just the visual when/then separation...

MatousJobanek: ok, then at least swap it, it's fine to put it into the `// then` section 
```suggestion
			assertions.Is(testSpc.NotReady()))

		// then
		require.NoError(t, spcNotReadyError)
		s.userIsNotProvisioned(t, userSignup)
```
but not after calling `s.userIsNotProvisioned(t, userSignup)`. When I saw it, I was wondering how is the execution of `s.userIsNotProvisioned` related to the returned error :-)

test/e2e/usersignup_test.go
metlos: Extracted this into a separate top-level test in the testsuite.

test/e2e/usersignup_test.go
metlos: The new test looks a bit different so this is no longer applicable.

metlos: uh, no, the above comment is completely wrong :)

test/e2e/usersignup_test.go
```diff
@@ -281,6 +281,47 @@ func (s *userSignupIntegrationTest) TestProvisionToOtherClusterWhenOneIsFull() {
...
+		// when
+		user1 := NewSignupRequest(s.Awaitilities).
+			Username("fill-up-user-1").
+			Email("fillup1@redhat.com").
+			EnsureMUR().
+			RequireConditions(wait.ConditionSet(wait.Default(), wait.ApprovedAutomatically())...).
+			Execute(s.T())
+
+			// then
+		VerifyResourcesProvisionedForSignup(t, s.Awaitilities, user1.UserSignup)
```
MatousJobanek, line 317: it's not really obvious, but calling the `EnsureMUR` triggers verification of the provisioned resources for the signup, so no need to do it for the second time.
apart from that, verifying resources is not important in this case so you can just wait until the MUR is present and slightly speed the test up:
```suggestion
			WaitForMUR().
			RequireConditions(wait.ConditionSet(wait.Default(), wait.ApprovedAutomatically())...).
			Execute(s.T())
			
		// then
```

MatousJobanek (approved):
Thanks :+1:

fbm3307 (approved):

PR #1090, opened 2025-01-08
fix(flaky): ensure that the space is always unblocked for its deletion

MatousJobanek:
make sure that the space is unblocked for the cleanup, even for the cases when the test execution is stopped in the middle and the sub-test (that is supposed to unblock the space) is not triggered: https://redhat-internal.slack.com/archives/C06MJ2DVBU4/p1736342796950679?thread_ts=1736333943.812749&cid=C06MJ2DVBU4

### DIFF
--- a/test/e2e/parallel/space_test.go
+++ b/test/e2e/parallel/space_test.go
@@ -77,6 +77,8 @@ func TestCreateSpace(t *testing.T) {
 		// then
 		VerifyResourcesProvisionedForSpace(t, awaitilities, space.Name)
 
+		// ensure that the space is reset back to the original (valid) target cluster
+		// so the cleanup logic can delete the Space
 		var resetOnce sync.Once
 		reset := func() {
 			resetOnce.Do(func() {
--- a/test/e2e/parallel/space_test.go
+++ b/test/e2e/parallel/space_test.go
@@ -3,6 +3,7 @@ package parallel
 import (
 	"context"
 	"sort"
+	"sync"
 	"testing"
 
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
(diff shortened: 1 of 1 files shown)


fbm3307 (approved):
Thanks for the quick fix

test/e2e/parallel/space_test.go
```diff
@@ -77,6 +77,20 @@ func TestCreateSpace(t *testing.T) {
 		// then
 		VerifyResourcesProvisionedForSpace(t, awaitilities, space.Name)
 
+		// ensure that the space is reset back to the original (valid) target cluster
+		// so the cleanup logic can delete the Space
+		var resetOnce sync.Once
```
alexeykazakov, line 82: Hm... what's the point of using `Once` here? resetOnce.Do() will be called a single time only (for this specific resetOnce instance) anyway. What am I missing?

MatousJobanek: or were you were asking why do I define the variable `resetOnce` outside of the function? Well, the variable has to stay outside, so when the function is called, it always uses the same instance of the `Once`. If it was moved inside of the function, then it would create a new instance of `Once` for every call of the function

mfrancisc: should not be a problem - but should we move this definition in the parent test `TestCreateSpace` ( where we define `awaitilities` , `hostAwait`, `memberAwait` ) ? Just to be on the safe side.

mfrancisc (approved):
Nice 👍

rajivnathan (approved):

test/e2e/parallel/space_test.go
alexeykazakov: What I was missing is that the `reset()` was called twice (not once!). One in the defer and the second time in the sub-test bellow. I saw it only in called by defer and was wondering why do you need `once` if it's already called only once :)
Sorry for the confusion.

alexeykazakov (approved):

test/e2e/parallel/space_test.go
MatousJobanek: > should we move this definition in the parent test TestCreateSpace ( where we define awaitilities , hostAwait, memberAwait ) ? Just to be on the safe side.

not really, it has to stay in the same (sub)test  where the space is provisioned, so it's tied to the actual space (and also the the function that executes this sub-test)

PR #1091, opened 2025-01-08
fix(flaky): ensure that the CM finalizer is always removed

MatousJobanek:
the same idea as for https://github.com/codeready-toolchain/toolchain-e2e/pull/1090

### DIFF
--- a/test/e2e/parallel/e2e_test.go
+++ b/test/e2e/parallel/e2e_test.go
@@ -2,6 +2,7 @@ package parallel
 
 import (
 	"context"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


test/e2e/parallel/e2e_test.go
```diff
@@ -284,6 +285,20 @@ func TestE2EFlow(t *testing.T) {
 		require.NoError(t, err)
 		require.NotEmpty(t, cm)
 
+		// ensure that the finalizer is removed in all cases, so the cleanup logic
+		// is not blocked and can delete the Space
+		var removeOnce sync.Once
```
mfrancisc, line 290: same as in the other PR - it should be safe but maybe we could move this definition up one level.

mfrancisc (approved):
Thanks for fixing it 👍

test/e2e/parallel/e2e_test.go
MatousJobanek: and the same answer - it should stay in the save sub-test where the ConfigMap is provisoned & modified

PR #1093, opened 2025-01-13
Replace the --lets-encrypt flag with --insecure-skip-tls-verify flag

metlos:
This is to update toolchain-e2e to the new flag in ksctl.

See https://github.com/kubesaw/ksctl/pull/95.

### DIFF
--- a/make/test.mk
+++ b/make/test.mk
@@ -33,8 +33,12 @@ endif
 
 E2E_TEST_EXECUTION ?= true
 
-ifneq ($(IS_OSD),true)
-LETS_ENCRYPT_PARAM := --lets-encrypt=false
+# let's always be explicit about TLS verify param so that we don't have to rely on the values
+# set in the kubeconfig (if any)
+ifeq ($(IS_OSD),true)
+TLS_VERIFY_PARAM := --insecure-skip-tls-verify=false
+else
+TLS_VERIFY_PARAM := --insecure-skip-tls-verify=true
 endif
 
 TESTS_RUN_FILTER_REGEXP ?= ""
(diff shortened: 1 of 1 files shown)


make/test.mk
```diff
@@ -33,8 +33,12 @@ endif
...
 E2E_TEST_EXECUTION ?= true
 
-ifneq ($(IS_OSD),true)
-LETS_ENCRYPT_PARAM := --lets-encrypt=false
+# let's always be explicit about TLS verify param so that we don't have to rely on the values
+# set in the kubeconfig (if any)
+ifeq ($(IS_OSD),true)
+TLS_VERIFY_PARAM := --insecure-skip-tls-verify=false
+else
+TLS_VERIFY_PARAM := --insecure-skip-tls-verify=true
```
mfrancisc, line 41: should we disable tls verification by default when running the tests "locally" ? Or should we document this and let the user decide ?

MatousJobanek: correct me if I'm wrong, if the flag is not specified, then it would use whatever is in the client's kubeconfig. The flag should be used only if you want to override (for some specific reason) the information in kubeconfig.

This means that the default behavior shouldn't use the flag at all, only if requested explicitly via a variable.

mfrancisc: +1 I thought the same, but maybe I've missed something.

metlos: I wanted to keep the same behavior as there was with the `--lets-encrypt` flag. The main reason being that I'm not fully sure about the consequences of relying on the contents of the kubeconfig.

If the intended behavior should change, I'm fine with it though. As far as I looked we ALWAYS set the `--lets-encrypt=false` in all environments (e.g. CI or default when running locally) - I haven't found a place where we'd set `IS_OSD=true` anywhere.

To keep this behavior, we'd need to set `--insecure-skip-tls-verify=false` only when `IS_OSD=true` and we should be explicit about it. 

I don't know what the kubeconfig looks like in the CI (whether it explicitly sets the insecure-skip-tls-verify or whether the connection to the clusters in the CI doesn't require it) so I didn't want to risk wreaking havoc in all our CI jobs.

MatousJobanek: > I haven't found a place where we'd set IS_OSD=true anywhere.

that was used for local dev use-cases. Some teams/developers had OSD cluster for the local dev (instead of a pure OCP)

> To keep this behavior, we'd need to set --insecure-skip-tls-verify=false only when IS_OSD=true and we should be explicit about it.

if it's OSD and you (as an admin) execute the register-member command, then you don't have this field defined in your kubeconfig, nor the cert as the OSD cluster uses let's encrypt cert which is a well-know one.

> I don't know what the kubeconfig looks like in the CI (whether it explicitly sets the insecure-skip-tls-verify or whether the connection to the clusters in the CI doesn't require it) 

does it matter? if the kubeconfig has `insecure-skip-tls-verify` defined, then the command copies it, if not, then it should have the self-sign certificate, which should be copied as well.

Let's do what we wanted to do and what was the reason for these changes - rely on the content of the kubeconfig if not explicitly defined otherwise.

metlos: I discovered we already have `DISABLE_KUBE_CLIENT_TLS_VERIFY` to have this behavior for the kubeclient used in the tests. I think we can use this also to specify whether to use `--insecure-skip-tls-verify=true` for the ksctl. I've updated the make files to do this.

mfrancisc (approved):
Looks good 👍 

Thanks for the additional changes.

fbm3307 (approved):

PR #1111, opened 2025-02-10
fix: panic for concurrent writing in parallel test (scheme)

fbm3307:
This PR is for the fix of e2e [Failures](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/codeready-toolchain_host-operator/1130/pull-ci-codeready-toolchain-host-operator-master-e2e/1887805053301952512/build-log.txt) .

The parallel e2e test are making concurrent writes to the global `Scheme` variable, used in the package `testsupport/wait` .

This PR ensures that a new Scheme is always used instead of the global one.

### DIFF
--- a/testsupport/wait/host.go
+++ b/testsupport/wait/host.go
@@ -33,7 +33,6 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
-	"k8s.io/kubectl/pkg/scheme"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
@@ -1734,7 +1733,7 @@ func (a *HostAwaitility) CreateAPIProxyConfig(t *testing.T, usertoken, proxyURL 
 func (a *HostAwaitility) CreateAPIProxyClient(t *testing.T, userToken, proxyURL string) (client.Client, error) {
 	proxyKubeConfig := a.CreateAPIProxyConfig(t, userToken, proxyURL)
 
-	s := scheme.Scheme
+	s := runtime.NewScheme()
 	builder := append(runtime.SchemeBuilder{}, corev1.AddToScheme)
 	require.NoError(t, builder.AddToScheme(s))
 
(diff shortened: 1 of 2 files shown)


testsupport/wait/host.go
```diff
@@ -1734,7 +1733,7 @@ func (a *HostAwaitility) CreateAPIProxyConfig(t *testing.T, usertoken, proxyURL
 func (a *HostAwaitility) CreateAPIProxyClient(t *testing.T, userToken, proxyURL string) (client.Client, error) {
 	proxyKubeConfig := a.CreateAPIProxyConfig(t, userToken, proxyURL)
 
-	s := scheme.Scheme
+	s := runtime.NewScheme()
```
mfrancisc, line 1736: is this actually different than using the variable ? I see it does the same thing, but maybe I've missed something:
```
var Scheme = runtime.NewScheme()
```

MatousJobanek, line 1736: let's make sure that there are no other cases where we could face the same issue

mfrancisc: oh right, this is defined in the scheme.go , never mind , I got confused 🤦‍♂️ .

MatousJobanek: just to be sure, I would change this one too: https://github.com/codeready-toolchain/toolchain-e2e/blob/78c7ae131fc87f5f35cceb124c9a6afad81c7390/testsupport/init.go#L262

fbm3307: this is creating a new variable, when we use the variable `var scheme`, thats being used at other places, its use of same variable, hence i changed here

fbm3307: there are 2-3 other places, which i intend to change but, i think it might  require a bit of discussion, hence if this fix is working, we can merge this and fix our broken ci.. and after that i can open a Pr with the others.

mfrancisc: yeah , right! I've noticed that only after commenting, thanks for looking into it!

fbm3307: updated it

mfrancisc (approved):
Great catch 🚀 . Thanks for fixing it 🙏

rajivnathan (approved):

alexeykazakov (approved):

PR #1119, opened 2025-02-14
avoid timeout panic when a parallel test fails

MatousJobanek:
When there is a failure in the parallel test suite, it may happen that the whole e2e test fails with timeout panic:
https://storage.googleapis.com/test-platform-results/pr-logs/pull/openshift_release/61582/rehearse-61582-pull-ci-codeready-toolchain-host-operator-master-e2e/1889811595995910144/build-log.txt
https://storage.googleapis.com/test-platform-results/pr-logs/pull/codeready-toolchain_toolchain-e2e/1111/pull-ci-codeready-toolchain-toolchain-e2e-master-e2e/1889337786046091264/build-log.txt

The reason is that:
* the watcher logic is executed as a sub-test inside of the goroutine
* the WaitGroup is incremented in the parent test (outside of the go routine) 

When one parallel test fails, then the test executor won't start any new sub-test. This means that if the failure happens before starting the sub-test that then the WaitGroup is incremented, but never incremented as the sub-test won't start.

### DIFF
--- a/test/e2e/parallel/proxy_test.go
+++ b/test/e2e/parallel/proxy_test.go
@@ -759,24 +759,22 @@ func runWatcher(t *testing.T, awaitilities wait.Awaitilities) *sync.WaitGroup {
 	waitForWatcher.Add(1)
 	// run the watch in a goroutine because it will take 40 seconds until the call is terminated
 	go func() {
-		t.Run("running the watcher", func(t *testing.T) {
-			defer waitForWatcher.Done()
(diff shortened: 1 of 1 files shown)


rsoaresd (approved):
Thanks!! 🚀

alexeykazakov (approved):

rajivnathan (approved):
Great catch!

fbm3307 (approved):

PR #1123, opened 2025-02-18
wait until ToolchainStatus is updated and in sync with counter

MatousJobanek:
This could minimize the test flakiness. It waits at the beginning of each test suite (and at the end of the migration test suites) until the ToolchainStatus is ready and updated, this means that the client-cache is initialized, the controller has started, and ToolchainStatus in sync with the counter.


mfrancisc (approved):
Makes sense 👍

MatousJobanek:
/retest
missing image

alexeykazakov (approved):

PR #1127, opened 2025-02-27
verify skipped phone-verified accounts in tracking provisioned time

MatousJobanek:
paired with https://github.com/codeready-toolchain/host-operator/pull/1150

[SANDBOX-957](https://issues.redhat.com/browse/SANDBOX-957)

### DIFF
--- a/test/metrics/metrics_test.go
+++ b/test/metrics/metrics_test.go
@@ -314,8 +314,15 @@ func TestVerificationRequiredMetric(t *testing.T) {
 			NewHTTPRequest(t).InvokeEndpoint("GET", route+"/api/v1/signup/verification/invalid", token0, "", http.StatusForbidden)
 			// verify with the correct code
 			NewHTTPRequest(t).InvokeEndpoint("GET", route+fmt.Sprintf("/api/v1/signup/verification/%s", verificationCode), token0, "", http.StatusOK)
-			hostAwait.WaitForMetricDelta(t, wait.UserSignupVerificationRequiredMetric, 1)  // no change after verification initiated
-			hostAwait.WaitForHistogramInfBucketDelta(t, wait.SignupProvisionTimeMetric, 0) // no tracking of provision time for users with phone verification step
+			hostAwait.WaitForMetricDelta(t, wait.UserSignupVerificationRequiredMetric, 1)                                         // no change after verification initiated
+			hostAwait.WaitForMetricDelta(t, wait.UserSignupsMetric, 1)                                                            // user provisioned
+			hostAwait.WaitForMetricDelta(t, wait.UsersPerActivationsAndDomainMetric, 0, "activations", "1", "domain", "internal") // never incremented
+			hostAwait.WaitForMetricDelta(t, wait.UsersPerActivationsAndDomainMetric, 1, "activations", "1", "domain", "external") // user activated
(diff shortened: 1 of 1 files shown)


test/metrics/metrics_test.go
```diff
@@ -312,8 +312,17 @@ func TestVerificationRequiredMetric(t *testing.T) {
...
+			hostAwait.WaitForMetricDelta(t, wait.UserSignupVerificationRequiredMetric, 1)                                         // no change after verification initiated
+			hostAwait.WaitForMetricDelta(t, wait.UserSignupsMetric, 1)                                                            // user provisioned
+			hostAwait.WaitForMetricDelta(t, wait.UsersPerActivationsAndDomainMetric, 0, "activations", "1", "domain", "internal") // never incremented
+			hostAwait.WaitForMetricDelta(t, wait.UsersPerActivationsAndDomainMetric, 1, "activations", "1", "domain", "external") // user activated
+			hostAwait.WaitForMetricDelta(t, wait.UserSignupsApprovedMetric, 1)                                                    // user approved
+			hostAwait.WaitForMetricDelta(t, wait.UserSignupsApprovedWithMethodMetric, 1, "method", "automatic")                   // user automatically approved
+			hostAwait.WaitForMetricDelta(t, wait.UserSignupsApprovedWithMethodMetric, 0, "method", "manual")                      // not manually approved
+			hostAwait.WaitForMetricDelta(t, wait.UserSignupsDeactivatedMetric, 0)                                                 // not deactivated
```
MatousJobanek, line 324: not related to the fix per se, but I thought that it's worth checking that the user was really automatically approved

alexeykazakov (approved):

rajivnathan (approved):

test/metrics/metrics_test.go
```diff
@@ -259,7 +259,7 @@ func TestVerificationRequiredMetric(t *testing.T) {
 	memberAwait := awaitilities.Member1()
 	memberAwait2 := awaitilities.Member2()
 	route := hostAwait.RegistrationServiceURL
-	hostAwait.UpdateToolchainConfig(t, testconfig.AutomaticApproval().Enabled(false)) // disable automatic approval so that users are created with verification required
+	hostAwait.UpdateToolchainConfig(t, testconfig.AutomaticApproval().Enabled(true))
```
fbm3307, line 262: Since we moved to label, thats the reason, checking with automatic approval would work?

fbm3307 (approved):
Looks good, just a question for my understanding

test/metrics/metrics_test.go
MatousJobanek: I'm sorry, but I don't follow. What exactly are you asking about? Why changing the approval strategy, or something else?

fbm3307: Yes why we switched from automatic approval false to true

MatousJobanek: The reason is that we don't track the provision time of manually approved UserSignups. If it stayed `false` then provision time wouldn't be tracked because of this, but this is already verified in different test case. And we would need to also approve the user manually when the phone number is verified. 
Changing it to `true` let me test that the provision time won't be tracked for this user only because it went through the phone verification step.

fbm3307: Thanks for the explanation.

PR #1128, opened 2025-03-05
fix: flaky nstemplateset spacerole check

mfrancisc:
My understanding of the check is that it looks if actual spaceroles are a subset of the expected roles , which makes sense.

It looks like sometimes it fails, see: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/codeready-toolchain_host-operator/1149/pull-ci-codeready-toolchain-host-operator-master-e2e/1896912619076325376/build-log.txt

I wasn't able to fully understand why it works , since apparently the `return false` would exit the function. Basically it can only check the first element match. But I'm sure I'm not understanding this correctly.

I've refactored the logic to deal with subset ( the actual spaceroles ) and supersets ( the expected spaceroles ), and replaced the usage of `continue spaceroles` statement with a simple Hash Table.

### DIFF
--- a/testsupport/tiers/nstemplateset.go
+++ b/testsupport/tiers/nstemplateset.go
@@ -133,21 +133,24 @@ func UntilNSTemplateSetHasTemplateRefs(expectedRevisions TemplateRefs) wait.NSTe
 			if !reflect.DeepEqual(actualNamespaceTmplRefs, expectedRevisions.Namespaces) {
 				return false
 			}
-			// checks that the actual SpaceRole templates match (ie, they are present in `expectedRevisions.SpaceRoles`)
-		spaceroles:
-			for _, r := range actual.Spec.SpaceRoles {
-				// look-up the templateRef
-				for _, ref := range expectedRevisions.SpaceRoles {
-					if r.TemplateRef == ref {
-						continue spaceroles
-					}
-					return false
-				}
-			}
-			return true
+			// checks that the actual spaceroles are at least subSet of the expected spaceroles ones
+			return IsSpaceRoleSubset(expectedRevisions.SpaceRoles, actual.Spec.SpaceRoles)
 		},
 		Diff: func(actual *toolchainv1alpha1.NSTemplateSet) string {
 			return fmt.Sprintf("expected NSTemplateSet '%s' to match the following cluster/namespace/spacerole revisions: %s\nbut it contained: %s", actual.Name, spew.Sdump(expectedRevisions), spew.Sdump(actual.Spec))
 		},
 	}
+}
+
+func IsSpaceRoleSubset(superset map[string]string, subset []toolchainv1alpha1.NSTemplateSetSpaceRole) bool {
+	checkset := make(map[string]bool)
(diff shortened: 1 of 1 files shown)


testsupport/tiers/nstemplateset.go
```diff
@@ -133,21 +133,24 @@ func UntilNSTemplateSetHasTemplateRefs(expectedRevisions TemplateRefs) wait.NSTe
...
 			}
-			// checks that the actual SpaceRole templates match (ie, they are present in `expectedRevisions.SpaceRoles`)
-		spaceroles:
-			for _, r := range actual.Spec.SpaceRoles {
-				// look-up the templateRef
-				for _, ref := range expectedRevisions.SpaceRoles {
-					if r.TemplateRef == ref {
-						continue spaceroles
-					}
-					return false
```
metlos, line 144: I think it should be enough to move this return statement just a single line down, i.e. out of the for loop. It would return false as soon as we find a role from the "actual" set that is not in the "expected" set.

mfrancisc: TBH I don't understand why it works ( kinda works ) right now. It should always fail. 

We can move it down, but I find it easier to read if we remove the continue statement, also we don't have to go over the expected ones for every iteration, makes it a bit faster ( using more memory maybe ).

metlos: This is test code, so I don't think we need to concentrate too much on performance. I personally am fine with both your approach and the nested loops.

That said, on smallish sets, nested loops tend to be generally faster than constructing and modifying maps because, even if they may do more iterations, they don't do any memory allocation.

But as I said, I think your approach is fine here if it improves readability for you.

MatousJobanek (approved):

testsupport/tiers/nstemplateset.go
mfrancisc: But again I'm not able to figure out why it works with the current code. It should always fail from what I'm seeing 😕

metlos (approved):

testsupport/tiers/nstemplateset.go
MatousJobanek: the reason why it worked was that the `expectedRevisions.SpaceRoles` was empty, so it never entered the `for` cycle:
```
for _, ref := range expectedRevisions.SpaceRoles {
```
which means it never got to the `return false` line which is inside the for cycle.

mfrancisc: It did fail sometimes like here: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/codeready-toolchain_host-operator/1149/pull-ci-codeready-toolchain-host-operator-master-e2e/1896912619076325376/build-log.txt, 

not sure if it was because the spaceroles. 

Anyways , now it makes more sense for me at least.

fbm3307:
Thanks for fixing this

PR #1130, opened 2025-03-12
Fix the invocation of the `wait-until-is-installed.sh`

metlos:
`setup-dev-sso.sh` cd's into its directory, so the invocation of the sibling script must not contain the path.

This fixes the `setup-dev-sso` make target.


mfrancisc (approved):
Thanks 👍

fbm3307 (approved):
Thanks

scripts/ci/setup-dev-sso.sh
```diff
@@ -61,7 +61,7 @@ check_command()
 
 run_wait_until_is_installed() {
     PARAMS="-crd keycloak.org -cs '' -n ${DEV_SSO_NS} -s ${SUBSCRIPTION_NAME}"
-    scripts/ci/wait-until-is-installed.sh ${PARAMS}
+    ./wait-until-is-installed.sh ${PARAMS}
```
MatousJobanek, line 64: wouldn't be this safer? 
```suggestion

SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )

run_wait_until_is_installed() {
    PARAMS="-crd keycloak.org -cs '' -n ${DEV_SSO_NS} -s ${SUBSCRIPTION_NAME}"
    ${SCRIPT_DIR}/wait-until-is-installed.sh ${PARAMS}
```

metlos: The script `cd`'s into the script directory - https://github.com/codeready-toolchain/toolchain-e2e/blob/master/scripts/ci/setup-dev-sso.sh#L147. If we were to introduce the `SCRIPT_DIR` which I generally think is good idea, the change would be much more wide-spread because in that case, I'd rework the rest of the script to not assume the PWD either.

MatousJobanek (approved):

PR #1137, opened 2025-03-21
feat: update nstemplateset revisions status

mfrancisc:
Add e2e checks for https://github.com/codeready-toolchain/member-operator/pull/648

Jira:  https://issues.redhat.com/browse/KUBESAW-142


metlos (approved):

test/e2e/parallel/nstemplatetier_test.go
```diff
@@ -231,7 +231,7 @@ func verifyResourceUpdatesForUserSignups(t *testing.T, hostAwait *wait.HostAwait
 		require.NoError(t, err)
 		require.NotNil(t, userAccount)
 
-		nsTemplateSet, err := memberAwaitility.WaitForNSTmplSet(t, usersignup.Status.CompliantUsername, wait.UntilNSTemplateSetHasTier(tier.Name))
+		nsTemplateSet, err := memberAwaitility.WaitForNSTmplSet(t, usersignup.Status.CompliantUsername, wait.UntilNSTemplateSetHasTier(tier.Name), tiers.UntilNSTemplateSetHasStatusTemplateRefs(tiers.GetTemplateRefs(t, hostAwait, tier.Name)))
```
MatousJobanek, line 234: correct me if I'm wrong, this criteria could be applied to all NSTemplateSet CRs, is that correct?
If yes, then why not moving it either here:
https://github.com/codeready-toolchain/toolchain-e2e/blob/63ca31925d122fddf659817e147bbc74baaaa370/testsupport/user_assertions.go#L263-L266
or here:
https://github.com/codeready-toolchain/toolchain-e2e/blob/63ca31925d122fddf659817e147bbc74baaaa370/testsupport/tiers/nstemplateset.go#L22-L24

so it can be verified for all provisioned NSTemplateSets

testsupport/tiers/nstemplateset.go
```diff
@@ -142,6 +143,42 @@ func UntilNSTemplateSetHasTemplateRefs(expectedRevisions TemplateRefs) wait.NSTe
 	}
 }
 
+// UntilNSTemplateSetHasStatusTemplateRefs checks if the NSTemplateTier has the expected template refs in the Status
+func UntilNSTemplateSetHasStatusTemplateRefs(expectedRevisions TemplateRefs) wait.NSTemplateSetWaitCriterion {
+	return wait.NSTemplateSetWaitCriterion{
+		Match: func(actual *toolchainv1alpha1.NSTemplateSet) bool {
+			if expectedRevisions.ClusterResources == nil ||
+				actual.Status.ClusterResources == nil ||
+				*expectedRevisions.ClusterResources != actual.Status.ClusterResources.TemplateRef {
```
MatousJobanek, line 152: shouldn't this rather be:
```suggestion
			if (expectedRevisions.ClusterResources == nil && actual.Status.ClusterResources != nil) ||
				(expectedRevisions.ClusterResources != nil && actual.Status.ClusterResources == nil) ||
				*expectedRevisions.ClusterResources != actual.Status.ClusterResources.TemplateRef {
```
or is there some small detail that I'm missing?

MatousJobanek: btw, couldn't we simply verify that the status equals the spec? checking that the spec matches the expected revisions is a separate verification.

testsupport/tiers/nstemplateset.go
```diff
@@ -142,6 +143,42 @@ func UntilNSTemplateSetHasTemplateRefs(expectedRevisions TemplateRefs) wait.NSTe
...
+			// check expected feature toggles, if any
+			featureAnnotation, featureAnnotationFound := actual.Annotations[toolchainv1alpha1.FeatureToggleNameAnnotationKey]
+			if featureAnnotationFound {
+				if !reflect.DeepEqual(utils.SplitCommaSeparatedList(featureAnnotation), actual.Status.FeatureToggles) {
+					return false
+				}
+			}
+
+			// checks that the actual spaceroles are at least subSet of the expected spaceroles ones
+			return IsSpaceRoleSubset(expectedRevisions.SpaceRoles, actual.Status.SpaceRoles)
```
MatousJobanek, line 174: this is actually the place where we should really verify that the applied SpaceRoles (that are set in the status) exactly match the spec, and not the subset of the expected revisions.

test/e2e/parallel/nstemplatetier_test.go
mfrancisc: I did that initially see: https://github.com/codeready-toolchain/toolchain-e2e/pull/1137/commits/9f4e208cfd5d72a15b1076523ed4e853a9fe42e7

but it will fail on the migration tests, since currently the field is not there. 

We can move it there once these PRs are merged maybe ?

testsupport/tiers/nstemplateset.go
mfrancisc: > shouldn't this rather be

I've copied the logic from `UntilNSTemplateSetHasTemplateRefs` without looking very carefully into it 🤦‍♂️ . I've fixed that now in https://github.com/codeready-toolchain/toolchain-e2e/pull/1137/commits/db0fa2bcc320561a41b48289f421609c9d3f96ea

mfrancisc: > btw, couldn't we simply verify that the status equals the spec? checking that the spec matches the expected revisions is a separate verification.

Yeah, initially I wanted to check for the expected templaterefs from the NSTemplateTier in order to decouple the logic from the spec checks, just in case there's an issue with that and we set wrong/unwanted templateref there. But I agree that checking that comparing the spec and status refs should be enough here. Done in https://github.com/codeready-toolchain/toolchain-e2e/commit/db0fa2bcc320561a41b48289f421609c9d3f96ea

testsupport/tiers/nstemplateset.go
mfrancisc: done in https://github.com/codeready-toolchain/toolchain-e2e/commit/db0fa2bcc320561a41b48289f421609c9d3f96ea

test/e2e/parallel/nstemplatetier_test.go
MatousJobanek: probably something to fix in the migration tests. Do you remember what was the failure?

testsupport/tiers/nstemplateset.go
```diff
@@ -142,6 +143,34 @@ func UntilNSTemplateSetHasTemplateRefs(expectedRevisions TemplateRefs) wait.NSTe
...
+			if !reflect.DeepEqual(actual.Status.Namespaces, actual.Spec.Namespaces) {
+				return false
+			}
+			// check expected feature toggles, if any
+			featureAnnotation, featureAnnotationFound := actual.Annotations[toolchainv1alpha1.FeatureToggleNameAnnotationKey]
+			if featureAnnotationFound {
+				if !reflect.DeepEqual(utils.SplitCommaSeparatedList(featureAnnotation), actual.Status.FeatureToggles) {
+					return false
+				}
+			}
```
MatousJobanek, line 165: we should also take care of the case when the annotation is missing, correct?
```suggestion
			} else {
				if len(actual.Status.FeatureToggles) != 0 {
					return false
				}
			}
```

MatousJobanek (approved):

test/e2e/parallel/nstemplatetier_test.go
mfrancisc: the call stack is: 
`verifyAppStudioProvisionedSignup` in `verify_migration_test.go` -> `VerifyResourcesProvisionedForSignupWithTiers` -> `VerifySpaceRelatedResources` in `user_assertions.go` -> `VerifyNSTemplateSet`

also `VerifyResourcesProvisionedForSignup` goes over the same chain of calls and it is used in multiple parts of the `verify_migration_test.go` if I'm seeing this correctly.

testsupport/tiers/nstemplateset.go
mfrancisc: If I'm not mistaken if the annotation is missing than we don't want to check the `actual.Status.FeatureToggles` field. Basically my understanding is that the annotation is optional we should not return false if it's not there. Only compare it with the status if it's set. 

Am I misunderstanding this ?

MatousJobanek: you are completely right - the annotation is optional, and the status should reflect the existence/content of the annotation. This means that if the annotation is not set, then the featureToggles in the status should be empty.

test/e2e/parallel/nstemplatetier_test.go
MatousJobanek: hmmm... the fact that it's being called in the `verify_migration_test.go` is correct (I thought that you were talking about the setup part), the existing NSTemplateSets should be all updated with the references in the status. If this didn't happen, then there was a bug.

mfrancisc: > the existing NSTemplateSets should be all updated with the references in the status. If this didn't happen, then there was a bug.

but the failure was when checking the NSTemplates sets deployed using the current `master` version of the code, which is expected to not have the status field.

MatousJobanek: right, the accounts were provisioned with the master version. But the version from your PR was deployed after that (before running `verify_migration_test.go`) and the controller reconciled over all NSTemplateSet CRs present in the cluster, and it should update the status in all of them.

mfrancisc: ok, not sure why I was convinced that we run the `verify_migration_tests.go` also after deploying the master version. Than maybe there was a glitch or there is a problem. 

I can try and move it back. Or we can merge it as is and I can open a PR to move it back and see what happens.

MatousJobanek: Let's move it back in this PR, to verify that the migration really works as expected and all NSTemplateSets will get updated

mfrancisc: ok, let's see. done in https://github.com/codeready-toolchain/toolchain-e2e/pull/1137/commits/88a82517979cda410dc308ef7adf375a1688b1d4

MatousJobanek: it looks like that it worked - e2e tests are green :slightly_smiling_face:

mfrancisc: Yeah not sure, maybe it was a PEBKAC 🤦‍♂️ 

Thanks for the suggestion. I'll run again the e2e tests in host operator just to be sure, and I'll merge it.

mfrancisc:
/retest 

updated member pr

mfrancisc:
/retest

updated member pr

fbm3307 (approved):

PR #1154, opened 2025-06-04
test for the bundled ns template tiers being annotated as such

metlos:
This tests the new functionality of automagically annotating the bundled `NSTemplateTier`s as such and makes sure that all the expected `NSTemplateTier`s have the expected annotation.

Related PRs:
- host-operator: https://github.com/codeready-toolchain/host-operator/pull/1180
- host-operator: https://github.com/codeready-toolchain/host-operator/pull/1186
- api: https://github.com/codeready-toolchain/api/pull/480

### DIFF
--- a/test/migration/verify/verify_migration_test.go
+++ b/test/migration/verify/verify_migration_test.go
@@ -18,6 +18,7 @@ import (
 	"github.com/codeready-toolchain/toolchain-e2e/testsupport/wait"
 	appsv1 "k8s.io/api/apps/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
 
 	"github.com/stretchr/testify/assert"
@@ -97,6 +98,7 @@ func runVerifyFunctions(t *testing.T, awaitilities wait.Awaitilities) {
 		func() { verifyBannedSignup(t, awaitilities, bannedSignup) },
 		func() { verifyAdditionalDeploymentsCreatedUsingSSA(t, &awaitilities) },
 		func() { verifyNSTemplateTiers(t, &awaitilities) },
+		func() { verifyResourcesDeployedUsingSSA(t, &awaitilities) },
 	}
 
 	// when & then - run all functions in parallel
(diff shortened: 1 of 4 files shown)


test/migration/verify/verify_migration_test.go
```diff
@@ -270,6 +273,27 @@ func verifyAdditionalDeploymentsCreatedUsingSSA(t *testing.T, awaitilities *wait
...
+	require.NoError(t, awaitilities.Host().Client.List(context.TODO(), list, client.InNamespace(awaitilities.Host().Namespace)))
+
+	// e2e tests have custom appstudio tiers
+	assert.GreaterOrEqual(t, len(list.Items), len(wait.BundledNSTemplateTiers))
+
+	unmatchedBundledTiers := make([]string, len(wait.BundledNSTemplateTiers))
+	copy(unmatchedBundledTiers, wait.BundledNSTemplateTiers)
+
+	for _, tier := range list.Items {
+		if tier.Annotations[toolchainv1alpha1.BundledAnnotationKey] == "host-operator" {
```
alexeykazakov, line 287: We could also check that the custom tiers do not have this annotation.
You could go over all the tiers and do
```
if tier.Annotations[toolchainv1alpha1.BundledAnnotationKey] == "host-operator" {
    assert.Contains(t, wait.BundledNSTemplateTiers, tier.Name)
} else {
    assert.Contains(t, wait.CustomE2eNSTemplateTiers, tier.Name)
}
```

alexeykazakov: You added the check for the custom tiers now. Thanks. But you still could simplify the code by using the approach I mentioned above, right?

```
var (
	BundledNSTemplateTiers   []string = []string{"advanced", "baseextendedidling", "baselarge", "test", "base1ns", "base1nsnoidling", "base1ns6didler", "intelmedium", "intellarge", "intelxlarge", "base"}
	CustomE2eNSTemplateTiers []string = append("appstudio", "appstudiolarge", "appstudio-env")
	E2eNSTemplateTiers       []string = append(BundledNSTemplateTiers, CustomE2eNSTemplateTiers...)
)
```
And then just:
```
list := &toolchainv1alpha1.NSTemplateTierList{}
require.NoError(t, awaitilities.Host().Client.List(context.TODO(), list, client.InNamespace(awaitilities.Host().Namespace)))
assert.Len(t, list.Items, len(wait.E2eNSTemplateTiers))

for _, tier := range list.Items {
	if tier.Annotations[toolchainv1alpha1.BundledAnnotationKey] == "host-operator" {
	    assert.Contains(t, wait.BundledNSTemplateTiers, tier.Name)
	} else {
	    assert.Contains(t, wait.CustomE2eNSTemplateTiers, tier.Name)
	}
}
```

alexeykazakov (approved):
Looks good. We could simplify the code though. See the comment above.

test/migration/verify/verify_migration_test.go
metlos: I wanted to keep my code even though your version is simpler because my code reports on what expected tiers are not present in the cluster whereas your code reports on unexpected tiers present in the cluster.

I wanted to take another look at this to see if I can simplify my code and still report on the missing expected tiers but I forgot :)

But I think I now found the best way of doing this taking inspiration from your suggestions...

alexeykazakov: Hm... My code would check all the tiers. Expected or otherwise. Wouldn't it? It checks that the overall tier number to match the expected tier number. Then it goes over all the tiers and check their annotations. If the expected tier is not present the test would fail too.

PR #1161, opened 2025-06-17
clean up the spaces after the test so that the ns template tier is deletable

metlos:
The new finalizer on the NSTemplateTier prevents it from being deleted as long as there are spaces that reference it.

Some tests didn't clean up the spaces properly leaving the NSTemplateTier not deleteable.


test/e2e/parallel/nstemplatetier_test.go
```diff
@@ -184,6 +184,12 @@ func setupSpaces(t *testing.T, awaitilities wait.Awaitilities, tier *tiers.Custo
 		name := fmt.Sprintf(nameFmt, i)
 		s, _, _ := CreateSpace(t, awaitilities, testspace.WithName(name), testspace.WithTierNameAndHashLabelFor(tier.NSTemplateTier), testspace.WithSpecTargetCluster(targetCluster.ClusterName))
 		spaces = append(spaces, s.Name)
+		t.Cleanup(func() {
+			cl := awaitilities.Host().Client
+			fresh := &toolchainv1alpha1.Space{}
+			require.NoError(t, cl.Get(context.TODO(), client.ObjectKeyFromObject(s), fresh))
+			require.NoError(t, cl.Delete(context.TODO(), fresh))
+		})
```
MatousJobanek, line 192: that's weird - I went thought the functions that are being called and the Spaces are already added to the cleanup tasks: https://github.com/codeready-toolchain/toolchain-e2e/blob/eecfa1c48450277f333bff37b474cd7b5ac91cd3/testsupport/wait/host.go#L2632-L2633 So they should be deleted

testsupport/tiers/tier_setup.go
```diff
@@ -135,7 +136,8 @@ func CreateCustomNSTemplateTier(t *testing.T, hostAwait *wait.HostAwaitility, na
 		err := modify(hostAwait, tier)
 		require.NoError(t, err)
 	}
-	err := hostAwait.CreateWithCleanup(t, tier.NSTemplateTier)
+	// NSTemplateTier can take a long time to delete because they wait for all their spaces to be deleted first...
+	err := hostAwait.CreateWithCleanupTimeout(t, tier.NSTemplateTier, 10*time.Minute)
```
MatousJobanek, line 140: 10 minutes is too much, let's use the double of the default cleanup timeout

PR #1169, opened 2025-07-02
add tests to ensure that the bundled resources are deployed using SSA

metlos:
This is support of the following PRs:
- toolchain-common: https://github.com/codeready-toolchain/toolchain-common/pull/486
- host-operator: https://github.com/codeready-toolchain/host-operator/pull/1185
- ksctl: https://github.com/kubesaw/ksctl/pull/119

### DIFF
--- a/make/clean.mk
+++ b/make/clean.mk
@@ -16,6 +16,9 @@ clean-users:
 	$(Q)-oc delete spaces --all --all-namespaces
 	$(Q)-oc wait --for=delete namespaces -l toolchain.dev.openshift.com/type
 
+clean-nstemplatetiers:
+	$(Q)-oc delete nstemplatetier --all --all-namespaces
+
 .PHONY: clean-cluster-wide-config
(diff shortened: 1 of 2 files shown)


make/clean.mk
```diff
@@ -16,6 +16,9 @@ clean-users:
 	$(Q)-oc delete spaces --all --all-namespaces
 	$(Q)-oc wait --for=delete namespaces -l toolchain.dev.openshift.com/type
 
+clean-nstemplatetiers:
+	$(Q)-oc wait --for=delete nstemplatetier --all --all-namespaces
```
MatousJobanek, line 20: Thanks for adding the wait, but you still need to call the `delete` right? it won't get deleted automatically

metlos: ah, I misread the docs for the wait - I thought it would run the operation, too! 🤦🏼

MatousJobanek (approved):

PR #1171, opened 2025-07-08
if the nstemplatetier deletion is lagging, force the reconciliation by an annotation update

metlos:
This tries to speed up the deletion of the {{NSTemplateTier}}s by changing a value of a "random" annotation on a tier that is being cleaned up in the e2e tests.

This should avoid the exponential back-off interval between finalization attempts as performed by the controller runtime.

Related PRs:
- host-operator: https://github.com/codeready-toolchain/host-operator/pull/1187

https://issues.redhat.com/browse/SANDBOX-1346


fbm3307:
i see the e2e failed with the cleanup wasn't finished in time..  does it make sense to have this for logic for e2e too? or i am misunderstanding the error

MatousJobanek (approved):
Thanks

PR #1172, opened 2025-07-10
pair repo only if ci pairing enable

rsoaresd:
# Description
We need to disable CI pairing in the periodic job

### DIFF
--- a/scripts/ci/manage-operator.sh
+++ b/scripts/ci/manage-operator.sh
@@ -16,7 +16,9 @@ get_repo() {
         # clone
         git clone https://github.com/codeready-toolchain/${REPOSITORY_NAME}.git ${REPOSITORY_PATH}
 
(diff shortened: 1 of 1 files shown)


metlos (approved):

mfrancisc (approved):

alexeykazakov (approved):
I guess we could also check if it's running from the master branch and if so then disable the pairing, right? Or I'm missing something?

rsoaresd:
> I guess we could also check if it's running from the master branch and if so then disable the pairing, right? Or I'm missing something?

yes, thanks, i will address in other pr

rsoaresd:
afterall, we do not need this change, i will revert it. we just need to pass `DEPLOY_LATEST=true` on the job config side: https://github.com/openshift/release/pull/66971

PR #1162, opened 2025-06-19
KUBESAW-143: E2E tests for Adding support for Go Templates in member operator

fbm3307:
This PR is related to Member-Operator : https://github.com/codeready-toolchain/member-operator/pull/680

The Member-Operator PR adds the logic of processing the TTR if found and return the runtimeclient object to be used. If not found then it process the current openshift template and returns the object

This PR adds the GoTemplate Tiers for e2e so that the above logic can be tested via the e2e test

Assisted By: Cursor

PR #1175, opened 2025-07-23
KUBESAW-143: E2E tests for Adding support for Go Templates in member operator

fbm3307:
This PR is related to Member-Operator : https://github.com/codeready-toolchain/member-operator/pull/680

The Member-Operator PR adds the logic of processing the TTR if found and return the runtimeclient object to be used. If not found then it process the current openshift template and returns the object

This PR adds the GoTemplate Tiers for e2e so that the above logic can be tested via the e2e test

Assisted By: Cursor

NOTE: Since i messed up the rebase of https://github.com/codeready-toolchain/toolchain-e2e/pull/1162 , and couldn't revive it hence opened this PR

## Summary by CodeRabbit

* **New Features**
  * Added the "base1ns-gotemplate" tier with Go-template cluster, namespace and admin role templates — parameterized cluster/namespace quotas, RBAC bindings, limit ranges, network policies and an idler.

* **Tests**
  * Added e2e coverage for provisioning with the Go-template tier; test suite and helpers now resolve template revision keys and exclude go-template tiers from certain checks.

* **Chores**
  * Host provisioning updated to apply the new go-template assets during setup.

### DIFF
--- a/deploy/base1ns-gotemplate/tier.yaml
+++ b/deploy/base1ns-gotemplate/tier.yaml
@@ -3,7 +3,7 @@ kind: NSTemplateTier
 metadata:
   name: base1ns-gotemplate
   labels:
-    go-template: toolchain-e2e
+    producer: toolchain-e2e
 spec:
   clusterResources:
     templateRef: cluster-resources-go
--- a/test/e2e/parallel/nstemplatetier_test.go
+++ b/test/e2e/parallel/nstemplatetier_test.go
@@ -56,10 +56,8 @@ func TestNSTemplateTiers(t *testing.T) {
 	allTiers := &toolchainv1alpha1.NSTemplateTierList{}
 	e2eProducer, err := labels.NewRequirement("producer", selection.NotEquals, []string{"toolchain-e2e"})
 	require.NoError(t, err)
-	goTemplateRequirement, err := labels.NewRequirement("go-template", selection.NotEquals, []string{"toolchain-e2e"})
-	require.NoError(t, err)
 	notCreatedByE2e := client.MatchingLabelsSelector{
-		Selector: labels.NewSelector().Add(*e2eProducer, *goTemplateRequirement),
+		Selector: labels.NewSelector().Add(*e2eProducer),
 	}
 	err = hostAwait.Client.List(context.TODO(), allTiers, client.InNamespace(hostAwait.Namespace), notCreatedByE2e)
 	require.NoError(t, err)
--- a/testsupport/wait/host.go
+++ b/testsupport/wait/host.go
@@ -1166,24 +1166,38 @@ func UntilNSTemplateTierSpec(matcher NSTemplateTierSpecMatcher) NSTemplateTierWa
 	}
 }
 
+// extractTemplateRefs extracts all template references from an NSTemplateTier spec
+// and returns them as a map for efficient lookups
+func extractTemplateRefs(spec toolchainv1alpha1.NSTemplateTierSpec) map[string]struct{} {
+	expected := make(map[string]struct{}, len(spec.Namespaces)+1+len(spec.SpaceRoles))
+
+	// Extract template refs from namespaces
+	for _, ns := range spec.Namespaces {
+		if ns.TemplateRef != "" {
+			expected[ns.TemplateRef] = struct{}{}
+		}
+	}
+
+	// Extract template ref from cluster resources
+	if cr := spec.ClusterResources; cr != nil && cr.TemplateRef != "" {
+		expected[cr.TemplateRef] = struct{}{}
+	}
+
+	// Extract template refs from space roles
+	for _, r := range spec.SpaceRoles {
+		if r.TemplateRef != "" {
+			expected[r.TemplateRef] = struct{}{}
+		}
+	}
+
+	return expected
+}
+
 func HasStatusTierTemplateRevisionKeys() NSTemplateTierWaitCriterion {
 	return NSTemplateTierWaitCriterion{
 		Match: func(actual *toolchainv1alpha1.NSTemplateTier) bool {
 			// Build a set of expected revision keys from the tier spec
-			expected := make(map[string]struct{}, len(actual.Spec.Namespaces)+1+len(actual.Spec.SpaceRoles))
-			for _, ns := range actual.Spec.Namespaces {
-				if ns.TemplateRef != "" {
-					expected[ns.TemplateRef] = struct{}{}
-				}
-			}
-			if cr := actual.Spec.ClusterResources; cr != nil && cr.TemplateRef != "" {
-				expected[cr.TemplateRef] = struct{}{}
-			}
-			for _, r := range actual.Spec.SpaceRoles {
-				if r.TemplateRef != "" {
-					expected[r.TemplateRef] = struct{}{}
-				}
-			}
+			expected := extractTemplateRefs(actual.Spec)
 			// Compare sets: size must match and all expected keys must exist in status
 			if len(actual.Status.Revisions) != len(expected) {
 				return false
@@ -1196,36 +1210,7 @@ func HasStatusTierTemplateRevisionKeys() NSTemplateTierWaitCriterion {
 			return true
 		},
 		Diff: func(actual *toolchainv1alpha1.NSTemplateTier) string {
-			// Recalculate expected keys and compute set diffs for clarity
-			expected := make(map[string]struct{}, len(actual.Spec.Namespaces)+1+len(actual.Spec.SpaceRoles))
-			for _, ns := range actual.Spec.Namespaces {
-				if ns.TemplateRef != "" {
-					expected[ns.TemplateRef] = struct{}{}
-				}
-			}
-			if cr := actual.Spec.ClusterResources; cr != nil && cr.TemplateRef != "" {
-				expected[cr.TemplateRef] = struct{}{}
-			}
-			for _, r := range actual.Spec.SpaceRoles {
-				if r.TemplateRef != "" {
-					expected[r.TemplateRef] = struct{}{}
-				}
-			}
-			var missing, extra, expectedKeys, actualKeys []string
-			for k := range expected {
-				expectedKeys = append(expectedKeys, k)
-				if _, ok := actual.Status.Revisions[k]; !ok {
-					missing = append(missing, k)
-				}
-			}
-			for k := range actual.Status.Revisions {
-				actualKeys = append(actualKeys, k)
-				if _, ok := expected[k]; !ok {
-					extra = append(extra, k)
-				}
-			}
-			return fmt.Sprintf("revision key mismatch: missing=%v extra=%v expectedKeys=%v actualKeys=%v",
-				missing, extra, expectedKeys, actualKeys)
+			return fmt.Sprintf("expected template refs: %v\nactual status revisions: %v", extractTemplateRefs(actual.Spec), actual.Status.Revisions)
 		},
 	}
 }
--- a/test/migration/verify/verify_migration_test.go
+++ b/test/migration/verify/verify_migration_test.go
@@ -286,7 +286,7 @@ func verifyNSTemplateTiers(t *testing.T, awaitilities *wait.Awaitilities) {
 	//
 	// This makes sure that the setup in the cluster is exactly how the e2e tests
 	// expect it to be.
-	goTemplateRequirement, err := labels.NewRequirement("go-template", selection.NotEquals, []string{"toolchain-e2e"})
+	goTemplateRequirement, err := labels.NewRequirement("producer", selection.NotEquals, []string{"toolchain-e2e"})
 	require.NoError(t, err)
 	notCreatedByGoTemplate := client.MatchingLabelsSelector{
 		Selector: labels.NewSelector().Add(*goTemplateRequirement),


fbm3307:
> about [#1162 (comment)](https://github.com/codeready-toolchain/toolchain-e2e/pull/1162#discussion_r2222605432) , 
@mfrancisc 
updated these too as per suggestions PTAL

testsupport/tiers/checks.go
```diff
@@ -30,6 +30,7 @@ const (
 	appstudioEnv    = "appstudio-env"
 	base            = "base"
 	base1ns         = "base1ns"
+	gobase1ns       = "base1ns-gotemplate"
```
mfrancisc, line 33: minor - but just for consistency we could name this `base1nsGOTemplate` . I see all the other variables follow the tier name.

mfrancisc (approved):
Great job 🏅 

mfrancisc:
@coderabbitai review

testsupport/tiers/templaterefs.go
```diff
@@ -15,6 +15,41 @@ type TemplateRefs struct {
...
+
+	if rev, ok := templateTier.Status.Revisions[templateTier.Spec.ClusterResources.TemplateRef]; ok {
+		clusterResources = &rev
+	}
+	return TemplateRefs{
+		Namespaces:       nsRefs,
+		ClusterResources: clusterResources,
+		SpaceRoles:       spaceRoleRefs,
+	}
+}
```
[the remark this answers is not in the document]
fbm3307: updated

deploy/base1ns-gotemplate/spacerole_admin.yaml
```diff
@@ -0,0 +1,81 @@
...
+        - list
+        - watch
+        - create
+        - update
+        - patch
+        - delete
+    parameters:
+    - name: NAMESPACE
+      required: true
+  templateObjects:
```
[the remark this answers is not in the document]
fbm3307: it is a run time parameter so no need to add it here

deploy/base1ns-gotemplate/spacerole_admin.yaml
```diff
@@ -0,0 +1,81 @@
...
+        - rolebindings
+        verbs:
+        - get
+        - list
+        - watch
+        - create
+        - update
+        - patch
+        - delete
+    parameters:
```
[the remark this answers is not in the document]
fbm3307: the oc template sec is kept in go templates as that sections is still a required field in api and not used in processing, will be removed once the field in api is marked optional

deploy/base1ns-gotemplate/ns_dev.yaml
```diff
@@ -0,0 +1,284 @@
...
+      resources:
+        - pods/exec
+      verbs:
+        - get
+        - list
+        - watch
+        - create
+        - delete
+        - update
+  - apiVersion: rbac.authorization.k8s.io/v1
```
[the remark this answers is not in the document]
fbm3307: updated

deploy/base1ns-gotemplate/ns_dev.yaml
```diff
@@ -0,0 +1,284 @@
...
+      name: allow-from-dev-sandbox-managed-ns
+      namespace: '{{.SPACE_NAME}}-dev'
+    spec:
+      ingress:
+        - from:
+            - namespaceSelector:
+                matchLabels:
+                  dev-sandbox/policy-group: ingress
+      policyTypes:
+        - Ingress
```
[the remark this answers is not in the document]
fbm3307: it is fine this is just for testing e2e tests

deploy/base1ns-gotemplate/ns_dev.yaml
```diff
@@ -0,0 +1,284 @@
...
+      namespace: '{{.SPACE_NAME}}-dev'
+    roleRef:
+      apiGroup: rbac.authorization.k8s.io
+      kind: ClusterRole
+      name: view
+    subjects:
+      - apiGroup: rbac.authorization.k8s.io
+        kind: Group
+        name: crtadmin-users-view
+  - apiVersion: rbac.authorization.k8s.io/v1
```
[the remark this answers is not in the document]
fbm3307: it should be okay

test/e2e/parallel/nstemplatetier_test.go
```diff
@@ -143,6 +143,35 @@ func TestUpdateNSTemplateTier(t *testing.T) {
...
+	user := NewSignupRequest(awaitilities).
+		Username("gotemplateuser").
+		ManuallyApprove().
+		TargetCluster(awaitilities.Member1()).
+		SpaceTier("base1ns-gotemplate").
+		EnsureMUR().
+		RequireConditions(wait.ConditionSet(wait.Default(), wait.ApprovedByAdmin())...).
+		Execute(t)
+	testingtiers := user.UserSignup
+	VerifyResourcesProvisionedForSignupWithTiers(t, awaitilities, testingtiers, "deactivate30", "base1ns-gotemplate")
```
MatousJobanek, line 171: during our calls I told you to reuse the `CustomNSTemplateTier` abstraction, could you please clarify why you didn't use that?
```suggestion
	base1Ns, err := hostAwait.WaitForNSTemplateTier(t, "base1ns")
	require.NoError(t, err)
	VerifyResourcesProvisionedForSpaceWithCustomTier(t, hostAwait, awaitilities.Member1(), user.Space.Name, &tiers.CustomNSTemplateTier{
		NSTemplateTier:         tier,
		ClusterResourcesTier:   base1Ns,
		NamespaceResourcesTier: base1Ns,
		SpaceRolesTier:         base1Ns,
	})
```

testsupport/tiers/templaterefs.go
```diff
@@ -15,6 +15,39 @@ type TemplateRefs struct {
 	SpaceRoles       map[string]string
 }
 
+// GetGoTemplateRefs returns the expected templateRefs populated from the NSTemplateTier.Status.Revisions field
+func GetGoTemplateRefs(t *testing.T, hostAwait *wait.HostAwaitility, tierName string) TemplateRefs { // nolint:unparam // false positive on unused return param `0`??
```
MatousJobanek, line 19: why do we need another function for getting the refs, why not updating the already existing one?

MatousJobanek (changes requested):

testsupport/tiers/templaterefs.go
fbm3307: Do you mean something of this sort? https://github.com/codeready-toolchain/toolchain-e2e/pull/1175/commits/954f5c127e09d503adc635ea4c88d65ba7349f81#diff-db6fcb225c7e6e46d0d48faa354433486832e8e96328ab1ddc21bbeeb5230cbbR26

MatousJobanek: you don't need any if-else for this, do you? You can use the `NSTemplateTier.Status.Revisions` for both types of ties, right?

MatousJobanek: btw, this is one of the comments that I made during our calls as well - we need to update all the helper functions to make them ready for the upcoming changes. The updated functions should be as generic as possible.

mfrancisc: > you don't need any if-else for this, do you? You can use the NSTemplateTier.Status.Revisions for both types of tiers, right?

It would be great to have one way for fetching the templaterefs for both new tiers and existing one, and that will allow us to remove the new checks as you pointed out in https://github.com/codeready-toolchain/toolchain-e2e/pull/1175/files#r2250840195

I remember we did that initially with @fbm3307 , we updated the existing `GetTemplateRefs` function to fetch the values from the `NStemplateTier.Status.Revision` but that was causing a flaky failure in the `TestTierTemplateRevision` , specifically in the "_update of the NSTemplateTier parameters should trigger creation of new TTR_" , since that test uses the `UpdateCustomNSTemplateTier`for updating the NSTemplateTier with new parameters which triggers creation of new TTRS while the check has already fetched the old ttrs at :
```
newTTier, err := hostAwait.WaitForNSTemplateTier(t, tier.Name,
	wait.HasStatusTierTemplateRevisions(GetTemplateRefs(t, hostAwait, tier.Name).Flatten()))
require.NoError(t, err)
``` 
Unfortunately while TTRS are created by the `nstemplatetier_controller` the Ready condition of the NSTemplateTier remains set to True, so we cannot rely on that right now for figuring out that an update on the tier is going on. 

So, unless I'm missing something, I think we can either:

1. fix the nstemplatier_controller to reflect the Updating phase in the Ready condition, and in the test we can check the ready condition of the NSTemplatier before fetching the refs from the `.Status.Revisions` field ( **recommend** )

2. remove this check from the `UpdateCustomNSTemplateTier` : https://github.com/codeready-toolchain/toolchain-e2e/blob/7ac3539c4d4d38078bdb96ee12e16b5232b2db96/testsupport/tiers/tier_setup.go#L166-L167 which was causing the flakiness ( but IIRC that will introduce other flakiness in some parts )

@fbm3307 @MatousJobanek pls let me know if I'm missing something, or there's some other idea on how to fix this.

MatousJobanek: > It would be great to have one way for fetching the templaterefs for both new tiers and existing one, and that will allow us to remove the new checks as you pointed out in

+1000

> fix the nstemplatier_controller to reflect the Updating phase in the Ready condition, and in the test we can check the ready condition of the NSTemplatier before fetching the refs from the .Status.Revisions field ( recommend )

we can do that, but I'm not sure how this would help. it would be still a problem with proper timing

> remove this check from the UpdateCustomNSTemplateTier : 

well, since the `GetTemplateRefs` will take the revisions from the `NSTemplateTier.Status.Revisions`, then executing the check `HasStatusTierTemplateRevisions` doesn't make much sense anymore because it would compare the same thing.
Looking at the logic of `HasStatusTierTemplateRevisions`, it only checks if the status has the keys from the spec. If this helps and is needed, then it can work in the same way, just the logic will be placed only in the `HasStatusTierTemplateRevisions` function.

The question is if we could face any flakiness because we don't properly verify:
* when the update was successful
* that the creation of the TTR was finished
* the status of the NsTS is updated with the latest TTR refs
Since the logic is not doing the verification now anyway, then I assume that it's OK to skip this verification (at least for now).

mfrancisc: > - update the HasStatusTierTemplateRevisions so it checks that all templates defined in the spec has a corresponding entity in the status. In the same way as it is now, but with the difference that the function takes the values from the spec on its own.
> - update the GetTemplateRefs so it takes the template refs from NSTemplateTier.Status.Revisions

I'm not sure if that is enough, because as mentioned one of the flakiness was caused by the fact the `TestTierTemplateRevision` updates the `NSTemplateSet.Spec.Parameters` field , and not any of the template refs in the spec, and the parameter update triggers ttr creation for all the tiertemplates. Unless I'm missing something this usecase will not be covered by this check.

fbm3307: @MatousJobanek Yes, i understand your point , 
Thanks @mfrancisc you are right about the error

this i already tried twice, [commit -1](https://github.com/codeready-toolchain/toolchain-e2e/pull/1175/commits/0409e515939b8e0b38c5bf2422b87cdf018cf37e#diff-db6fcb225c7e6e46d0d48faa354433486832e8e96328ab1ddc21bbeeb5230cbbR26), [commit -2 ](https://github.com/codeready-toolchain/toolchain-e2e/pull/1175/commits/8629d66c63a2e5cf8578b24b6598e4e16a1d48b1)
it failled itermittently, due to timing issue, that status.revisions was not yet set and the test went ahead.

`

        diffs:
        expected revision keys [] not found in: map[ttrfrombase1ns-admin-88b275d-88b275d:ttr-ttrfrombase1ns-admin-88b275d-88b275d-psgll ttrfrombase1ns-clusterresources-5b0d541-5b0d541:ttr-ttrfrombase1ns-clusterresources-5b0d541-5b0d541-69zm4 ttrfrombase1ns-dev-0bff31a-0bff31a:ttr-ttrfrombase1ns-dev-0bff31a-0bff31a-nml2m]
        
    tier_setup.go:144: 
        	Error Trace:	/go/src/github.com/codeready-toolchain/toolchain-e2e/testsupport/tiers/tier_setup.go:144
        	            				/go/src/github.com/codeready-toolchain/toolchain-e2e/test/e2e/parallel/nstemplatetier_test.go:417
        	      `

[e2e error1](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/codeready-toolchain_toolchain-e2e/1175/pull-ci-codeready-toolchain-toolchain-e2e-master-e2e/1948288344232300544#1:build-log.txt%3A14944)

[e2e error2](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/codeready-toolchain_toolchain-e2e/1175/pull-ci-codeready-toolchain-toolchain-e2e-master-e2e/1949694023530713088#1:build-log.txt%3A15352) 

also about checking the spec

> > * update the HasStatusTierTemplateRevisions so it checks that all templates defined in the spec has a corresponding entity in the status. In the same way as it is now, but with the difference that the function takes the values from the spec on its own.
> >   which is being covered by `HasStatusTierTemplateRevisionKeys`
> 

this was also tried in [commit](https://github.com/codeready-toolchain/toolchain-e2e/pull/1175/commits/f2cd330fcab705df3706e27a4fd966092ea40e8f)

Since these error came with all of the above  approaches and then couldn't figure out any other way to remove the error and the flakiness,
 so created another method for handling go templates

So not sure about any other way to have a common implementation .

test/e2e/parallel/nstemplatetier_test.go
fbm3307: I had tried that too, 
the suggestion which you proposed is already being covered here in test [TestNSTemplateTiers](https://github.com/codeready-toolchain/toolchain-e2e/pull/1175/files#diff-d1f261f3430b9095a3c1639d087ee7fdb0c4a33756c21552ed0a53095c7d9326L36)  , since i had got a review [comment](https://github.com/codeready-toolchain/toolchain-e2e/pull/1175/files#diff-2379a187b4778a2330f68b2b1a930fb26d3b8e1b567efe2966e4cb6d8aafe989L54) to add it to the list in host 

Also, this means the the test `TestGoTemplate` is a duplicate and isn't specifically required, so i am thinking to remove  it , WDYT?

testsupport/tiers/templaterefs.go
MatousJobanek: and as for `GetTemplateRefs` - just simply replace the old content
https://github.com/codeready-toolchain/toolchain-e2e/blob/25d45278b28908bd67a1419a836f426638418be6/testsupport/tiers/templaterefs.go#L53-L68
with the new one
https://github.com/codeready-toolchain/toolchain-e2e/blob/25d45278b28908bd67a1419a836f426638418be6/testsupport/tiers/templaterefs.go#L20-L48

testsupport/tiers/checks.go
```diff
@@ -328,6 +331,74 @@ func commonNetworkPolicyChecks() []namespaceObjectsCheck {
 	}
 }
 
+type base1nsGoTemplateTierChecks struct {
+	tierName string
+}
```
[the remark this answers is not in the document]
MatousJobanek: well, the resource is the same, right? if not then there is a bug. 
Looking at the types you wrote above ` corev1.resourseList` and `map[corev1.ResourceName]resource.Quantity`
should be technically the same:
```
type ResourceList map[ResourceName]resource.Quantity
```
https://github.com/kubernetes/api/blob/d1a00b8cc38fa8aee1a24b03aeb7e7ec55ff10d8/core/v1/types.go#L6884-L6885

Have you tried to change the existing functions to use `ResourceList` instead?

test/e2e/parallel/nstemplatetier_test.go
MatousJobanek: > the suggestion which you proposed is already being covered here in test [TestNSTemplateTiers](https://github.com/codeready-toolchain/toolchain-e2e/pull/1175/files#diff-d1f261f3430b9095a3c1639d087ee7fdb0c4a33756c21552ed0a53095c7d9326L36)

I'm sorry, but the link you provided doesn't point me to any specific line in the code, could you please use permanent links? The suggestion is reusing the already existing checks, and relying on the `CustomNSTemplateTier` abstraction - it's related to my other comment above, that you don't need to implement a new set of checks which are technically a duplication of the already existing ones. 

> since i had got a review [comment](https://github.com/codeready-toolchain/toolchain-e2e/pull/1175/files#diff-2379a187b4778a2330f68b2b1a930fb26d3b8e1b567efe2966e4cb6d8aafe989L54) to add it to the list in host

this doesn't point me to any review comment, but to a line
https://github.com/codeready-toolchain/toolchain-e2e/blob/25d45278b28908bd67a1419a836f426638418be6/testsupport/wait/host.go#L54

> Also, this means the the test TestGoTemplate is a duplicate and isn't specifically required, because everuthing is covered and tested in the TestNSTemplateTiers so i am thinking to remove it , WDYT?

if we stayed with the code as it is now (with the special implementation of checks for `base1ns-gotemplate`), then technically yes, but we don't need the duplication of the checks, do we?

testsupport/tiers/templaterefs.go
MatousJobanek: >  do you mean removing the HasStatusTierTemplateRevisions(revisions []string) and refernces from everywhere. and use HasStatusTierTemplateRevisionKeys instead?

exactly :point_up_2: 

> because that is being used at many many places and many test case so was not sure what it might break. (or what refactoring might be required in whole e2e test)

until you do it you won't be able to figure it out. Given the logic of the functions, I don't think that it would cause any additional refactoring/changes/problems.

Changing the existing code that is obsolete to something more generic that works for all cases is completely OK thing to do. This is also what I meant in our calls by mentioning that we need to update the existing helper functions so they are ready for the upcoming changes. This PR is an excellent opportunity for it.

mfrancisc: Yeah I agree , we can try that. Thanks for the suggestions and help @MatousJobanek . This might help with avoding the flakiness we saw in the test.

testsupport/tiers/checks.go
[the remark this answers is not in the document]
fbm3307: have updated to this suggestion in [168d916](https://github.com/codeready-toolchain/toolchain-e2e/pull/1175/commits/168d916460f4a4f490f930e15d8300e0da8efb07)

will clean up the rest if this works in next commit while i sort out the below comment and suggestion

deploy/base1ns-gotemplate/tier.yaml
```diff
@@ -0,0 +1,45 @@
...
+  - name: CPU_LIMIT
+    value: "30000m"
+  - name: EPHEMERAL_STORAGE_LIMIT
+    value: "15Gi"
+  - name: EPHEMERAL_STORAGE_REQUEST
+    value: "15Gi"
+  - name: STORAGE_REQUEST
+    value: "80Gi"
+  - name: PERSISTENT_VOLUME_CLAIM
+    value: "10"
```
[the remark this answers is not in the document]
fbm3307: yes right updated it

testsupport/tiers/checks.go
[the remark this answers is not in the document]
MatousJobanek: I see that you did something different, have you tried to simply change the original function so it uses `corev1.resourseList` instead of `map[ResourceName]resource.Quantity`? 

BTW, don't forget that you can easily test this locally by deploying the modified member operator and running only this single test on it's own - you don't have to rely on openshift-ci and wait for the results.

test/e2e/parallel/nstemplatetier_test.go
fbm3307: i have already dropped special checks, just what i do is do the base1ns checks for go template https://github.com/codeready-toolchain/toolchain-e2e/pull/1175/files#diff-656de6540cafb22fc3a2da30b608ceff7c22769a72bb58bf9ec90ad4fc110f75R59

because from what i tried on the suggestion above i see if i use `base1ns` custom-tier on space `base1ns-gotemplate`, the tier name won't match 
`
        diffs:
        expected tier name to match:
        
        expected:
        ----
        base1ns----
        
        actual:
        ----
        base1ns-gotemplate----
        -expected
        +actual
          string(
        - 	"base1ns",
        + 	"base1ns-gotemplate",
          )
        
        expected space to contain label toolchain.dev.openshift.com/base1ns-tier-hash:084ab6416818552177c3047bc6112149:
        (map[string]string) (len=3) {
         (string) (len=33) "toolchain.dev.openshift.com/state": (string) (len=16) "cluster-assigned",
         (string) (len=56) "toolchain.dev.openshift.com/base1ns-gotemplate-tier-hash": (string) (len=32) "3f3c7010c63c8794e469cccd9cb1c2fb",
         (string) (len=35) "toolchain.dev.openshift.com/creator": (string) (len=14) "gotemplateuse1"
        }
`

deploy/base1ns-gotemplate/cluster.yaml
```diff
@@ -0,0 +1,166 @@
...
+        toolchain.dev.openshift.com/feature: openshift-lightspeed
+      name: 'osl-access-{{.SPACE_NAME}}'
+    roleRef:
+      apiGroup: rbac.authorization.k8s.io
+      kind: ClusterRole
+      name: lightspeed-operator-query-access
+    subjects:
+      - apiGroup: rbac.authorization.k8s.io
+        kind: User
+        name: '{{.SPACE_NAME}}'
```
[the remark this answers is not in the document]
fbm3307: its not rejected and it works fine

test/e2e/parallel/nstemplatetier_test.go
MatousJobanek: Do you know why?

fbm3307: because we are setting the spacetier in user to `base1ns-gotemplate`
 https://github.com/codeready-toolchain/toolchain-e2e/pull/1175/files#diff-d1f261f3430b9095a3c1639d087ee7fdb0c4a33756c21552ed0a53095c7d9326R166
 
 so the space label will contain `base1ns-gotemplate `and not `base1ns`

MatousJobanek: how is it different compared to the other cases where we use `CustomNSTemplateTier` abstraction?

fbm3307: nstemplateset ref will be different 

`
expected NSTemplateSet 'gotemplateuser' to match the following cluster/namespace/spacerole revisions: (tiers.TemplateRefs) {
         Namespaces: ([]string) (len=1 cap=1) {
          (string) (len=6) "dev-go"
         },
         ClusterResources: (*string)(0xc00020b3b0)((len=20) "cluster-resources-go"),
         SpaceRoles: (map[string]string) (len=1) {
          (string) (len=5) "admin": (string) (len=18) "spacerole-admin-go"
         }
        }
        
        but it contained: (v1alpha1.NSTemplateSetSpec) {
         TierName: (string) (len=18) "base1ns-gotemplate",
         Namespaces: ([]v1alpha1.NSTemplateSetNamespace) (len=1 cap=4) {
          (v1alpha1.NSTemplateSetNamespace) {
           TemplateRef: (string) (len=31) "base1ns-gotemplate-dev-go-lnvgj"
          }
         },
         ClusterResources: (*v1alpha1.NSTemplateSetClusterResources)(0xc00062c1d0)({
          TemplateRef: (string) (len=45) "base1ns-gotemplate-cluster-resources-go-9ll6j"
         }),
         SpaceRoles: ([]v1alpha1.NSTemplateSetSpaceRole) (len=1 cap=4) {
          (v1alpha1.NSTemplateSetSpaceRole) {
           TemplateRef: (string) (len=43) "base1ns-gotemplate-spacerole-admin-go-dbkkp",
           Usernames: ([]string) (len=1 cap=4) {
            (string) (len=14) "gotemplateuser"
           }
          }
         }
        }
`

MatousJobanek: cool, so we moved forward. Now let's try to figure out why it failed.

fbm3307:
/test e2e
Not related & Flaky test failure

test/e2e/parallel/nstemplatetier_test.go
fbm3307: Okay finaly made it work with the above suggestions.

test/e2e/parallel/nstemplatetier_test.go
```diff
@@ -143,6 +143,41 @@ func TestUpdateNSTemplateTier(t *testing.T) {
...
 
+func TestGoTemplate(t *testing.T) {
+	t.Parallel()
+
+	// This test verifies that a go-template NSTemplateTier can be used to provision a user.
+	// It creates a custom NSTemplateTier based on the "base1ns-gotemplate" tier, labels it as produced by "toolchain-e2e",
+	// and provisions a user with this tier. It then verifies that the resources are provisioned as expected.
+
+	// awaitilities: Helper struct that provides Awaitility instances for host and member clusters, used for waiting on resources.
+	// hostAwait: Awaitility instance for the host cluster, used to interact with and wait for resources in the host namespace.
```
MatousJobanek, line 154: why is this here?

test/e2e/parallel/nstemplatetier_test.go
```diff
@@ -143,6 +143,41 @@ func TestUpdateNSTemplateTier(t *testing.T) {
 	verifyResourceUpdatesForSpaces(t, hostAwait, memberAwait, spaces, chocolateTier)
 }
 
+func TestGoTemplate(t *testing.T) {
```
MatousJobanek, line 146: can you add `// given` `// when` `// then` comments?

testsupport/tiers/checks.go
```diff
@@ -25,13 +25,14 @@ import (
...
-	base            = "base"
-	base1ns         = "base1ns"
-	base1ns6didler  = "base1ns6didler"
-	base1nsnoidling = "base1nsnoidling"
+	appstudio         = "appstudio"
+	appstudiolarge    = "appstudiolarge"
+	appstudioEnv      = "appstudio-env"
+	base              = "base"
+	base1ns           = "base1ns"
+	base1nsGoTemplate = "base1ns-gotemplate"
```
MatousJobanek, line 33: can you clean it up?

testsupport/tiers/checks.go
```diff
@@ -102,23 +105,31 @@ func (c *customTierChecks) GetClusterObjectChecks() []clusterObjectsCheck {
...
+		}
+	}
 	return TemplateRefs{
-		ClusterResources: clusterResourcesTmplRef,
-		Namespaces:       namespaceTmplRefs,
-		SpaceRoles:       spaceRolesTmplRefs,
+		Namespaces:       nsRefs,
+		ClusterResources: clusterResources,
+		SpaceRoles:       spaceRoleRefs,
 	}
```
MatousJobanek, line 133: couldn't we use `GetTemplateRefs` function here? The logic seems to be the same.

fbm3307: yes i was thinking the same.. updated it 
e1ec75c

test/e2e/parallel/nstemplatetier_test.go
fbm3307: leftovers, cleaned it 
e1ec75c

testsupport/tiers/checks.go
MatousJobanek: The whole purpose of the `CustomNSTemplateTier` is to reuse the already existing tiers and checks so we don't have to modify the `checks.go` file for every single custom tier that we define/use in any of the tests.
Apart from that, if what you have said was true, then the existing tests for other custom tier would fail, wouldn't they?

testsupport/wait/host.go
```diff
@@ -51,7 +51,7 @@ import (
 
 var (
 	BundledNSTemplateTiers []string = []string{"base1ns", "base1nsnoidling", "base1ns6didler", "base"}
-	CustomNSTemplateTiers  []string = []string{"appstudio", "appstudiolarge", "appstudio-env"}
+	CustomNSTemplateTiers  []string = []string{"appstudio", "appstudiolarge", "appstudio-env", "base1ns-gotemplate"}
```
MatousJobanek, line 54: let's drop this one as well
```suggestion
	CustomNSTemplateTiers  []string = []string{"appstudio", "appstudiolarge", "appstudio-env"}
```

testsupport/tiers/checks.go
fbm3307: this was the reason the other custom tiers test was passing,

testsupport/wait/host.go
```diff
@@ -1166,22 +1166,46 @@ func UntilNSTemplateTierSpec(matcher NSTemplateTierSpecMatcher) NSTemplateTierWa
...
+			if actual.Spec.ClusterResources != nil {
+				expectedRevisions = append(expectedRevisions, actual.Spec.ClusterResources.TemplateRef)
+			}
+			for r := range actual.Spec.SpaceRoles {
+				expectedRevisions = append(expectedRevisions, actual.Spec.SpaceRoles[r].TemplateRef)
+			}
+			return fmt.Sprintf("expected revision keys %v not found in: %v", expectedRevisions, actual.Status.Revisions)
 		},
 	}
 }
```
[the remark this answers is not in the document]
fbm3307: updated now

test/e2e/parallel/nstemplatetier_test.go
```diff
@@ -56,8 +56,10 @@ func TestNSTemplateTiers(t *testing.T) {
 	allTiers := &toolchainv1alpha1.NSTemplateTierList{}
 	e2eProducer, err := labels.NewRequirement("producer", selection.NotEquals, []string{"toolchain-e2e"})
 	require.NoError(t, err)
+	goTemplateRequirement, err := labels.NewRequirement("go-template", selection.NotEquals, []string{"toolchain-e2e"})
+	require.NoError(t, err)
 	notCreatedByE2e := client.MatchingLabelsSelector{
-		Selector: labels.NewSelector().Add(*e2eProducer),
+		Selector: labels.NewSelector().Add(*e2eProducer, *goTemplateRequirement),
```
MatousJobanek, line 62: couldn't you simply add the label
```
producer: toolchain-e2e
```
to the tier?

testsupport/wait/host.go
```diff
@@ -1166,22 +1166,66 @@ func UntilNSTemplateTierSpec(matcher NSTemplateTierSpecMatcher) NSTemplateTierWa
...
+				}
+			}
+			if cr := actual.Spec.ClusterResources; cr != nil && cr.TemplateRef != "" {
+				expected[cr.TemplateRef] = struct{}{}
+			}
+			for _, r := range actual.Spec.SpaceRoles {
+				if r.TemplateRef != "" {
+					expected[r.TemplateRef] = struct{}{}
+				}
+			}
```
MatousJobanek, line 1213: seems to be the same as in the `Match` part - can you move it to a common function?

testsupport/wait/host.go
```diff
@@ -1166,22 +1166,66 @@ func UntilNSTemplateTierSpec(matcher NSTemplateTierSpecMatcher) NSTemplateTierWa
...
+			for k := range actual.Status.Revisions {
+				actualKeys = append(actualKeys, k)
+				if _, ok := expected[k]; !ok {
+					extra = append(extra, k)
+				}
+			}
+			return fmt.Sprintf("revision key mismatch: missing=%v extra=%v expectedKeys=%v actualKeys=%v",
+				missing, extra, expectedKeys, actualKeys)
```
MatousJobanek, line 1228: I guess that this is a bit overcomplicated solution - simply print out the `expected` refs and the actual content of the status.

MatousJobanek (approved):
Looks good, some minor things

PR #1178, opened 2025-07-25
Fix Test case failure of testtiertemplate

fbm3307:
fixing the number of ttr to be checked 
https://redhat-internal.slack.com/archives/CHK0J6HT6/p1753440934539929

### DIFF
--- a/test/e2e/parallel/nstemplatetier_test.go
+++ b/test/e2e/parallel/nstemplatetier_test.go
@@ -267,7 +267,7 @@ func TestTierTemplates(t *testing.T) {
 	require.NoError(t, err)
 	// We have 19 tier templates (base: 3, base1ns: 2, base1nsnoidling: 2, base1ns6didler: 3, appstudio: 3, appstudiolarge: 3, appstudio-env: 3)
 	// But we cannot verify the exact number of tiers, because during the operator update it may happen that more TierTemplates are created
-	assert.GreaterOrEqual(t, len(allTiers.Items), 24)
+	assert.GreaterOrEqual(t, len(allTiers.Items), 19)
 }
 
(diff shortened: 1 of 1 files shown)


test/e2e/parallel/nstemplatetier_test.go
```diff
@@ -267,7 +267,7 @@ func TestTierTemplates(t *testing.T) {
 	require.NoError(t, err)
 	// We have 19 tier templates (base: 3, base1ns: 2, base1nsnoidling: 2, base1ns6didler: 3, appstudio: 3, appstudiolarge: 3, appstudio-env: 3)
 	// But we cannot verify the exact number of tiers, because during the operator update it may happen that more TierTemplates are created
-	assert.GreaterOrEqual(t, len(allTiers.Items), 38)
+	assert.GreaterOrEqual(t, len(allTiers.Items), 24)
```
mfrancisc, line 270: why 24 is more accurate ?

fbm3307: just a calculated guess, i saw the previous was 28/29 and we were checking somewhere around 33 ..

mfrancisc: TBH If we cannot predict how many tier there would be at the time of the check but we know that we have in total 19 tiers, why not using 19 as the number for the check ?

fbm3307: updated now

mfrancisc: mine was more of a question, just wanted to understand the rationale behind this number and the purpose of the check ( what is this trying to ensure ? ) . What makes sense for me is to check for the total number of expected tiers.

PR #1185, opened 2025-08-21
Don't cache the tests that are part of the e2e testsuite.

metlos:
This prevents spurious errors when running the tests repeatedly without any change in toolchain-e2e itself.


alexeykazakov (approved):
Looks good. But could you add a comment to the makefile(s) to explain that using "-count=1" (the default count anyway) is a trick to disable caching. It's not obvious. I had to google it :)

PR #1195, opened 2025-09-08
drop 'Install Go' step from govulncheck workflow

MatousJobanek:
drop 'Install Go' step from govulncheck workflow because it's done in the GHA composite


rsoaresd (approved):
🚀

PR #1207, opened 2025-09-25
QoL: add early checks for being logged in to quay.io

metlos:
add early checks for being logged in to quay.io to not waste time on e2e tests that are going to fail anyway because of that.

### DIFF
--- a/make/podman.mk
+++ b/make/podman.mk
@@ -1,0 +1,22 @@
+# WILL_PUBLISH is true when the e2e goals will publish new operator image(s)
+# to quay. This enables the check to conditionally execute only when needed.
+WILL_PUBLISH := false
+ifneq ($(HOST_REPO_PATH),)
+WILL_PUBLISH := true
+endif
+ifneq ($(MEMBER_REPO_PATH),)
+WILL_PUBLISH := true
+endif
+ifneq ($(REG_REPO_PATH),)
+WILL_PUBLISH := true
+endif
+
+.PHONY: check-quay-login-if-needed
+check-quay-login-if-needed:
+	@if [ "${WILL_PUBLISH}" = "true" -o "${PUSH_SANDBOX_IMAGE}" = "true" ]; then \
+		podman login --get-login quay.io >/dev/null; \
+	fi
+
+.PHONY: check-quay-login
+check-quay-login:
+	@podman login --get-login quay.io >/dev/null
--- a/make/sandbox-ui.mk
+++ b/make/sandbox-ui.mk
@@ -35,7 +35,7 @@ deploy-sandbox-ui: HOST_NS=$(shell oc get projects -l app=host-operator --output
 deploy-sandbox-ui: REGISTRATION_SERVICE_API=https://$(shell oc get route registration-service -n ${HOST_NS} -o custom-columns=":spec.host" | tr -d '\n')/api/v1
 deploy-sandbox-ui: HOST_OPERATOR_API=https://$(shell oc get route api -n ${HOST_NS} -o custom-columns=":spec.host" | tr -d '\n')
 deploy-sandbox-ui: RHDH=https://rhdh-${SANDBOX_UI_NS}.$(shell oc get ingress.config.openshift.io/cluster -o jsonpath='{.spec.domain}')
-deploy-sandbox-ui:
+deploy-sandbox-ui: check-quay-login-if-needed
 	$(MAKE) check-sso-credentials
 	@echo "sandbox ui will be deployed in '${SANDBOX_UI_NS}' namespace"
 	$(MAKE) create-namespace SANDBOX_UI_NS=${SANDBOX_UI_NS}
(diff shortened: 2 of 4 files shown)


alexeykazakov (approved):

make/sandbox-ui.mk
```diff
@@ -35,7 +35,7 @@ deploy-sandbox-ui: HOST_NS=$(shell oc get projects -l app=host-operator --output
 deploy-sandbox-ui: REGISTRATION_SERVICE_API=https://$(shell oc get route registration-service -n ${HOST_NS} -o custom-columns=":spec.host" | tr -d '\n')/api/v1
 deploy-sandbox-ui: HOST_OPERATOR_API=https://$(shell oc get route api -n ${HOST_NS} -o custom-columns=":spec.host" | tr -d '\n')
 deploy-sandbox-ui: RHDH=https://rhdh-${SANDBOX_UI_NS}.$(shell oc get ingress.config.openshift.io/cluster -o jsonpath='{.spec.domain}')
-deploy-sandbox-ui:
+deploy-sandbox-ui: check-quay-login-if-needed
```
mfrancisc, line 38: we could add it to `push-sandbox-plugin` so that it covers both `deploy-sandbox-ui` and `test-sandbox-ui-in-container`. And eventually future targets that will have to push the image.

make/test.mk
```diff
@@ -62,7 +62,7 @@ prepare-and-deploy-e2e: prepare-e2e deploy-e2e
 	@echo "To clean the cluster run 'make clean-e2e-resources'"
 
 .PHONY: verify-migration-and-deploy-e2e
-verify-migration-and-deploy-e2e: prepare-projects e2e-deploy-latest e2e-migration-setup get-publish-and-install-operators e2e-migration-verify
+verify-migration-and-deploy-e2e: check-quay-login-if-needed prepare-projects e2e-deploy-latest e2e-migration-setup get-publish-and-install-operators e2e-migration-verify
```
mfrancisc, line 65: it looks like you already added the check as part of `get-publish-and-install-operators` , so not sure if we need to add it again here as well.

make/sandbox-ui.mk
metlos: The idea is to perform the check as early as possible. So we really need to find the 'entry points' into the different 'flows' that we invoke using make and figure out a proper place where to add this check.

This is my stab at the entry point to the flow of deploying the UI. @rsoaresd, could you please take a look and tell what good places this check should be put?

rsoaresd: @metlos, if the goal is to check as early as possible, I suggest adding at the top level of `deploy-sandbox-ui` and `test-sandbox-ui-in-container` make target (which I saw you already did).

For now, `push-sandbox-plugin` is used on:
- `.github/workflows/publish-sandbox-ui-for-ui-e2e-tests.yml`, where we already log in to quay before calling `push-sandbox-plugin`, so we are good
- when running `make test-sandbox-ui-in-container`
- when running `test-ui-e2e` or `test-ui-e2e-local` locally which calls `deploy-sandbox-ui`

@mfrancisc has a good point too, if we add future make targets, we could miss adding the early check for being logged into quay.io

make/test.mk
MatousJobanek: btw, do you still face the issue after using the generated token?

make/podman.mk
```diff
@@ -0,0 +1,22 @@
...
+
+.PHONY: check-quay-login-if-needed
+check-quay-login-if-needed:
+	@if [ "${WILL_PUBLISH}" = "true" -o "${PUSH_SANDBOX_IMAGE}" = "true" ]; then \
+		podman login --get-login quay.io >/dev/null; \
+	fi
+
+.PHONY: check-quay-login
+check-quay-login:
+	@podman login --get-login quay.io >/dev/null
```
MatousJobanek, line 22: wouldn't it be easier to execute the check everytime when any makefile target is executed and the `WILL_PUBLISH` is true?

at first, we would need to set the variable aslo when `PUSH_SANDBOX_IMAGE` is `true`:
```makefile
ifeq ($(PUSH_SANDBOX_IMAGE),true)
WILL_PUBLISH := true
endif
```
and then either use `.EXTRA_PREREQS` which is in available in GNU Make 4.3+ (not sure how this would work with Mac
```makefile
.PHONY: check-quay-login
check-quay-login:
ifeq ($(WILL_PUBLISH),true)
	@podman login --get-login quay.io >/dev/null
endif

.EXTRA_PREREQS := check-quay-login
```

or second option do it via `.SHELLSTATUS`, which should work everywhere
```makefile
ifeq ($(WILL_PUBLISH),true)
_ := $(shell podman login --get-login quay.io >/dev/null)
ifneq ($(.SHELLSTATUS),0)
$(error ERROR: Not logged in to quay.io. Run 'podman login quay.io')
endif
endif
```

MatousJobanek: found another option - this could also work:
```makefile
ifeq ($(WILL_PUBLISH),true)
_ := $(or $(shell podman login --get-login quay.io >/dev/null && echo ok),$(error ERROR: Not logged in to quay.io))
endif
```

PR #1210, opened 2025-10-03
test: add test-e2e-with-ui make target

rsoaresd:
# Description
`make test-e2e` is shared between toolchain-e2e, host-operator, member-operator, and registration-service repos. Since we first want to enable the ui e2e tests on toolchain-e2e, we need a new target that runs the current e2e tests plus the ui e2e tests


make/test.mk
```diff
@@ -42,9 +42,19 @@ TESTS_RUN_FILTER_REGEXP ?= ""
 .PHONY: test-e2e
 ## Run the e2e tests
 test-e2e: INSTALL_OPERATOR=true
-test-e2e: prepare-e2e verify-migration-and-deploy-e2e e2e-run-parallel e2e-run e2e-run-metrics
+test-e2e:
```
rsoaresd, line 45: @MatousJobanek, I know that we discussed the possibility of using the inverted logic - checking if REPO_NAME is not 'toolchain-e2e' - but since we need to identify the periodic job by checking if REPO_NAME is empty, I am suggesting this approach. Let me know what you think, please

make/test.mk
```diff
@@ -42,9 +42,19 @@ TESTS_RUN_FILTER_REGEXP ?= ""
...
 test-e2e: INSTALL_OPERATOR=true
-test-e2e: prepare-e2e verify-migration-and-deploy-e2e e2e-run-parallel e2e-run e2e-run-metrics
+test-e2e: 
+	$(MAKE) prepare-e2e verify-migration-and-deploy-e2e e2e-run-parallel e2e-run e2e-run-metrics
 	@echo "The tests successfully finished"
 	@echo "To clean the cluster run 'make clean-e2e-resources'"
+ifeq ($(CI),true)
+  # if we are running on CI, we want to run the ui e2e tests in the toolchain-e2e presubmit and periodic CI jobs
+  # if REPO_NAME is not set, it means that the e2e tests were triggered by the periodic CI job
+  ifneq ($(or $(filter toolchain-e2e,$(REPO_NAME)),$(if $(REPO_NAME),,1)),)
```
MatousJobanek: couldn't this be simplified to:
```suggestion
  ifeq ($(filter-out toolchain-e2e,$(REPO_NAME)),)
```

PR #1285, opened 2026-06-04
feat: add Twilio Lookup v2 API tests

fbm3307:
Add twilio v2 lookup api and test related to it . 

related Prs:
Registration-Service: https://github.com/codeready-toolchain/registration-service/pull/595
toolchain-Common: https://github.com/codeready-toolchain/toolchain-common/pull/532
host operator: https://github.com/codeready-toolchain/host-operator/pull/1268
API: https://github.com/codeready-toolchain/api/pull/509

## Summary by CodeRabbit

* **New Features**
  * Added a configurable phone lookup mode (disabled, enabled, log) affecting signup verification: disabled skips lookups, log records lookup results, enabled can block rejected numbers while skipping lookups for US numbers.

* **Tests**
  * Added end-to-end tests covering phone lookup modes and signup verification outcomes.
  * Added test utilities to wait for signup annotations and validate phone-lookup configuration.

### DIFF
--- a/test/e2e/parallel/phone_lookup_test.go
+++ b/test/e2e/parallel/phone_lookup_test.go
@@ -30,11 +30,12 @@ const (
 )
 
 func TestPhoneLookupMode(t *testing.T) {
-	await := WaitForDeployments(t)
-	hostAwait := await.Host()
+	t.Parallel()
+	awaitilities := WaitForDeployments(t)
+	hostAwait := awaitilities.Host()
 	route := hostAwait.RegistrationServiceURL
 
-	t.Run("log mode stores phone lookup annotations when lookup succeeds", func(t *testing.T) {
+	t.Run("log mode stores annotations and user is provisioned", func(t *testing.T) {
 		// given
 		hostAwait.UpdateToolchainConfig(t, testconfig.RegistrationService().Verification().PhoneLookupMode(toolchainv1alpha1.PhoneLookupModeLog))
 		userSignup, token := signup(t, hostAwait)
@@ -44,7 +45,7 @@ func TestPhoneLookupMode(t *testing.T) {
 			InvokeEndpoint("PUT", route+"/api/v1/signup/verification", token,
 				fmt.Sprintf(`{ "country_code":"+44", "phone_number":"%s" }`, twilioMagicPhoneHighRiskBlocked), http.StatusNoContent)
 
-		// then
+		// then — verification code is set and lookup details are recorded
 		userSignup, err := hostAwait.WaitForUserSignup(t, userSignup.Name,
 			wait.UntilUserSignupHasAnnotationNotEmpty(toolchainv1alpha1.UserSignupVerificationCodeAnnotationKey))
 		require.NoError(t, err)
@@ -52,6 +53,42 @@ func TestPhoneLookupMode(t *testing.T) {
 		assert.NotEmpty(t, userSignup.Annotations[toolchainv1alpha1.UserSignupPhoneLookupDetailsAnnotationKey])
 		assert.NotEmpty(t, userSignup.Annotations[toolchainv1alpha1.UserSignupVerificationCodeAnnotationKey])
 		assert.False(t, states.Rejected(userSignup), "UserSignup should NOT be rejected in log mode")
+
+		// complete verification and confirm user is provisioned
+		NewHTTPRequest(t).InvokeEndpoint("GET", route+fmt.Sprintf("/api/v1/signup/verification/%s",
+			userSignup.Annotations[toolchainv1alpha1.UserSignupVerificationCodeAnnotationKey]), token, "", http.StatusOK)
+
+		userSignup, err = wait.For(t, hostAwait.Awaitility, &toolchainv1alpha1.UserSignup{}).
+			Update(userSignup.Name, hostAwait.Namespace,
+				func(instance *toolchainv1alpha1.UserSignup) {
+					states.SetApprovedManually(instance, true)
+				})
+		require.NoError(t, err)
+
+		VerifyResourcesProvisionedForSignup(t, awaitilities, userSignup)
+	})
+
+	t.Run("enabled mode rejects high-risk phone number", func(t *testing.T) {
+		// given
+		hostAwait.UpdateToolchainConfig(t, testconfig.RegistrationService().
+			Verification().PhoneLookupMode(toolchainv1alpha1.PhoneLookupModeEnabled).
+			Verification().PhoneLookupExcludedCountries([]string{"US", "CA"}))
+		userSignup, token := signup(t, hostAwait)
+
+		// when
+		responseMap := NewHTTPRequest(t).
+			InvokeEndpoint("PUT", route+"/api/v1/signup/verification", token,
+				fmt.Sprintf(`{ "country_code":"+44", "phone_number":"%s" }`, twilioMagicPhoneHighRiskBlocked), http.StatusForbidden).
+			UnmarshalMap()
+
+		// then
+		require.NotEmpty(t, responseMap)
+		assert.Equal(t, "Forbidden", responseMap["status"])
+
+		userSignup, err := hostAwait.WaitForUserSignup(t, userSignup.Name)
+		require.NoError(t, err)
+		assert.True(t, states.Rejected(userSignup), "UserSignup should be rejected in enabled mode with high-risk phone")
+		assert.Empty(t, userSignup.Annotations[toolchainv1alpha1.UserSignupVerificationCodeAnnotationKey])
 	})
 
 	t.Run("enabled mode blocks verification for previously rejected signup", func(t *testing.T) {
@@ -82,7 +119,7 @@ func TestPhoneLookupMode(t *testing.T) {
 		assert.Empty(t, userSignup.Annotations[toolchainv1alpha1.UserSignupVerificationCodeAnnotationKey])
 	})
 
-	t.Run("enabled mode skips lookup for US numbers", func(t *testing.T) {
+	t.Run("enabled mode skips lookup for US numbers and user is provisioned", func(t *testing.T) {
 		// given
 		hostAwait.UpdateToolchainConfig(t, testconfig.RegistrationService().
 			Verification().PhoneLookupMode(toolchainv1alpha1.PhoneLookupModeEnabled).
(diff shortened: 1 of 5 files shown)


test/e2e/parallel/phone_lookup_test.go
```diff
@@ -0,0 +1,129 @@
...
+		userSignup, err := hostAwait.WaitForUserSignup(t, userSignup.Name,
+			wait.UntilUserSignupHasAnnotationNotEmpty(toolchainv1alpha1.UserSignupVerificationCodeAnnotationKey))
+		require.NoError(t, err)
+
+		assert.Empty(t, userSignup.Annotations[toolchainv1alpha1.UserSignupPhoneLookupResultAnnotationKey])
+		assert.Empty(t, userSignup.Annotations[toolchainv1alpha1.UserSignupPhoneLookupPhoneHashAnnotationKey])
+	})
+
+	t.Run("log mode stores phone lookup annotations when lookup succeeds", func(t *testing.T) {
+		hostAwait.UpdateToolchainConfig(t, testconfig.RegistrationService().Verification().PhoneLookupMode(toolchainv1alpha1.PhoneLookupModeLog))
```
metlos, line 40: So this will actually go out to Twilio even in our e2e tests?

fbm3307: that was a good catch, i switched to magic test numbers provided by twilio. so it will actually go to twilio but that wont cost us anything .

test/e2e/parallel/phone_lookup_test.go
```diff
@@ -0,0 +1,135 @@
...
+func TestPhoneLookupMode(t *testing.T) {
+	await := WaitForDeployments(t)
+	hostAwait := await.Host()
+	route := hostAwait.RegistrationServiceURL
+
+	t.Run("phoneLookupMode can be set on ToolchainConfig", func(t *testing.T) {
+		// Non-default value: CRD default "log" is omitted in spec when set explicitly, so use "enabled" to verify persistence.
+		hostAwait.UpdateToolchainConfig(t, testconfig.RegistrationService().Verification().PhoneLookupMode(toolchainv1alpha1.PhoneLookupModeEnabled))
+		VerifyToolchainConfig(t, hostAwait, wait.UntilToolchainConfigHasPhoneLookupMode(toolchainv1alpha1.PhoneLookupModeEnabled))
+	})
```
MatousJobanek: Do we really need to have a test for this? Let's keep the e2e tests focused on the functionality

test/e2e/parallel/phone_lookup_test.go
```diff
@@ -0,0 +1,135 @@
...
+	raw := signup.Annotations[toolchainv1alpha1.UserSignupPhoneLookupDetailsAnnotationKey]
+	if raw == "" {
+		return ""
+	}
+	var details map[string]string
+	require.NoError(t, json.Unmarshal([]byte(raw), &details))
+	return details["result"]
+}
+
+func TestPhoneLookupMode(t *testing.T) {
```
MatousJobanek, line 32: can you please add `// given //when // then` comments?

test/e2e/parallel/phone_lookup_test.go
```diff
@@ -0,0 +1,135 @@
...
+		NewHTTPRequest(t).
+			InvokeEndpoint("PUT", route+"/api/v1/signup/verification", token,
+				fmt.Sprintf(`{ "country_code":"+44", "phone_number":"%s" }`, twilioMagicPhoneHighRisk), http.StatusNoContent)
+
+		userSignup, err := hostAwait.WaitForUserSignup(t, userSignup.Name,
+			wait.UntilUserSignupHasAnnotationNotEmpty(toolchainv1alpha1.UserSignupVerificationCodeAnnotationKey))
+		require.NoError(t, err)
+
+		assert.Empty(t, userSignup.Annotations[toolchainv1alpha1.UserSignupPhoneLookupDetailsAnnotationKey])
+	})
```
MatousJobanek: I think that this is a duplication of the already existing tests, isn't it? Keep in mind that adding more tests increases e2e test execution.

test/e2e/parallel/phone_lookup_test.go
```diff
@@ -0,0 +1,135 @@
...
+
+		userSignup, err := hostAwait.WaitForUserSignup(t, userSignup.Name,
+			wait.UntilUserSignupHasAnnotationNotEmpty(toolchainv1alpha1.UserSignupVerificationCodeAnnotationKey))
+		require.NoError(t, err)
+
+		assert.Equal(t, "rejected", phoneLookupDetailsResult(t, userSignup))
+		assert.Equal(t, "high", userSignup.Annotations[toolchainv1alpha1.UserSignupPhoneLookupCarrierRiskAnnotationKey])
+		assert.Equal(t, "true", userSignup.Annotations[toolchainv1alpha1.UserSignupPhoneLookupNumberBlockedAnnotationKey])
+		assert.NotEmpty(t, userSignup.Annotations[toolchainv1alpha1.UserSignupVerificationCodeAnnotationKey])
+	})
```
MatousJobanek: The important thing here is also that the UserSignup is not rejected and thus also provisioned - I'm missing the verification here

test/e2e/parallel/phone_lookup_test.go
```diff
@@ -0,0 +1,135 @@
...
+
+		userSignup, token := signup(t, hostAwait)
+		_, err := wait.For(t, hostAwait.Awaitility, &toolchainv1alpha1.UserSignup{}).
+			Update(userSignup.Name, hostAwait.Namespace,
+				func(us *toolchainv1alpha1.UserSignup) {
+					if us.Annotations == nil {
+						us.Annotations = map[string]string{}
+					}
+					us.Annotations[toolchainv1alpha1.UserSignupPhoneLookupDetailsAnnotationKey] = `{"result":"rejected"}`
+				})
```
MatousJobanek, line 103: Let's use the rejected state at UserSignup instead

test/e2e/parallel/phone_lookup_test.go
fbm3307: Removed

test/e2e/parallel/phone_lookup_test.go
fbm3307: done

test/e2e/parallel/phone_lookup_test.go
fbm3307: Thanks, removed that duplication, and yes you are right

test/e2e/parallel/phone_lookup_test.go
fbm3307: Added assert.False(t, states.Rejected(userSignup), "UserSignup should NOT be rejected in log mode") to verify the signup is not rejected and can proceed to provisioning.

test/e2e/parallel/phone_lookup_test.go
fbm3307: Done — replaced the manual annotation update

test/e2e/parallel/phone_lookup_test.go
MatousJobanek: why not using `VerifyResourcesProvisionedForSignup`?

test/e2e/parallel/phone_lookup_test.go
```diff
@@ -0,0 +1,105 @@
...
+				fmt.Sprintf(`{ "country_code":"+1", "phone_number":"%s" }`, usTestPhone), http.StatusNoContent)
+
+		// then
+		userSignup, err := hostAwait.WaitForUserSignup(t, userSignup.Name,
+			wait.UntilUserSignupHasAnnotationNotEmpty(toolchainv1alpha1.UserSignupVerificationCodeAnnotationKey))
+		require.NoError(t, err)
+
+		assert.Empty(t, userSignup.Annotations[toolchainv1alpha1.UserSignupPhoneLookupDetailsAnnotationKey])
+		assert.NotEmpty(t, userSignup.Annotations[toolchainv1alpha1.UserSignupVerificationCodeAnnotationKey])
+	})
```
MatousJobanek, line 155: the same - check that the account was not rejected and provisioned as expected

test/e2e/parallel/phone_lookup_test.go
```diff
@@ -0,0 +1,105 @@
...
+//	+441234567890 → high risk, not blocked, score 2
+//	+441234567891 → high risk, blocked, score 34
+//	+911234567890 → low risk, not blocked, score 2
+const (
+	twilioMagicPhoneHighRisk        = "1234567890" // +44 prefix → high risk, not blocked
+	twilioMagicPhoneHighRiskBlocked = "1234567891" // +44 prefix → high risk, blocked
+	usTestPhone                     = "2025550123" // +1 prefix → fictional NANPA 555-01XX range, safe with test credentials
+)
+
+func TestPhoneLookupMode(t *testing.T) {
```
MatousJobanek, line 32: I'm maybe blind, but I don't see the test case where it would verify mode=enabled with high risk phone number.

test/e2e/parallel/phone_lookup_test.go
```diff
@@ -0,0 +1,105 @@
+package parallel
```
MatousJobanek, line 1: just thinking aloud - it should not negatively affect the other registration service tests that do phone verification and are executed as part of the parallel test-suite as well, right? This test enables the lookup api, which means that it can be applied also in the case of the other tests when executed at the same time. However, since the other tests use some dummy phone numbers, then the lookup api should return an error, correct?

MatousJobanek: also, when running in parallel, you need to add 
```go
t.Parallel()
```
to the beginning of the test

test/e2e/parallel/phone_lookup_test.go
fbm3307: okay done

test/e2e/parallel/phone_lookup_test.go
fbm3307: yea it was not there , added it now

test/e2e/parallel/phone_lookup_test.go
fbm3307: Yes, it's safe to run in parallel. With Twilio test credentials, only magic numbers return deterministic responses — any other number (like the +61408999999 used in TestPhoneVerification) triggers an API error from Twilio, which causes fail-open behavior (error is logged, verification proceeds normally). So even if this test temporarily enables lookup mode while another test is running, the other test's non-magic phone number will get an API error and proceed without disruption.

added the t.parallel() at the start

test/e2e/parallel/phone_lookup_test.go
fbm3307: there was no particular reason, we are checking that now

test/e2e/parallel/phone_lookup_test.go
```diff
@@ -44,14 +45,50 @@ func TestPhoneLookupMode(t *testing.T) {
...
+		// complete verification and confirm user is provisioned
+		NewHTTPRequest(t).InvokeEndpoint("GET", route+fmt.Sprintf("/api/v1/signup/verification/%s",
+			userSignup.Annotations[toolchainv1alpha1.UserSignupVerificationCodeAnnotationKey]), token, "", http.StatusOK)
+
+		userSignup, err = wait.For(t, hostAwait.Awaitility, &toolchainv1alpha1.UserSignup{}).
+			Update(userSignup.Name, hostAwait.Namespace,
+				func(instance *toolchainv1alpha1.UserSignup) {
+					states.SetApprovedManually(instance, true)
+				})
+		require.NoError(t, err)
```
MatousJobanek, line 66: we don't need this, do we? it's sufficient to only verify that the user was provisioned.

test/e2e/parallel/phone_lookup_test.go
```diff
@@ -101,5 +138,19 @@ func TestPhoneLookupMode(t *testing.T) {
...
```
MatousJobanek, line 152: the same

MatousJobanek (approved):

test/e2e/parallel/phone_lookup_test.go
fbm3307: @MatousJobanek  I am a bit confused, do you mean to remove the `VerifyResourcesProvisionedForSignup` check ? my understanding is the above code is required if we want to verify user provisioned, the above code is pre steps . 

or may be i can use a lighter version similar to https://github.com/codeready-toolchain/toolchain-e2e/blob/a85e1f4d02b270c41683beef6784df217cb76154/test/e2e/parallel/registration_service_test.go#L627-L637

MatousJobanek: ah, I missed that - forgot that the user stays in verification-required state. all good then

PR #1291, opened 2026-07-06
SANDBOX-1889: update e2e wait helper default to PhoneLookupModeDisabled

fbm3307:
Update UntilToolchainConfigHasPhoneLookupMode to treat nil as PhoneLookupModeDisabled instead of PhoneLookupModeLog.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of an unset phone lookup setting so it now matches the disabled default and reports differences more accurately.


fbm3307:
wrong branch

PR #1309, opened 2026-07-31
Handle compliant username prefix in Space/Idler lookups

### LINKED ISSUE #1308: Perf tool assumes Space name matches UserSignup name
## Bug

The perf tool's `ForSpace` function in `setup/wait/wait.go:20` looks up the Space using the same name as the UserSignup. However, the sandbox operator may transform the username via `compliantUsername` (e.g. `default-0005` becomes `crt-default-0005`), causing `ForSpace` to wait for a Space that will never exist under that name.

This causes the test to fail with:
```
failed to create default template resources for user 'default-0005': space 'default-0005' is not ready yet: context deadline exceeded
```

The same assumption exists in `setup/cmd/root.go:287` and `setup/resources/create_resources.go:34`.

## Reproduction

Run the perf tool with `--username default`:
```
go run setup/main.go --users 10 --default 10 --custom 0 --username default --testname=test --interactive=false
```

UserSignups are created as `default-NNNN` but Spaces are created as `crt-default-NNNN`.

## Expected behavior

The tool should read `status.compliantUsername` from the UserSignup and use that as the Space name for lookups.

Victoremepunto:
## Summary
- The sandbox operator transforms certain usernames via `compliantUsername` (adding a `crt-` prefix for names starting with `default`, `sandbox`, `redhat`, `kube`, `openshift`), but the perf setup tool assumed resource names always match the UserSignup name
- Adds fallback `crt-` prefix lookups in `ForSpace`, `getIdler`, and `CreateUserResourcesFromTemplateFiles` so the tool works regardless of username transformation
- Adds `ForSpaceWithName` to return the resolved space name for correct namespace derivation

Fixes #1308

## Test plan
- [x] New tests for all three affected paths (wait, idlers, resources) — TDD: tests written first, verified red, then implementation made them green
- [x] All existing tests still pass
- [ ] Manual verification with `--username default` on a live cluster

Signed-off-by: Víctor M. Múgica <vmugicag@redhat.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved setup reliability when usernames are transformed into compliant resource names.
  * Space and Idler detection now recognizes both requested and transformed names.
  * Resource creation waits for a resolved space once before deploying resources, reducing unnecessary delays.
  * Preserved existing space-waiting behavior while providing the resolved space name for setup operations.
* **Tests**
  * Added coverage for transformed usernames, Idler timeout updates, space resolution, and resource deployment.

### DIFF
--- a/setup/configuration/configuration.go
+++ b/setup/configuration/configuration.go
@@ -182,6 +182,14 @@ func homeDir() string {
 	return os.Getenv("USERPROFILE") // windows
 }
 
+// CandidateNames returns the possible resource names for a given username.
+// The sandbox operator may add a "crt-" prefix to certain usernames via
+// compliantUsername, so downstream resources (Space, Idler, namespace)
+// may be named differently from the original UserSignup.
+func CandidateNames(name string) []string {
+	return []string{name, "crt-" + name}
+}
+
 func ResultsDir() string {
 	return resultsDir
 }
--- a/setup/idlers/update_idler.go
+++ b/setup/idlers/update_idler.go
@@ -34,16 +34,18 @@ func UpdateTimeout(cl client.Client, username string, timeout time.Duration) err
 func getIdler(cl client.Client, name string) (*toolchainv1alpha1.Idler, error) {
 	idler := &toolchainv1alpha1.Idler{}
 	err := k8swait.PollUntilContextTimeout(context.TODO(), cfg.DefaultRetryInterval, cfg.DefaultTimeout, true, func(ctx context.Context) (bool, error) {
-		err := cl.Get(context.TODO(), types.NamespacedName{
-			Name: name,
-		}, idler)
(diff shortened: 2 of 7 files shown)


rajivnathan:
@Victoremepunto thanks for this PR. While the fix is good, I think it only handles the cases where `crt-` is prefixed but there are other cases for username transformations. I think the performance tool has a lot of room for improvements and I have some ideas but for now I put up [another PR](https://github.com/codeready-toolchain/toolchain-e2e/pull/1311) for consideration that simply blocks using names that require transformation. Since the usernames aren't that important for the purposes of performance testing I think it addresses the problem in the simplest way.

rajivnathan:
closing in favour of https://github.com/codeready-toolchain/toolchain-e2e/pull/1311

PR #1311, opened 2026-08-04
Add username prefix restrictions to avoid unexpected failures

rajivnathan:
Another approach to try to address https://github.com/codeready-toolchain/toolchain-e2e/issues/1308

## Summary by CodeRabbit

* **Bug Fixes**
  * Setup now validates username prefixes against Dev Sandbox restrictions before proceeding.
  * Invalid prefixes produce a clear error explaining the required format and transformed value.


alexeykazakov (approved):

MatousJobanek (approved):
I'm wondering, why not using `TransformUsername` when calling the `ForSpace`?

rajivnathan:
> I'm wondering, why not using `TransformUsername` when calling the `ForSpace`?

Since the originating issue came up due to a naming issue I just figured restricting the username for performance tests was easier to eliminate username naming considerations entirely. This way naming is predictable and easier for other operator teams to use the perf test tool. I think we can actually remove the naming entirely to make it even easier but figured this is a reasonable quick fix.

